PR #89, opened 2016-11-03
Fix scrollbar artifact in tags list

### LINKED ISSUE #87: Thick "scrollbar" on tags list on OS X Sierra
On OS X Sierra (10.12.1) a thick bar appears on the tags list where one would expect to see a scroll bar. In the dark theme it's white, and in the light theme it's black.

It only appears if there are more tags than can fit in the app window and it does not actually scroll, but appears over the entire height of the app.

simplenote simplenote 2

Reported via 2894296-t. Confirmed in v 1.1.7 of the app, but it doesn't appear to happen consistently.

roundhill:
Fixes #87 by unticking the 'auto hide scrollbar' options on the tags list scrollview. 

**To test**
When you build/launch the app, the tags list should appear normally without the artifacts reported in #87.

@jleandroperez I've noticed that the drop-down arrows are not longer appearing next to the tags. Any chance you can take a look at that?

### DIFF
--- a/Simplenote/TagListViewController.m
+++ b/Simplenote/TagListViewController.m
@@ -728,6 +728,7 @@ NSString * const kDidEmptyTrash = @"SPDidEmptyTrash";
 {
     [tableView setBackgroundColor:[[[VSThemeManager sharedManager] theme] colorForKey:@"tableViewBackgroundColor"]];
     [tagBox setFillColor:[[[VSThemeManager sharedManager] theme] colorForKey:@"tableViewBackgroundColor"]];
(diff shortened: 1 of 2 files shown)


jleandroperez:
@roundhill added a `setNeedsDisplay` call that takes care of redrawing the Notes List whenever the user switches themes. I was still getting a white rectangle at the top.

Re: Arrows, working fine on my end, but will take a deeper look. Updating Cocoapods / project settings as well, soon.

Thanks Dan!!

:shipit:

PR #110, opened 2017-01-13
Updates last window behavior

jleandroperez:
Whenever the last window ges closed, the main process will now get killed. We're doing so in order to address iTC's Review Feedback.

cc @roundhill


roundhill:
:shipit:

PR #123, opened 2017-04-27
Updating Last Window Behavior

jleandroperez:
### Details:
In this PR we're restoring the "Dismissing Last Window won't close Simplenote" behavior.

### Testing:
1. Launch Simplenote
2. Click the `x` at the top left corner of the window
3. In the system menu, click over `Note` > `New`
4. Verify that the app is back in the foreground
5. Close Simplenote again
6. Click over the Simplenote icon in the Dock
7. Verify that we're back onscreen!

@roundhill may i bug you with a quick review?
Thanks in advance!!!

cc @drw158 apologies about the late fix, it's been a long road!!


drw158:
Cool! Thanks!

So does this mean we added an item under the Window menu? That's why the app was rejected by Apple if I'm not mistaken.

jleandroperez:
@drw158 i've opted for your `New note` (ensure Window is Onscreen) suggestion. But on second (or third thoughts), since i'm on this, a Window item doesn't sound _that_ off. Mostly because... you may not wanna create a New Note right?.

Pushing a commit in a couple minutes, thanks Dave!

jleandroperez:
**Update:** Added the `Main Window` menuItem (available via the System Bar's `Menu` item.

Ready for review! thanks!!

drw158:
Main Window makes sense to me

roundhill:
Confirmed both `New Note` and `Main window` options launches a new window 👍

roundhill:
When there was no window showing, I clicked move to trash and the action worked, is that expected behavior?

jleandroperez:
@roundhill niiice catch! i've disabled both, **Print** and **Move to Trash** whenever the main window is not visible.

Ready for another look!

Thank you Dan!!

Simplenote/NoteEditorViewController.m
```diff
@@ -526,11 +526,18 @@ - (void)unpublishNote
...
     // Note menu
+    if (menuItem == deleteItem || menuItem == printItem) {
+        return [[SimplenoteAppDelegate sharedDelegate] isMainWindowVisible];
+    }
     if (menuItem == newItem) {
         return !self.viewingTrash;
-    } else if (menuItem == deleteItem) {
+    }
+
+    if (menuItem == deleteItem) {
```
roundhill: This code is never reached now, right? The options are now enabled when I'm viewing the trash.

jleandroperez:
@roundhill my *GOD* what a silly mistake. I totally missed we've got more logic for the `Delete` / `Print` actions below.

Fixed now, thanks!

roundhill:
Looks good! :shipit:

jleandroperez:
Thanks for the review @roundhill!!

roundhill:
> Thanks for the review @roundhill!!

No problem! ❤️

PR #137, opened 2017-11-13
Use Sparkle Pod instead of embedded framework

roundhill:
I was getting a bunch of strange compile errors related to Sparkle. We can use a `Sparkle` pod now, so I've updated the app to use that and it fixes the compile errors.

**To Test**
* `pod install`
* Add `USE_HOCKEY=1` in the preprocessor macros for `Debug` (this will include sparkle in the app delegate):
screen shot 2017-11-13 at 11 02 33 am

* Run the app, it should compile!


jleandroperez:
Looking great Dan!

:shipit:

roundhill:
Thanks @jleandroperez!

PR #138, opened 2017-11-13
Use Native Font / Remove Source Sans Pro

roundhill:
To match the design updates to the other apps, we should remove source sans pro and just use the system default which is San Francisco.

Note: this creates a few oddities with the tags list, but I will fix that in another PR as I start to update its design.

screen shot 2017-11-13 at 12 11 03 pm


jleandroperez:
Note: merge conflicts fixed!

PR #135, opened 2017-11-10
Update Toolbar to Latest Design

roundhill:
I'm starting to update the app to the refreshed design set forth by @drw158. This PR mainly modifies the toolbar layout and some of its behaviors. And at last we get the search box in the top where it belongs :)

screen shot 2017-11-10 at 2 41 17 pm

Most of the changes are new icons and XIB changes to outlets.

**To Test**
Launch the app, and try out all buttons in the toolbar. Expand the notes list size and app size to ensure resizing works as expected. Try out the new trash/restore behavior.

There's already one bug with this, the search field no longer styles itself properly when you change themes.

I'll send over some additional PRs for color and font updates.

PR #139, opened 2017-11-13
Sharing is Caring

roundhill:
_Depends on #135_

Take advantage of the macOS `NSSharingService` (available as of 10.8) to allow the content of a note to be shared out to whatever services are on the system.

I also removed the publishing option from the info menu and added it as an option in the share menu.

screen shot 2017-11-13 at 2 39 47 pm

**To Test**
* Fire up the app, select a note and click on the new publish option.
* Select from a service of your choice, did it share the note content properly?
* Also test the 'Publish to Web' option to ensure it works properly.
* Create a new note, the share button should be disabled until the note has some content.


jleandroperez:
Reverting this merge OP, since the base branch was `update/app-layout`, which was already merged via #135. 

I'll change the base to develop

jleandroperez:
## Note:
The base of this PR was off (got merged into `update/app-layout`). I've moved these changes from `update/app-layout` back into develop via #143.

PR #142, opened 2017-11-27
Search Improvements

roundhill:
I'm aiming to simplify our _previously-kinda-crazy_ search field styling. As of macOS 10.10 we can apply an `NSAppearance` to the search field which gives it a halfway decent look when using the dark theme. With this we can throw out all the stuff we were doing before to try and change background colors and borders around the field. With this change however, macOS 10.9 users will see a white box even for the dark theme. But, it still works and I'm ok with that :)

Light:
screen shot 2017-11-27 at 12 23 57 pm

Dark:
screen shot 2017-11-27 at 12 23 49 pm

I've also updated the search algorithm to match up to iOS. It's a bit smarter now by matching on any terms used in the search string. Previously searching for "notes awesome" with a note content of "notes are awesome" wouldn't work. Now it should work!


roundhill:
Merged via #143. Thanks @jleandroperez!

roundhill:
Whoops, this actually wasn't merged yet.

PR #171, opened 2018-01-25
Adding alert when signing out of the app with unsynced notes.

roundhill:
Some users are losing notes when they sign out because for one reason or another they have unsynced notes. This is a quick solution to that which searches for unsynced notes upon signout and warns the user if so. An option is presented to view the web app to reconcile their unsynced notes:

screen shot 2018-01-25 at 11 59 52 am

**To Test**
* Take the app offline, and create a note.
* Click Simplenote menu and then `Sign Out`. You should see the alert dialog.
* Verify all of the buttons work as expected.
* Also verify that you don't see the prompt if all of your notes have synced.

Refs: https://github.com/Automattic/Simplenote-United/issues/7

### DIFF
--- a/Simplenote/SimplenoteAppDelegate.m
+++ b/Simplenote/SimplenoteAppDelegate.m
@@ -565,10 +565,47 @@
 	return (SimplenoteAppDelegate *)[[NSApplication sharedApplication] delegate];
 }
 
+- (BOOL)hasUnsyncedNotes
+{
+    SPBucket *notesBucket = [self.simperium bucketForName:@"Note"];
+    NSArray *notes = [notesBucket allObjects];
+    for (Note *note in notes) {
+        if ([note.version isEqualToString:@"0"]) {
+            return true;
+        }
+    }
+    
+    return false;
+}
 
 #pragma mark - Actions
 
 - (IBAction)signOutAction:(id)sender
+{
+    // Safety first: Check for unsynced notes before they are deleted!
+    if ([self hasUnsyncedNotes]) {
+        NSAlert *alert = [[NSAlert alloc] init];
+        [alert addButtonWithTitle:NSLocalizedString(@"Sign Out", @"Sign out of the app")];
+        [alert addButtonWithTitle:NSLocalizedString(@"Visit Web App", @"Visit app.simplenote.com in the browser")];
+        [alert addButtonWithTitle:NSLocalizedString(@"Cancel", @"Cancel the action")];
+        [alert setMessageText:NSLocalizedString(@"Unsynced Notes Detected", @"Alert title displayed in when an account has unsynced notes")];
+        [alert setInformativeText:NSLocalizedString(@"Signing out will delete any unsynced notes. You can verify your synced notes by signing in to the Web App.", @"Alert message displayed when an account has unsynced notes")];
+        [alert setAlertStyle:NSAlertStyleCritical];
+        
+        [alert beginSheetModalForWindow:[self window] completionHandler:^(NSInteger result) {
+            if (result == NSAlertSecondButtonReturn) {
+                NSURL *linkUrl = [NSURL URLWithString:@"https://app.simplenote.com"];
+                [[NSWorkspace sharedWorkspace] openURL:linkUrl];
+            } else if (result == NSAlertFirstButtonReturn) {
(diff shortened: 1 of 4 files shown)


jleandroperez:
@roundhill Dan! (unless i'm missing something!!) this one would only let you know if there's a local note that never made it's way to the backend. But local pending changes won't be detected (so you could have a note with changes that won't show up elsewhere).

Is that expected / the desired behavior?. If we wanted to detect local pending changes, we could probably reuse some of the code we've got in `SPIntegrityHelper` (which involves hacking few things, so that we expose a private Simperium method).

jleandroperez:
@roundhill that should do the trick!. Scenarios should be:

### Testing: Note created offline
1. Go offline (shutdown your mac's wifi)
2. Add a new note
3. Try to signout

Verify that the Unsent Changes alert shows up.

### Testing: Existing Note
1. Go offline (shutdown your mac's wifi)
2. Edit an existing note
3. Try to signout

Verify that the Unsent Changes alert shows up.

If all looks good, please, feel free to merge sir!! LOVE this PR!!

jleandroperez:
cc @dmsnell

dmsnell:
I wonder if @drw158 could review those buttons. To me the `Cancel` on the left and the `Sign Out` on the right both seem potentially destructive and therefore their symmetry is confusing. I feel like I might see the cancel, say "no no no," then accidentally click on sign-out and lose my notes because I jumped to the opposite side of the dialog where I expect the safe action to be.

Simplenote/SimplenoteAppDelegate.m
```diff
@@ -569,19 +570,45 @@ + (SimplenoteAppDelegate *)sharedDelegate
...
+    [alert setInformativeText:NSLocalizedString(@"Signing out will delete any unsynced notes. You can verify your synced notes by signing in to the Web App.", @"Alert message displayed when an account has unsynced notes")];
+    [alert setAlertStyle:NSAlertStyleCritical];
+
+    [alert beginSheetModalForWindow:[self window] completionHandler:^(NSInteger result) {
+        if (result == NSAlertSecondButtonReturn) {
+            NSURL *linkUrl = [NSURL URLWithString:@"https://app.simplenote.com"];
+            [[NSWorkspace sharedWorkspace] openURL:linkUrl];
+        } else if (result == NSAlertFirstButtonReturn) {
+            [self signOut];
+        }
```
dmsnell: is this being passed into the `alert` as a body of code to run or is it returning a value as a parameter to `alert`?

drw158:
I agree with Dennis, it should probably say Delete Notes instead of Sign Out. It seems a bit counter-intuitive since the action is actually signing out, but Delete Notes will be super clear and grab their attention.

roundhill:
I think I'll swap the `Cancel` and `View Web App` button placement as well, e.g.:

screen shot 2018-01-29 at 1 47 12 pm

Simplenote/SimplenoteAppDelegate.m
roundhill: I'm not super clear on your question, but the result constants like `NSAlertFirstButtonReturn` represent which button was clicked in the alert.

roundhill:
@jleandroperez Your changes looked good. I updated the button layout and copy, if you think that looks good let's :shipit:

Simplenote/SimplenoteAppDelegate.m
jleandroperez: @dmsnell That block of code is an anonymous function (Closure in Swift, Block in ObjC). And precisely, as you mentioned, it's passed to the alert, to be executed whenever it's dismissed.

**ObjC Block:**
```
^(Type parameter) {
// ..
}
```

**Swift Closure:**
```
{ parameter in
// ..
}
```

Swift can infer the parameter's type, so, as long as it's defined in the receiver function (or property, depending on what you're doing!) you can get away with a compact syntax, like that one.

dmsnell: thanks @jleandroperez - TIL about `^(…) { … }` :smile:

jleandroperez: Anytime sir!! happy to see you picking up iOS!!!

PR #173, opened 2018-02-08
Update about view to match other apps

roundhill:
The other apps had their about views updated last year. macOS still needs it! Here's what it looks like:

screen shot 2018-02-07 at 3 29 40 pm

Some notes on the code: Click events on `NSTextField` objects are actually kind of a pain. I had to do some tricks in the code to get the `NSTextFields` to be clickable (Using `NSClickGestureRecognizer`) and also have a hand pointer icon upon hover (using a subclass).

It also dynamically displays the copyright year and the version of the app 🚀 

I've added a story board and also [got Swifty](https://www.youtube.com/watch?v=I1188GO4p1E)!

**To Test**
* Select Simplenote-About.
* Observe the new screen and click on all the things and ensure they load the correct URL.
* Try opening multiple windows, it should only allow one.
* Bring the main Simplenote window in front of the about view. Selecting Simplenote->About again should bring it to the front.


Gatada:
Clearly I'm a bit too late to comment on this, but the app should display a range of years. The range begins with the year for initial publication, and ends on the current year which is dynamically selected.

Ref.: https://www.copyrightlaws.com/copyright-notice-year/

roundhill:
> I suggest that the app displays a range of years, beginning with the year for initial publication, and ending on the current year - which is dynamically selected.

Thanks @Gatada we'll look into that.

Simplenote/AboutViewController.swift
```diff
@@ -0,0 +1,92 @@
...
+class AboutViewController: NSViewController {
+    let blogUrl =       "https://simplenote.com/blog"
+    let twitterUrl =    "https://twitter.com/simplenoteapp"
+    let appsUrl =       "https://simplenote.com"
+    let githubUrl =     "https://github.com/automattic/simplenote-macos"
+    let hiringUrl =     "https://automattic.com/work-with-us"
+    let privacyUrl =    "https://simplenote.com/privacy/"
+    let termsUrl =      "https://simplenote.com/terms/"
+    
+    @IBOutlet var blogLabel:        SPAboutTextField?
```
jleandroperez: IBOutlets are usually set up as Implicitly Unwrapped Optionals (i _suspect_ the IUO was even made up for this scenario!). I think, mostly, the reason is you want to crash if there's a missing outlet (and it's easier than adding asserts).

Simplenote/AboutViewController.swift
```diff
@@ -0,0 +1,92 @@
...
+import Cocoa
+
+class AboutViewController: NSViewController {
+    let blogUrl =       "https://simplenote.com/blog"
+    let twitterUrl =    "https://twitter.com/simplenoteapp"
+    let appsUrl =       "https://simplenote.com"
+    let githubUrl =     "https://github.com/automattic/simplenote-macos"
+    let hiringUrl =     "https://automattic.com/work-with-us"
+    let privacyUrl =    "https://simplenote.com/privacy/"
+    let termsUrl =      "https://simplenote.com/terms/"
```
jleandroperez: Another Swift trick!

```
struct Constants {
   static let blogUrl = "..."
}
```

(No need to change anything, this is perfect the way it is, just passing along few minor things we do around!)

jleandroperez:
I've missed something that came up in the console. This happens after opening the About Window:

```
2018-02-13 11:08:48.962795-0300 Simplenote[14375:736823] Unknown Window class (null) in Interface Builder file,
	 creating generic Window instead
```

roundhill:
> I've missed something that came up in the console. This happens after opening the About Window:

Hmm appears to be something that will be fixed by 🍏 soon: http://www.openradar.me/35511761

Simplenote/AboutViewController.swift
roundhill: That's... beautiful 😍. Updated to that struct pattern. Mmmm reminds me of tasty Java :)

roundhill:
Merging this one, thanks for the review @jleandroperez

PR #176, opened 2018-02-09
Fix a few markdown bugs

roundhill:
I noticed a few markdown bugs 😄 

This PR should fix the following scenarios:

* If you switch themes while viewing a markdown preview, the preview should reload with the new stylesheet
* If you add a new tag to a note that doesn't have markdown enabled, the preview icon _should not_ appear.
* Viewing the trash should hide the markdown preview, and leaving the trash should show it again as long as the first selected note has markdown enabled.
* If you disable markdown for a note while viewing the preview, it should switch back to the editor.


roundhill:
Thanks @jleandroperez!

Issue #190, opened 2018-03-06
Editor makes change upon opening of note


roundhill:
Notes are getting their modified date updated when simply opening a note for viewing. It's probably something to do with the new editor.

PR #191, opened 2018-03-07
Disable the editor delegate while processing links.

roundhill:
#190 was caused by the new way we process links in the editor using `checkTextInDocument`. It turns out that this causes the `textDidChange` notification to fire for the text view, so the app was modifying the note and saving it which caused the bug.

I have disabled the text view delegate while `checkTextInDocument` is processing. It adds it back after finishing. I think this is OK? If not we can also use a boolean or something to not do any of the `textDidChange` processing while it is running.

**To Test**
Open the app, and the web app for the same account in a browser. Clicking on notes that have URLs in them should not modify them at all and move them anywhere in the notes list.

### DIFF
--- a/Simplenote/NoteEditorViewController.m
+++ b/Simplenote/NoteEditorViewController.m
@@ -293,8 +293,11 @@ static NSInteger const SPVersionSliderMaxVersions       = 10;
 - (void)checkTextInDocument
 {
     dispatch_async(dispatch_get_main_queue(), ^() {
+        // Disable editor delegate while processing links to prevent the editor from modifying the note
+        [self.noteEditor setDelegate:nil];
         [self.noteEditor checkTextInDocument:nil];
         [self.noteEditor setNeedsDisplay:YES];
+        [self.noteEditor setDelegate:self];
     });
 }
 
--- a/Simplenote/NoteEditorViewController.m
+++ b/Simplenote/NoteEditorViewController.m
@@ -293,7 +293,8 @@ static NSInteger const SPVersionSliderMaxVersions       = 10;
 - (void)checkTextInDocument
 {
     dispatch_async(dispatch_get_main_queue(), ^() {
-        // Disable editor delegate while processing links to prevent the editor from modifying the note
+        // Temporarily remove the editor delegate because `checkTextInDocument`
+        // fires `textDidChange` which will erroneously modify the note
         [self.noteEditor setDelegate:nil];
         [self.noteEditor checkTextInDocument:nil];
         [self.noteEditor setNeedsDisplay:YES];
--- a/Simplenote/NoteEditorViewController.m
+++ b/Simplenote/NoteEditorViewController.m
@@ -294,7 +294,8 @@ static NSInteger const SPVersionSliderMaxVersions       = 10;
 {
     dispatch_async(dispatch_get_main_queue(), ^() {
         // Temporarily remove the editor delegate because `checkTextInDocument`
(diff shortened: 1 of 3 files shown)


Simplenote/NoteEditorViewController.m
```diff
@@ -293,8 +293,11 @@ - (void)displayNotes:(NSArray *)notes
 - (void)checkTextInDocument
 {
     dispatch_async(dispatch_get_main_queue(), ^() {
+        // Disable editor delegate while processing links to prevent the editor from modifying the note
```
dmsnell: can we add a little more explanation how `checkTextInDocument` modifies the note?

roundhill: I updated it @dmsnell, does it make more sense now?

dmsnell: definitely better, but I'm still not sure why that changes the note - does it only change it if the editor has unsaved changes? I'm probably just ignorant - don't let that hold up this PR

roundhill: No problem. `checkTextInDocument` is a poorly named native API that does formatting changes to the text in the editor, so it makes links clickable and runs the spell checker. It's firing a textDidChange notification, which is what we use to save a note after a user makes a change to it (like types in a new character).

https://github.com/Automattic/simplenote-macos/blob/7680fc1323143581994902785d9a5ae0378fadb4/Simplenote/NoteEditorViewController.m#L449

dmsnell: so this would mean that every character entered in would sync a new revision in Simperium? is that the error? or is the error the link attributes in the note that iOS adds?

roundhill: The content of the note doesn't change when this fires, but the modification date does. (and this is for macOS)

roundhill: > so this would mean that every character entered in would sync a new revision in Simperium?

No, we use the same two second save timer delay like in electron. All the apps do that.

dmsnell: ah, so we open a note, then `checkTextInDocument` fires, then the modification date changes and the note jumps to the top of the list even though all we did was read it? that makes sense

roundhill: Yes, that's the issue. And this wasn't caught because the macOS app doesn't move a note up in the list even after it is modified so that it keeps its place in the notes list. I wonder if we should change that behavior.

Simplenote/NoteEditorViewController.m
```diff
@@ -293,8 +293,12 @@ - (void)displayNotes:(NSArray *)notes
 - (void)checkTextInDocument
 {
     dispatch_async(dispatch_get_main_queue(), ^() {
+        // Temporarily remove the editor delegate because `checkTextInDocument`
+        // fires `textDidChange` which will erroneously modify the note
```
dmsnell:
```objective-c
// Temporarily remove the editor delegate because `checkTextInDocument`
// fires `textDidChange` which will erroneously modify the note's modification
// time and unintentionally change the sort order of the note in the list as a result
```

^^^ how about something like that?

roundhill: Hahah ok, we'll see what @jleandroperez thinks about long comments 😛

jleandroperez: LOVE long comments, specially if it's hard to understand what the code below does, end to end!!

PR #208, opened 2018-04-29
Feature/better width management

coffeeandscripts:
Screen resolutions of standard MacBooks is 1440x900. A lot of people use two applications adjacent to each other horizontally, so the width would be 720. Currently it is set at minimum width of 740. This was reduced to 720.

Changes also made to the tagList and noteList to conform with aesthetic constraints. If at minimum size noteEditor takes up 360px width, then tagList and noteList should be at 1:2 ratio. tagList therefore set at 120 and noteList at minimum 240.


Podfile.lock
```diff
@@ -53,4 +53,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 8f9476118ed71d8ae1bbd11cc4c58d884bfbdfcf
 
-COCOAPODS: 1.3.1
+COCOAPODS: 1.2.1
```
roundhill: Can you remove this change?

Simplenote.xcodeproj/project.pbxproj
```diff
@@ -1495,7 +1495,7 @@
 				CODE_SIGN_STYLE = Automatic;
 				COMBINE_HIDPI_IMAGES = YES;
 				COPY_PHASE_STRIP = NO;
-				DEVELOPMENT_TEAM = PZYM8XX95Q;
+				DEVELOPMENT_TEAM = 79Z92ZJAS3;
```
roundhill: Can you please revert all changes to this project file? These shouldn't be merged.

roundhill:
:shipit: Thanks @coffeeandscripts!

PR #212, opened 2018-05-21
Add WordPress.com Connect/Sign In

roundhill:
Basically the same as https://github.com/Automattic/simplenote-ios/pull/202 :)

This PR adds support for signing to Simplenote via your WordPress.com account. It supports creating a new Simplenote account if one didn't exist yet, or alternatively it will sign the user to an existing account.

screen shot 2018-05-21 at 12 55 28 pm

**To Test**
* From the login screen tap 'Sign in with WordPress.com'
* Authorize a WordPress.com account
* 3 things should happen:
 ** If a Simplenote account already exists for the account, it should be signed in successfully.
 ** If a Simplenote account didn't exist for the account, a new one should be created and you'll see the welcome note.
 ** If the WordPress.com account hasn't been verified by email, you should get an error message.

We're storing the WP.com token in the keychain for future integrations. If a user signed out, it will be deleted.

_Note: The reviewer will need some new secrets to add to their config file. Ping me on slack to get those._


Simplenote/LoginWindowController.m
```diff
@@ -7,11 +7,91 @@
...
+    [wpccButton setFont:[NSFont systemFontOfSize:16.0]];
+    
+    // A lot of code just to color the button text :|
+    NSMutableAttributedString *colorString = [[NSMutableAttributedString alloc] initWithAttributedString:[wpccButton attributedTitle]];
+    NSRange titleRange = NSMakeRange(0, [colorString length]);
+    NSColor *textColor = [NSColor colorWithCalibratedWhite:120.0/255.0 alpha:1.0];
+    [colorString addAttribute:NSForegroundColorAttributeName value:textColor range:titleRange];
+    [wpccButton setAttributedTitle:colorString];
+    
+    CGFloat centerPosition = (rootView.frame.size.width / 2) - (SPLoginWPButtonWidth / 2);
```
jleandroperez: Can we please round this?. (Non-rounded floats may result in blurry Views!)

Simplenote/LoginWindowController.m
```diff
@@ -7,11 +7,91 @@
...
+
+- (IBAction)wpccSignInAction:(id)sender
+{
+    NSString *plistPath = [[NSBundle mainBundle] pathForResource:@"config" ofType:@"plist"];
+    NSDictionary *config = [[NSDictionary alloc] initWithContentsOfFile:plistPath];
+    
+    NSString *sessionState = [[NSUUID UUID] UUIDString];
+    sessionState = [@"app-" stringByAppendingString:sessionState];
+    [[NSUserDefaults standardUserDefaults] setObject:sessionState forKey:SPAuthSessionKey];
+    NSString *authUrl = @"https://public-api.wordpress.com/oauth2/authorize?response_type=code&scope=global&client_id=%@&redirect_uri=%@&state=%@";
```
jleandroperez: Can we move this one to SPConstants?

Simplenote/LoginWindowController.m
```diff
@@ -7,11 +7,91 @@
...
+    NSString *requestUrl = [NSString stringWithFormat:authUrl, config[@"WPCCClientID"], config[@"WPCCRedirectURL"], sessionState];
```
jleandroperez: I'll definitely need to bug you with an updated `config.plist`!!!

Simplenote/LoginWindowController.m
```diff
@@ -7,11 +7,91 @@
...
+    NSString *errorMessage = NSLocalizedString(@"An error was encountered while signing in.", @"Sign in error message");
+    if (notification.userInfo != nil && notification.userInfo[@"errorString"]) {
+        errorMessage = [notification.userInfo valueForKey:@"errorString"];
+    }
+    
+    NSAlert *alert = [[NSAlert alloc] init];
+    [alert setAlertStyle:NSAlertStyleCritical];
+    [alert setMessageText: NSLocalizedString(@"Couldn't Sign In", @"Alert dialog title displayed on sign in error")];
+    [alert setInformativeText:errorMessage];
+    [alert addButtonWithTitle: @"OK"];
```
jleandroperez: NSLocalizedString(@"OK", @"...")

Simplenote/SimplenoteAppDelegate.m
```diff
@@ -251,6 +258,34 @@ - (void)hookWindowNotifications
...
+    if ([WPAuthHandler isWPAuthenticationUrl: url]) {
+        if (self.simperium.user.authenticated) {
+            // We're already signed in
+            [[NSNotificationCenter defaultCenter] postNotificationName:SPSignInErrorNotificationName
+                                                                object:nil];
+            return;
+        }
+        
+        NSString *plistPath = [[NSBundle mainBundle] pathForResource:@"config" ofType:@"plist"];
+        NSDictionary *config = [[NSDictionary alloc] initWithContentsOfFile:plistPath];
```
jleandroperez: Since this snippet is (at least) in 3 spots (Login, AppDelegate 2x), how about if we do something like....

```
@objc class Secrets: NSObject {

     static let shared = Secrets()

     private let config: [String: Any]
     private init() {
          config = // .. load code!
     }
     var simperiumAppID: String {
         return config["..."
      }
}
```

OR even better, implement the secrets replacement tool we've chatted over Slack. (But if we do so, it probably deserves a different PR, since it's out of the scope of this one!)

roundhill:
Thanks @jleandroperez! I made most of your suggested changes... except the `config` declaration and usage. I agree that we should update that in a separate PR.

Simplenote/LoginWindowController.m
```diff
@@ -7,11 +7,91 @@
...
+static CGFloat const SPLoginWPButtonWidth           = 270.0f;
+static NSString *SPAuthSessionKey                   = @"SPAuthSessionKey";
 
 @implementation LoginWindowController
 
+- (instancetype)init {
+    self = [super init];
+    
+    // Sanity check for accessing the root view
+    if (self.window.contentView.subviews.count < 1) {
```
jleandroperez: Q! can we move this to `windowDidLoad` or it's breaking there?

roundhill: > Q! can we move this to windowDidLoad or it's breaking there?

I tried overriding `windowDidLoad` and it didn't seem to ever get called? 🤷‍♂️

jleandroperez: I've digged a bit into this, and apparently, you need to use one of the designated initializers to get that method called (windowDidLoad).

For the record: because of the way this class is initialized, neither `awakeFromNib`, `windowWillLoad` / `windowDidLoad` nor `loadWindow` will get called.

Probably the only improvement we could do is move the Window Customization code to a method (and shrink a little bit the init), but the win is really marginal.

Squirrel still stands Dan, and thanks for checking!!! :shipit:

PR #213, opened 2018-06-11
Do wp.com sign-in via a separate app window

roundhill:
Previously, wp.com sign in did the authorization via the external web browser, which many users during testing found confusing when the authorization page would stay in the browser tab without closing.

This PR moves that flow to a new`WPAuthWindowController` that has a web view and some simple forward/back buttons. It also is nice because you don't need to authorize the app to open the `simplenote://` protocol.

Screenshot:
screen shot 2018-06-11 at 9 58 43 am

**To Test**
* Sign in using the wp.com sign in flow. Approving or Denying should work as expected.
* Try opening multiple auth windows, it should only allow one instance.
* Completing the sign in or when an error is shown should close the auth window.

### DIFF
--- a/Simplenote/LoginWindowController.h
+++ b/Simplenote/LoginWindowController.h
@@ -18,8 +18,6 @@
  *              extra functionality, Simplenote-Y.
  */
 
-@interface LoginWindowController : SPAuthenticationWindowController <NSWindowDelegate> {
-    WPAuthWindowController *wpAuthWindowController;
-}
-
+@interface LoginWindowController : SPAuthenticationWindowController <NSWindowDelegate>
+    @property (nonatomic, strong) WPAuthWindowController *wpAuthWindowController;
 @end
--- a/Simplenote/LoginWindowController.h
+++ b/Simplenote/LoginWindowController.h
@@ -7,6 +7,7 @@
 //
 
 #import <Foundation/Foundation.h>
+#import "Simplenote-Swift.h"
 @import Simperium_OSX;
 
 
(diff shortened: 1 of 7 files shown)


Simplenote/WPAuthWindowController.swift
```diff
@@ -0,0 +1,16 @@
...
+
+import Cocoa
+import WebKit
+
+class WPAuthWindowController: NSWindowController {
+    
+    @IBOutlet var webView: WebView!
+    
+    @objc public func loadUrl(url: URL) {
+        webView.mainFrame.load(URLRequest.init(url: url));
```
jleandroperez: No need to explicitly call the `.init` + extra `;` at the end 😄 

```
webView.mainFrame.load(URLRequest(url: url))
```

Simplenote/LoginWindowController.h
```diff
@@ -17,6 +18,8 @@
  *              extra functionality, Simplenote-Y.
  */
 
-@interface LoginWindowController : SPAuthenticationWindowController
+@interface LoginWindowController : SPAuthenticationWindowController <NSWindowDelegate> {
+    WPAuthWindowController *wpAuthWindowController;
```
jleandroperez: The "latest" ObjC syntax would look somehow like this:

```
@property (nonatomic, strong) WPAuthWindowController *wpAuthWindowController;
```

That automatically generates the ivar and synthesized setter / getter. (Nevertheless, this snippet is also valid as well!!)

Simplenote/WPAuthWindowController.swift
roundhill: Haha those semicolons will never leave me I guess.

roundhill:
Thanks again @jleandroperez!

Issue #219, opened 2018-06-23
Simplenote 1.3.x doesn't sync on macOS 10.11.6 El Capitan


pauljacobson:
This is a report from a user in 1229935-zen:

>Simplenote works well in Mac Sierra 10.12.6 on my Mac Mini. But I can't get it to sync at all on my Macbook with El Capitan 10.11.6

#### Steps to reproduce

>I've installed the app 3 x today after download from your site and deleted it twice with Appcleaner to be sure all files are flushed. The progress wheel just spins into eternity when I start up the application. In fact it doesn't show a login window at all, just the main window with a welcome screen. I assume it should open to a fresh login screen? In Sierra everything is fine. It downloads my files and also signs out and back in with the proper screen if need be.

>I was using Notational Velocity to access my Simplenote account for years, which still works fine on both machines. Even when I disable Notational on the El Capitan machine, Simplenote just refuses to load any files or sign out at all, just spins.

#### What I expected

Simplenote is not syncing in macOS 10.11.6.

#### What happened instead

Simplenote should sync correctly on macOS 10.11.x. The app officially supports macOS 10.10 according to the Mac App Store description.

#### Browser / OS version

macOS 10.11.6 El Capitan.

mgalbritton:
There is another report of the same issue in 1227265-zen.

mgalbritton:
Another report of the same issue in 1255336-zen.

pcdonohue:
Also reported at 1273052-zen - OSX 10.11.16 user is also not able to sign out of their app.

roundhill:
Taking a look at this.

lizswafford:
Reported on 1273990-zen - **MacOS 10.11.2**, Simplenote 1.3.5.  There is no sign in button, only a sign out button on the login screen for the app. Requested a screenshot. Reinstalling the app didn't change the outcome.

roundhill:
Seeing this exception in the logs on El Capitan: 

```
7/12/18 2:01:01.171 PM Simplenote[354]: +[NSButton buttonWithTitle:target:action:]: unrecognized selector sent to class 0x7fff76375410
```

I think the login window isn't showing because of this exception.

Robertght:
#1273052-zen - another similar case running Mac OS X El Capitan version 10.11.6 (15G22010)

PR #225, opened 2018-07-12
Changed NSButton construction to be compatible with macOS < 10.12

roundhill:
#219 was caused by usage of a convenience constructor for `NSButton` that is [only available as of macOS 10.12](https://developer.apple.com/documentation/appkit/nsbutton/1644256-buttonwithtitle?language=objc).

_This is why we can't have nice things!_ It's weird that Xcode doesn't warn about usage of the constructor...

I've tested this on a El Capitan VM, and the login window now appears and works as expected:

screen shot 2018-07-12 at 2 26 47 pm


roundhill:
Thanks @jleandroperez!

PR #182, opened 2018-02-20
add focus mode

jtoy:
hit control+f and it goes into focus mode. could have used command f, but didnt want to block that frequently used combination. here is a video of it in action: https://cl.ly/413C2t0N2A2M

PR #229, opened 2018-07-25
Add Focus Mode Setting

roundhill:
Continuing the good work started by @jtoy in #182!

This adds a new `Focus Mode` that hides both the tags and notes list to make the editor full width. It also fades the UI buttons in the toolbar when entering focus mode. You can toggle focus mode in the view menu, or by pressing `cmd+shift+f`.

screen shot 2018-07-25 at 2 53 52 pm

**To Test**
* Enable focus mode, and test that the editor works as expected. The search bar should be disabled.
* Try toggling the tags toolbar, it should show the notes list.
* The app should remember if the tags list was opened when you started focus mode and restore its state when focus mode ends.

### DIFF
--- a/Simplenote/SPToolbarView.m
+++ b/Simplenote/SPToolbarView.m
@@ -21,6 +21,9 @@
 #define kSearchCollapsedWidth   120
 #define kSearchExpandedMargin   156
 #define kSearchExpandedWidth    79
+#define kFocusModeOnAlpha       0.5f
+#define kFocusModeOffAlpha      1.0f
+#define kFocusModeDuration      0.8f
 
 @implementation SPToolbarView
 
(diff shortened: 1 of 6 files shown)


jtoy:
yay, so excited to see it live!

Simplenote/SPToolbarView.m
```diff
@@ -86,6 +86,28 @@ - (void)enableButtons:(BOOL)enabled {
...
+        previewButton.animator.alphaValue = alpha;
+        searchField.animator.alphaValue = alpha;
+    }
+    completionHandler:nil];
+}
+
+- (void)configureForFocusMode:(BOOL)enabled {
+    [searchField setEnabled:!enabled];
+    
+    [self setButtonsAlpha:enabled? 0.5 : 1.0];
```
jleandroperez: Suggestion! constant maybe?

Simplenote/SPToolbarView.m
```diff
@@ -86,6 +86,28 @@ - (void)enableButtons:(BOOL)enabled {
     [previewButton setEnabled:enabled];
 }
 
+- (void)setButtonsAlpha:(CGFloat)alpha {
+    [NSAnimationContext runAnimationGroup:^(NSAnimationContext *context) {
+        context.duration = 0.8;
```
jleandroperez: Suggestion! constant maybe?

roundhill:
Thanks @jleandroperez! I added the constants like you suggested.

jleandroperez:
Awesome!!!

PR #231, opened 2018-08-17
Add Mojave Dark Mode Support(ish)

roundhill:
Mojave fixes a lot of the pain points we had with applying a dark theme to the app, particularly around the search box. This PR uses the new system colors added in 10.14 for the background, divider and text colors. The light theme looks pretty much the same, but the dark version is now less 'blue' in the background and has better contrast to boot.

We don't really need the theme menu in Mojave and beyond, so I've also conditionally removed that.

New Light Theme:
screen shot 2018-08-17 at 3 44 00 pm

New Dark Theme:
screen shot 2018-08-17 at 3 43 34 pm

Users running pre-Mojave should still receive the old theme coloring. I also updated a few icons to match better with the new dark theme colors. It was really easy to do with the new asset catalog options for the dark appearance.

I say `ish` in the title, because we could take this a step further in the future and apply `Material` to the views so we could have the gradients and transparency that you see in other native macOS apps. But I think for now this preserves the design of the app while fitting in with Mojave's color scheme for both light and dark.


roundhill:
cc @drw158 for any thoughts!

DB5/VSTheme.m
```diff
@@ -134,7 +134,19 @@ - (NSImage *)imageForKey:(NSString *)key {
...
+        if ([key isEqualToString:@"dividerColor"]) {
+            return NSColor.separatorColor;
+        } else if ([key isEqualToString:@"tableViewBackgroundColor"]) {
+            return NSColor.controlBackgroundColor;
+        } else if ([key isEqualToString:@"textColor"] || [key isEqualToString:@"noteHeadlineFontColor"]) {
+            return NSColor.textColor;
+        } else if ([key isEqualToString:@"secondaryTextColor"] || [key isEqualToString:@"noteBodyFontPreviewColor"]) {
+            return NSColor.secondaryLabelColor;
+        }
+    }
```
jleandroperez: Suggestion: Since this is probably bound to grow, maybe move the contents of the if to a method?

roundhill:
Thanks! I'm going to hold off merging this until Apple gives the go-ahead for submitting Mojave ready apps.

PR #232, opened 2018-08-20
Add a maximum note editor width and setting

roundhill:
This PR limits the note editor width to 750px to match the Electron app. Since some users will most likely want the full width editor, I've added an option to the menu so that it can be enabled again.

Before:
screen shot 2018-08-20 at 12 20 36 pm

After:
screen shot 2018-08-20 at 12 20 22 pm

**To Test**
* Open the app, and make the note editor larger than 750px wide. You should see padding to the left and right of the editor content.
* Select View->Note Editor->Full Width. The note editor should go full width.
* The markdown preview should also show the padding or not depending on the full width setting.
* Ensure the editor looks and works properly in Focus Mode with the full width setting enabled and disabled.

### DIFF
--- a/Simplenote/SPTextView.m
+++ b/Simplenote/SPTextView.m
@@ -8,6 +8,8 @@
 
 #import "SPTextView.h"
 
+#define kMaxEditorWidth 750 // Note: This matches the Electron apps max editor width
+
 @implementation SPTextView
 
 // Workaround NSTextView not allowing clicks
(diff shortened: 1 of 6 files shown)


Simplenote/SPTextView.m
```diff
@@ -25,4 +27,22 @@ - (void)mouseDown:(NSEvent *)theEvent
...
+    CGFloat insetX = [self shouldCalculateInset:viewWidth] ? [self getAdjustedInsetX:viewWidth] : kMinEditorPadding;
+    [self setTextContainerInset: NSMakeSize(insetX, kMinEditorPadding)];
+}
+
+- (BOOL)shouldCalculateInset: (CGFloat)viewWidth {
+    return viewWidth > kMaxEditorWidth && ![[NSUserDefaults standardUserDefaults] boolForKey:kEditorWidthPreferencesKey];
+}
+
+- (CGFloat)getAdjustedInsetX: (CGFloat)viewWidth {
+    CGFloat adjustedInset = (viewWidth - kMaxEditorWidth) / 2;
```
jleandroperez: We probably need to round this value. (Positions with floating values might look blurry onscreen!)

jleandroperez:
### systemBar Screenshot:

screen shot 2018-08-22 at 1 43 21 pm

roundhill:
@drw158 I just had 'full width' with a check mark. But maybe it'd make more sense to do as you suggested and do View->Note Editor->Line Length->Narrow & Full Width options?

screen shot 2018-08-22 at 12 39 47 pm

PR #234, opened 2018-08-24
Hide some toolbar elements when in focus mode

roundhill:
Based on design feedback. The search bar, new button and divider in the toolbar will now hide when in focus mode:

screen shot 2018-08-24 at 10 57 43 am


roundhill:
Thanks, as always @jleandroperez!

Issue #207, opened 2018-04-27
Highlighted text color change for better readability


Robertght:
Hey!

One of our awesome users, pointed out that the highlighted text while using the Dark theme on the Mac app is quite hard to read. I've given a shot myself at this and it appears to be so.
![simplenote 2018-04-27 10-08-05](https://user-images.githubusercontent.com/6737345/39349250-ea63d712-4a02-11e8-8bef-2f3a002871d0.png)

Can we consider a different color for the text to increase readability?

pauljacobson:
I received another report of this issue in 1578132-zen where the user included this screenshot:

![1578132_simplenote_dark-mode_text-highlight](https://user-images.githubusercontent.com/2912202/48832473-e7c7df80-ed81-11e8-9fca-e3a99dcc3631.png)

The user is running macOS High Sierra (they're unable to upgrade to Mojave). Please look into this issue as soon as possible.

w-:
Hi. This is still an issue. A fix would be great here. thanks

bummytime:
As alluded to above, this is not an issue in Mojave:

<img width="827" alt="Screen Shot 2019-05-15 at 11 37 49 AM" src="https://user-images.githubusercontent.com/154014/57793031-f7f3c280-7705-11e9-9fab-d180c644dbab.png">

jleandroperez:
Closing this one since it's no longer relevant. Cannot repro in macOS Mojave (and higher), and older macOS releases aren't officially supported anymore.

Thanks everyone!!

PR #235, opened 2018-08-30
Make the selected text readable in dark mode

vishaltelangre:
Similar issue is already reported here - https://github.com/Automattic/simplenote-macos/issues/207. Therefore, instead of reporting a duplicate issue, I forked and added a simple solution in this pull request to address the readability problem with the selected text in dark mode.

### Problem

As of now, when using SimpleNote app on macOS in dark mode, the selected text is not readable at all as show in the below screenshot (v1.3.6).

image

### Solution

Please check the following GIF image.

ly39m5rykq

### DIFF
--- a/Simplenote/NoteEditorViewController.m
+++ b/Simplenote/NoteEditorViewController.m
@@ -993,7 +993,18 @@ static NSInteger const SPVersionSliderMaxVersions       = 10;
 		[theNoteEditor addObserver:self forKeyPath:property options:NSKeyValueObservingOptionNew context:nil];
 	}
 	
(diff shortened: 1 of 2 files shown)


roundhill:
This became naturally fixed in Mojave, so I guess we need to decide if this is still worth merging for older versions 🤔

bummytime:
Thank you for the PR @vishaltelangre. 🙇  As @roundhill mentioned, Mojave fixed this issue so we are going to close this PR.

vishaltelangre:
@bummytime That’s fine. I love Simplenote’s dark mode in Mojave.

PR #246, opened 2018-10-08
Allow Theme Selection to Override System Theme Setting

### LINKED ISSUE #245: Allow theme to be changed in Mojave and beyond
Some users would like to be able to change the theme, regardless of the system light or dark setting.

roundhill:
Brings back the theme menu in macOS >= 10.14 that will override the system theme setting. It will only override once the user actively changes the theme setting, otherwise it just obeys the global macOS setting.

**To Test**
* On Mojave, use the theme menu to change the theme. It should persist when you restart the app.
* In `applicationDidFinishLaunching`, add this line to simulate that you've never selected a theme setting: `[[NSUserDefaults standardUserDefaults] removeObjectForKey:VSThemeManagerThemePrefKey];`
* Verify that switching the global mojave theme also changes the theme in Simplenote.

I've tested this on macOS 10.10 and it still works as expected there as well.

Fixes #245.

### DIFF
--- a/DB5/VSThemeManager.h
+++ b/DB5/VSThemeManager.h
@@ -22,5 +22,6 @@ extern NSString *const VSThemeManagerThemePrefKey;
 - (VSThemeLoader *)themeLoader;
 
 - (void)swapTheme:(NSString *)theme;
+- (BOOL)isDarkMode;
 
 @end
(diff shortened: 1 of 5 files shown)


roundhill:
> * Perhaps we could encapsulate the NSUserDefaults check in didFinishLaunching?

I moved the code to `VSThemeManager` 👍 

> * When relaunching the app, you can see the "Initial Dark Mode", and after a split second the switch to "Light Mode" (in case of an override).

Nice catch! I think I've fixed this by applying the appearance override in `SPWindow` instead of the app delegate. Give it another run through if you don't mind!

JasDhaliwal:
Thank you @roundhill version 1.3.8 looks great.

PR #250, opened 2018-10-10
Tracks now requires user opt-in

roundhill:
Changes the Privacy Settings to default to tracks being turned off by default. I added a new UserDefault that simply tracks a boolean when the user has flipped the switch. Once it has been switched, we can then enable tracks.

**To Test**
* Launch the app and view privacy settings. Verify the switch is *Off*.
* Add a breakpoint in `SPTracker` to verify that events are not sent.
* Flip the switch to on! Verify that track events start sending.

### DIFF
--- a/Simplenote/Preferences.m
+++ b/Simplenote/Preferences.m
@@ -10,9 +10,9 @@
 
 @dynamic analytics_enabled;
 
-- (void)awakeFromInsert
+- (void)awakeFromLocalInsert
 {
-    [super awakeFromInsert];
+    [super awakeFromLocalInsert];
     self.analytics_enabled = @(false);
 }
 
--- a/Simplenote/Preferences.m
+++ b/Simplenote/Preferences.m
@@ -10,4 +10,10 @@
 
 @dynamic analytics_enabled;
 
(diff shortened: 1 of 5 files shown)


jleandroperez:
@roundhill i crashed your PR Dan!! we're gonna need to cross review this puppy, i'm afraid.

### Testing:
0. Nuke the Store data, located here: `/Users/[YOUR_USER]/Library/Containers/com.automattic.Simplenote/Data/Library/Simplenote/`. 
1. Install version 1.3.6 (this is why step 0 is there!)
2. Log into a testing account (if possible, it would be an account that lacks the **Preferences** object).
3. Install this branch on top
4. Open the Preferences Panel

Verify that Analytics are disabled by default.

Thanks in advance sir!!

Simplenote/Preferences.m
```diff
@@ -10,4 +10,10 @@ @implementation Preferences
 
 @dynamic analytics_enabled;
 
+- (void)awakeFromLocalInsert
```
roundhill: Simperium dark magic! ✨

jleandroperez:
Thank you Dan!!

PR #263, opened 2018-10-22
Allow user to reset to system appearance setting

### LINKED ISSUE #256: Theme: dark/light mode does not change with OS
I believe the expected behavior is for a manual theme change to override the OS theme change. However, it looks like this is being overridden if the theme was changed before the update as well, which makes it look like it has just stopped working altogether.

@roundhill mentioned that a future update can instead have this behavior reset upon sign out from the app.

#### Steps to reproduce
1. Change theme from light to dark
2. Update to the newest Simplenote app version for MacOS (1.3.8)
3. Change OS to dark mode or light mode

#### What I expected
I expected that the Simplenote theme would change along with the OS, at least until I made a manual change.

#### What happened instead
I am unable to get the Simplenote app to respond to an OS theme change. 

#### Simplenote version
Simplenote 1.3.8

#### OS version
Mac 10.14 - Mojave

roundhill:
In this PR we're adding a new item to the Theme menu on Mojave and beyond that will allow the user to go back to using the system appearance:

screen shot 2018-10-22 at 3 44 05 pm

This will be the default checked item if the user has never manually selected a theme (on Mojave only, otherwise it still defaults to `Light`).

**To Test**
* Open the app in Mojave, and select `View -> Theme -> System Appearance`.
* Switch the appearance in the main macOS settings, the app should adjust appearance accordingly.
* Now manually select a dark or light theme in `View -> Theme`. 
* Switching the macOS appearance should have no effect on the appearance of Simplenote.

Fixes #256

### DIFF
--- a/DB5/VSThemeManager.h
+++ b/DB5/VSThemeManager.h
@@ -23,5 +23,6 @@ extern NSString *const VSThemeManagerThemePrefKey;
 
 - (void)swapTheme:(NSString *)theme;
 - (BOOL)isDarkMode;
+- (BOOL)isMojaveWithNoThemeSet;
 
 @end
--- a/DB5/VSThemeManager.m
+++ b/DB5/VSThemeManager.m
@@ -91,4 +91,12 @@ NSString *const VSThemeManagerThemePrefKey = @"VSThemeManagerThemePrefKey";
     return isDarkTheme;
 }
 
+- (BOOL)isMojaveWithNoThemeSet {
+    if (@available(macOS 10.14, *)) {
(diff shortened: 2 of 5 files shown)


jleandroperez:
@roundhill i may have found a glitch, details below!

### Scenario:
0. Set macOS's Appearance to Dark
1. Launch Simplenote + Log in
2. Select Dark Theme `View > Theme > Dark`
3. Select Dark Theme `View > Theme > System Appearance`
4. Switch macOS's system appearance to Light

As a result, Simplenote will be stuck in Dark Mode (until you actually relaunch the app)

roundhill:
Thanks @jleandroperez it should be fixed now for you. I should have set the appearance to `nil` there so that macOS applies the system appearance again.

PR #278, opened 2018-12-17
Checklists!

roundhill:
Using a very similar approach (and copied code!) from https://github.com/Automattic/simplenote-ios/pull/258, this PR adds a checklist UI, using markdown syntax behind the scenes.

screen shot 2018-12-17 at 3 33 30 pm

**To Test**
* Open a note, and create a checklist (look in the `Format` menu)! Pressing return at the end of a checklist item should automatically add a new one on the next line.
* Open the note on app.simplenote.com, it should have the correct markdown syntax displayed.
* Update the note on the web browser, the app should update and display any modified checklists appropriately.
* Test things that modify the note content, like the history browser and ensure they work as expected.


Simplenote/SPTextAttachment.swift
```diff
@@ -0,0 +1,28 @@
+//
+//  SPTextAttachment.swift
+//  Simplenote
+//  Used in the note editor to distinguish if a checklist item is ticked or not.
+//
+
+import Foundation
+
+@objcMembers class SPTextAttachment: NSTextAttachment {
+    var checked = false
```
jleandroperez: Maybe `private` so that it cannot be swapped from the outside?

PR #281, opened 2019-01-04
Don’t use global CocoaPods

jkmassel:
We're trying to get away from global CocoaPods installations for our iOS (and macOS!) projects in order to ensure tooling consistency – this applies that policy to this project.

It just requires a `bundle exec` in front of `pod` commands. :) 

(see: https://github.com/Automattic/Automattic-Tracks-iOS/pull/85)


roundhill:
Sounds good. Should we update the `README`? Do you just run `bundle install` then `bundle exec pod install`?

jkmassel:
That's right! I'll adjust this PR to update the `README` as well, thanks! :)

jkmassel:
Added! If you're good with it @roundhill, I think this is ready to merge!

roundhill:
Looks good! Should we apply the same to simplenote-ios?

PR #283, opened 2019-01-07
Updating Checklist Insertion Behavior

roundhill:
This is essentially the same code changes as https://github.com/Automattic/simplenote-ios/pull/266

**To Test**
* Test inserting a checkbox using the `Format -> Insert Checklist` menu.
* If the cursor is on a line with an existing checkbox, that checkbox should be removed and the cursor position should adjust accordingly.
* If the cursor is on a line without an existing checkbox, a checkbox should be inserted and the cursor position should adjusted accordingly.

### DIFF
--- a/Simplenote/SPTextView.m
+++ b/Simplenote/SPTextView.m
@@ -17,9 +17,10 @@
 NSString *const CheckListRegExPattern = @"^- (\\[([ |x])\\])";
 NSString *const MarkdownUnchecked = @"- [ ]";
 NSString *const MarkdownChecked = @"- [x]";
+NSString *const TextAttachmentCharacterCode = @"\U0000fffc"; // Represents the glyph of an NSTextAttachment
 
 // One unicode character plus a space and a newline
-NSInteger const ChecklistCursorAdjustment = 3;
+NSInteger const ChecklistCursorAdjustment = 2;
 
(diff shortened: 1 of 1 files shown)


roundhill:
> * Constant to replace `2` in code.
> * Toggling when cursor is at start of line.
> * Toggling when multiple lines are selected.

Great catches! Should be all sorted now!

Simplenote/SPTextView.m
```diff
@@ -134,33 +134,67 @@ - (BOOL)checkForChecklistClick:(NSEvent *)event
...
-        // Add a checkbox
+        // Add checkboxes to the selection
         NSString *checkboxString = [MarkdownUnchecked stringByAppendingString:@" "];
-        lineString = [checkboxString stringByAppendingString:[self.string substringWithRange:lineRange]];
+        NSArray *stringLines = [lineString componentsSeparatedByString:@"\n"];
+        for (int i=0; i < [stringLines count]; i++) {
+            NSString *line = stringLines[i];
+            // Skip any empty lines, except the first one
+            if (i != 0 && [line length] == 0) {
+                continue;
```
frosty: I think if there's an empty line, we need to add a newline to the result string? Otherwise we lose any empty new lines when toggling a multi-line selection.

PR #284, opened 2019-01-07
Fix versions slider with checklists

roundhill:
The checklists were not displaying properly in the note history browser. I've also taken the opportunity to show the history slider from beneath the history button, instead of to the side of the note editor which makes more visual sense to me.

screen shot 2019-01-07 at 2 39 59 pm

I also removed the text color from being changed while viewing history, it was really low contrast.

**To Test**
* View the history for a note that has a checklist.
* Changing/restoring versions should work as expected.

### DIFF
--- a/Simplenote/NoteEditorViewController.m
+++ b/Simplenote/NoteEditorViewController.m
@@ -639,7 +639,6 @@ static NSInteger const SPVersionSliderMaxVersions       = 10;
         [versionSlider setObjectValue:[NSNumber numberWithInteger:versionSlider.maxValue]];
         [self updateVersionLabel:self.note.modificationDate];
         [self.noteEditor setEditable:NO];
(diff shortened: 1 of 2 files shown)


roundhill:
> * The Restore button in the versions view doesn't seem necessary? I can just drag the slider and then start editing the older version of the note, and it'll replace the latest version? Should we flip the note content back if you dismiss the versions panel without clicking Restore?

Turns out you do need to click the button to actually restore the note. We weren't resetting to the latest version of the note if the popover was canceled. I fixed that!



> * The Restore button text seems really low contrast. Could we make it white to match the text?

Good call, I replaced it with the default button style:
screen shot 2019-01-08 at 1 19 02 pm
screen shot 2019-01-08 at 1 24 02 pm

I also noticed that the Restore button was enabled when it first loads. It should be disabled when on the latest version, so I just set it to disabled in the xib to fix that.

roundhill:
Thanks @frosty!

PR #306, opened 2019-02-21
Update Tracks to the most recent version

jkmassel:
This PR updates the Tracks library to the latest version.

**To Test:**
- Perform any action that would generate a Tracks event, and ensure it arrives correctly


bummytime:
@jkmassel I took the opportunity to merge the latest from `develop` into this PR as well as pin the `Automattic-Tracks-iOS` pod to `0.3.3`. If you are good with it, we can merge this in.

(I also tested this PR and everything is working ✅ )

jkmassel:
This should be ok to merge – the main reason this PR existed was because we changed how the Tracks project works for macOS targets. If everything compiles nicely with this PR, and tracks events end up in the right place, it's ok by me to merge! 😄

PR #312, opened 2019-03-20
Enable hoedown's Span-Extentions

forevernull:
Enable hoedown's Span-Extentions including strikethrough

Then markdown can support these extends:

- strikethrough： ~~stikethrough~~
- underline： _underline_
- highlight : ==highlight==
- superscript:  100^2
- math: $$math$$


bummytime:
> Looks amazing, thank you for helping us improve Simplenote!

Indeed. Many thanks @forevernull!

PR #323, opened 2019-05-07
Added build tooling + XCode 10.2/Swift 5 updates

bummytime:
### Fix

This SNmacOS PR is very similar to https://github.com/Automattic/simplenote-ios/pull/300

Lots of tooling and dependency updates here — this PR updates Simplenote for macOS in the following ways:
* XCode 10.2 is now supported
* Swift 5 conversions occurred (and is now the project default)
* A number of deprecation warnings were addressed
* `Gemfile` was updated
* `Podfile` was updated
* `.gitignore` was updated
* `Rakefile` was added (it is essentially the [same version as WPiOS](https://github.com/wordpress-mobile/WordPress-iOS/blob/develop/Rakefile))
* `readme.md` was updated accordingly

**Note 1:** fastlane was not added in this PR

**Note 2:** While the bulk of deprecation warnings were addressed in this PR, some remain — they will be addressed in a subsequent PR. 

### Test

1. From a freshly cloned repo, verify `bundle install`, and/or `rake dependencies` works properly
2. Open `Simplenote.xcworkspace` in Xcode 10.2 and verify it builds/installs/runs (when running the app, try a few basic operations — login, create a new note, edit an existing note, delete a note, etc)
3. Test anything else that y'all think is appropriate.

Note: the SNmacOS project does not contain any unit tests....yet. 😄 

@jkmassel and @loremattei → I know you are all crazy-busy, but if you could put some eyes on this, I would _really_ appreciate it!! 🙇


bummytime:
@jkmassel — just added `.bundle/config` back in :-D Ready for another peek

bummytime:
Thanks @jkmassel

PR #325, opened 2019-05-14
Replace Fabric with sentry

jkmassel:
Adds support for using Sentry via Tracks.

Most of the logic is over in Automattic/Automattic-Tracks-iOS#103. Once that's merged, I'll update this PR with a real version number, instead of a `pod` branch.

**To Test**

1. Put `CrashLogging.crash()` (you'll need to `import AutomatticTracks`) in `AboutViewController.swift` 
2. Create a release version of the app using Build > Archive. Export it to your desktop to run it.
3. Cause a crash by viewing the app's About Screen.
4. Re-run the app, and let it idle for a moment (this'll ensure the crash report goes to Sentry)
5. Note your shiny new crash entry in Sentry (it'll likely get attached to an existing sample crash from my testing. However yours should be the most recent event.


bummytime:
@jkmassel can the podfile in this branch be updated with `pod 'Automattic-Tracks-iOS', '~> 0.4'` now?

jkmassel:
Yeah! Done in https://github.com/Automattic/simplenote-macos/pull/325/commits/4e8c39bea39e7a822e72e7c0933904fa61abd938

Simplenote/AboutViewController.swift
```diff
@@ -28,7 +29,9 @@ class AboutViewController: NSViewController {
     
     override func viewDidLoad() {
         super.viewDidLoad()
-        
+
+        CrashLogging.crash()
```
bummytime: I'm assuming you'll remove this line before merging into `develop`? 😄

jkmassel: Removed in 65bdace1adc4cd559b62b3de1e40b0a3b6cc7e72! 😅

PR #332, opened 2019-05-28
Cleanup

jleandroperez:
### Details:

Okay. I'm not really sure how this happened (or where it came from), but:

- We had a **nib** file with no name in the project. Looks like a prototype of the **About** screen
- We were copying these files inside the bundle [`.gitignore` + `README.md` + `Credits.rtf`]. They came from **INAppStoreWindow** (which is a tool we use to adjust the NSWindow's appearance).

cc @bummytime 
Thanks in advance!!


jleandroperez:
Thank you sir!!

PR #364, opened 2019-08-09
Merge Release/1.6 into develop

jtreanor:
Merge Release/1.6 into develop

### DIFF
--- a/config/Version.public.xcconfig
+++ b/config/Version.public.xcconfig
@@ -1,4 +1,4 @@
 VERSION_SHORT=1.6
 
 // Public long version example: VERSION_LONG=9.9.0.0
-VERSION_LONG=1.6.0.2
+VERSION_LONG=1501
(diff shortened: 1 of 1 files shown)


config/Version.public.xcconfig
```diff
@@ -1,4 +1,4 @@
 VERSION_SHORT=1.6
 
 // Public long version example: VERSION_LONG=9.9.0.0
-VERSION_LONG=1.6.0.1
+VERSION_LONG=1501
```
jtreanor: @loremattei I have to change the version format here to avoid AppStore Connect rejecting the version saying that 1.6.0.1 is smaller than the existing version (1500). We will need to figure out how to update the tooling for this.

loremattei: @jtreanor sounds good! I think the version here should be `1601` though as the "old" style naming was to use the long version without the dots.

jtreanor: Yeah, I realised that after I submitted. I think we can leave it this time 😄

loremattei: Yep! If you already submitted, I agree that we can live with it :-)

PR #394, opened 2019-11-06
Fixing crash at Launch

### LINKED ISSUE #393: Crash at Launch
### Details:
No precise repro steps. Happens (randomly) right after signing in, caught with the debugger.

Screen Shot 2019-11-05 at 5 42 29 PM

jleandroperez:
### Fix
In this PR we're fixing a crash (caught with the debugger) after logging in.

@aerych May I bug you with a really quick one?
Thanks in advance!!

Closes #393

### Test
I'm afraid there aren't repro steps for the original crash, it was quite random. Please verify the change makes sense, and verify logging into your account works fine!

### Release
`RELEASE-NOTES.txt` was updated in 2b2c084 with:

> Fixed a bug that crashed the app at launch

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,3 +1,8 @@
+1.7.1
+-----
+- Fixed a bug that crashed the app at launch
+
+
 1.6.0
 -----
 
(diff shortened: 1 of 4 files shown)


jleandroperez:
Thank you Eric!!

PR #396, opened 2019-11-08
Fixing Emoji's AWOL Bug

### LINKED ISSUE #395: Editor Breaks when adding emojis
### Details:
Verified in Xcode 11.1 / 11.2 in macOS Catalina.

1. Log into your account
2. Add a new note
3. Enter an emoji
4. Notice it appears onscreen
5. Add another emoji

As a result, the newly added emoji will be the only one displayed. 

Demo II

jleandroperez:
### Fix
In this PR we're fixing a bug that caused emojis to disappear in macOS Catalina + Xcode 11.x.

@aerych Last quick one, I promise!! Thanks in advance!!!
Closes #395 

### Test
1. Log into your account
2. Create a new note
3. Insert 😃
4. Insert ☺️

- [x] Verify both emojis show up side by side!

### DIFF
--- a/External/Notepad/Storage.swift
+++ b/External/Notepad/Storage.swift
@@ -100,7 +100,7 @@ class Storage: NSTextStorage {
         replaceBackingStringSubrange(range, with: attrString.string)
 
         let change = attrString.length - range.length
-        self.edited(.editedCharacters, range: range, changeInLength: change)
+        self.edited([.editedCharacters, .editedAttributes], range: range, changeInLength: change)
         self.endEditing()
     }
 
--- a/External/Notepad/Storage.swift
+++ b/External/Notepad/Storage.swift
@@ -150,10 +150,28 @@ class Storage: NSTextStorage {
         let indexRange = string.lineRange(for: nsRange)
         let extendedRange = NSUnionRange(editedRange, NSRange(indexRange, in: string))
 
-        perserveRealEditedRange {
-            applyStyles(extendedRange)
-            super.processEditing()
+        /// In macOS 10.15 (Catalina), editing documents that contain Emojis end up disappearing . We restore them by reapplying our font to the full edited range.
+        /// *But* in macOS Catalina, *UNLESS* we signal we've `.editedAttributes` with the fully edited range, the UI ends up broken.
+        ///
+        /// Now, the side effect of doing so, is that the `selectedRange` ends up being kicked to the end of the document (because, alledgedly, we've edited the full string).
+        ///
+        /// This is a major hack that allows us to:
+        ///
+        ///     A. Apply a given font to the entire BackingStore
+        ///     B. Signal that we've edited the font (so that emojis are properly rendered)
+        ///     C. Preserve the *actual* selectedRange (rather than kicking the cursor to the end of the string).
+        ///
+        ///  Ref. https://github.com/Automattic/simplenote-macos/pull/396
+        ///
+        if #available(macOS 10.15, *) {
+            shouldOverrideSelectionRange = true
+            overrideSelectionRange = NSRange(location: editedRange.location + editedRange.length, length: 0)
         }
+
+        applyStyles(extendedRange)
+        super.processEditing()
+
+        shouldOverrideSelectionRange = false
     }
 
     /// Applies styles to a range on the backingString.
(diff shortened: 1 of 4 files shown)


dmsnell:
Can you explain in the description/summary above what about this change fixes the glitch and why it occurred in the first place? I know there's something about Catalina in there but what was the bug?

jleandroperez:
Writing this down, here, for the record:

- Emojis going AWOL is a macOS catalina bug
- We're using a NSTextStorage subclass
- Editing the font within `processEditing` and **not calling** `edited(.attributes` signaling the full range causes a broken UI (again, only in Catalina).

Obvious fix was to signal we've edited the attributes, for the full range. Now, downside? that causes the caret to be repositioned at the end of the document.

We've implemented a really REALLY hacky workaround [via df207ae](https://github.com/Automattic/simplenote-macos/pull/396/commits/df207aed3e7d6beb6b96c62f8eda869dddb8e63a). We're in the talks with AAPL regarding the font error in console (yet another SDK bug), I'll be touching base with them, perhaps there's a cleaner approach.

Ready for another look @aerych, thanks in advance!!!

Simplenote/SPTextView.m
```diff
@@ -209,4 +209,22 @@ - (NSString *)getLeadingWhiteSpaceForString: (NSString *)string
...
 
+- (void)setSelectedRange:(NSRange)selectedRange
+{
+    NSRange patchedSelectionRange = [self fixSelectedRange:selectedRange];
+    [super setSelectedRange:patchedSelectionRange];
+}
+
+// HACK HACK: Only required in macOS Catalina.
+// See Storage.swift `perserveRealEditedRange` for details!
+- (NSRange)fixSelectedRange:(NSRange)selectedRange
```
aerych: `Storage` internally knows whether it needs to provide the default `selectedRange` or the `overrideSelectionRange` courtesy of the `shouldOverrideSelectionRange`.   It might be nicer if `Storage.selectedRange` could internally check for the flag and return the correct value rather than having to expose the flag to external callers and have the conditional logic below.   The encapsulation would be nicer.  What do you think?

External/Notepad/Storage.swift
```diff
@@ -15,9 +15,21 @@
...
+    /// See `perserveRealEditedRange` for details.
+    ///
+    @objc
+    private(set) var shouldOverrideSelectionRange = false
+
+    /// Contains the "Real" Edited Range.
+    /// See `perserveRealEditedRange` for details.
+    ///
+    @objc
+    private(set) var overrideSelectionRange = NSRange()
```
aerych: If `overrideSelectionRange` was an optional we could maybe do away with the boolean `shouldOverrideSelectionRange`.  Just guard for `nil` to know whether to use the default selectedRange or not.

Just a thought.

External/Notepad/Storage.swift
```diff
@@ -143,31 +150,36 @@ class Storage: NSTextStorage {
         let indexRange = string.lineRange(for: nsRange)
         let extendedRange = NSUnionRange(editedRange, NSRange(indexRange, in: string))
 
-        applyStyles(extendedRange)
-        super.processEditing()
+        perserveRealEditedRange {
+            applyStyles(extendedRange)
+            super.processEditing()
+        }
```
aerych: I'm not a fan of calling `super` from within a closure this way.  It works but feels wrong, ya know? 

Related to my earlier comment about loosing the boolean flag, it might work to call `applyStyles(extendedRange)` from within the block, move the `super` call to after the block, and follow up the super call by always setting `overrideSelectionRange` to `nil`.   I think that would encapsulate all the relevant logic nicely, and keep the `super` call out of the block scope. 
Thoughts?

aerych:
Tested the changes and the issue with the return is resolved. :)

Simplenote/SPTextView.m
jleandroperez: That was my first approach... but unfortunately `Storage`'s `super.processEditing()` needs the actual `editedRange` (that is, including the full range that got the right font), otherwise the Emojis go missing.

Storage posts the edited range over to the LayoutManager, which is in charge of updating the TextView's selectedRange. I've **tried** overriding the LM's mechanism, but doing so implies overwriting a method (and **not** calling super, otherwise you get the faulty behavior).

A bit of a cornered situation, I'm afraid. I suspect the true underlying problem are the NSFont warnings we are getting in console (and that's what we've asked Apple about!).

Thanks for the suggestion Eric!!

External/Notepad/Storage.swift
jleandroperez: Definitely approached things that way, **but** Objective C. Only way to determine, in ObjC, if a range is invalid, was to use a negative location as a marker, that's why I deferred to the dual property approach.

Like those old ObjC API's... `-(BOOL)doSomethingWithError:(NSError **)error`. I wish we could swift our way out, though!!

External/Notepad/Storage.swift
jleandroperez: We can loose the closure mechanism altogether! Detail is: both the flag + range must be set **before** `super.processEditing()`, and cleaned up afterwards.

Would it be okay with you, perhaps, if we remove the closure mechanism, and merge both things in the same spot? (only reason why I've opted for a closure is the "old behavior preservation" we've got, the code ends up much cleaner).

WDYT? (Thanks!!!!)

aerych: Would love to lose the closure! Let's try what you described :)

Simplenote/SPTextView.m
aerych: Ouch!  Understood sir.

jleandroperez:
@aerych I kinda love the way this ended up looking like!! Thanks for all the feedback, ready for another look!!

dmsnell:
Are we sure something else isn’t broken with the way we use the storage view? A bug in Catalina of this nature seems like it would have crashed the release immediately.

Did we file a bug report with Apple?

jleandroperez:
@dmsnell Been thru the docs x5 times, reached out AAPL's code level support (still waiting for their reply). And spent around 8 hours digging into this.

You can never be sure with this kind of bug, but if it's something wrong in our side, it's definitely not obvious, I'm afraid.


It's not the first time we see an SDK bug that requires a major workaround, Aztec is... full of this kind of "fixes"!

jleandroperez:
Thank you Eric!!!

PR #398, opened 2019-11-14
Fixing Emojis Crash

### LINKED ISSUE #397: Crash: Emojis Again!
### Details:
1. Add a new note
2. Enter the following emojis: `☺️🖖🏿`
3. Insert the following emoji in between: `😃`

As a result the app will crash. We'll end up with a malformed unicode sequence, which effectively breaks Apple's JSON Parser.

---

### Extra!
Because of the malformed diff, AAPL's JSON Encoder breaks in the following spot:

Screen Shot 2019-11-06 at 8 43 01 PM

jleandroperez:
### Fix
In this PR we're mapping a (patched) version of Simperium, which makes sure `Surrogate Pairs` aren't split in the middle, and thus, causing invalid unicode strings.

@danielebogo May I bug you with a real fun PR?
Thanks in advance!!

Closes #397

### Test
1. Add a new note
2. Enter the following emojis: ☺️🖖🏿
3. Insert the following emoji in between: 😃

- [x] Verify the app doesn't break!

### Release
`RELEASE-NOTES.txt` was updated in 3af363e with:
 
> Fixed a bug that caused certain emojis combinations to crash the app


jleandroperez:
Thank you @danielebogo !!! I'll dig deeper into the Welcome Note issue (couldn't trigger it over here on the first pass).

PR #408, opened 2019-11-28
Add beta build lane

loremattei:
This PR adds a lane to build and upload a beta release to AppCenter. The lane takes also care of uploading symbols to Sentry.

### Test
1. Update your `~/.simplenotemacos-default.env` to include the new required params (look at `fastlane/env.example` for the list)
2. `bundle install`
3. Run `bundle exec fastlane build_and_upload_beta`.
4. Verify that the build succeeds. 
5. Verify that the binary has been uploaded to AppCenter.
6. Verify that the symbols have been uploaded to Sentry.


fastlane/Fastfile
```diff
@@ -99,13 +117,45 @@ ENV["PUBLIC_CONFIG_FILE"]="./config/Version.Public.xcconfig"
...
+  # Usage:
+  # bundle exec fastlane build_and_upload_beta [skip_confirm:<skip confirm>]
+  #
+  # Example:
+  # bundle exec fastlane build_and_upload_beta 
+  # bundle exec fastlane build_and_upload_beta skip_confirm:true 
+  #####################################################################################
+  desc "Builds and updates for beta distribution"
+  lane :build_and_upload_beta do | options |
+    build_app(
```
jkmassel: I hit an error here because my pods weren't up to date – do we want to run `pod install` as part of this?

fastlane/Fastfile
```diff
@@ -99,13 +117,45 @@ ENV["PUBLIC_CONFIG_FILE"]="./config/Version.Public.xcconfig"
...
+      workspace: "Simplenote.xcworkspace", 
+      scheme: "Simplenote",
```
jkmassel: I got a warning here:

```
Couldn't find specified scheme 'Simplenote'. Please make sure that the scheme is shared, see https://developer.apple.com/library/content/documentation/IDEs/Conceptual/xcode_guide-continuous_integration/ConfigureBots.html#//apple_ref/doc/uid/TP40013292-CH9-SW3
```

The scheme is shared, but I'm still seeing the issue, which is odd 🤔

It's not failing, so we could ship with it like this, but perhaps something might break later?

fastlane/Fastfile
loremattei: Yep! You're right. I added that and some other checks.

fastlane/Fastfile
loremattei: Mmm... you are right. 

I dug a bit and found that the `Simplenote` scheme is the default one, but it seems to be autogenerated by XCode and it's not stored in `Simplenote.xcodeproj` in the repository.
If you open XCode and try to edit the scheme, the repository ends up to be dirty because the default `Simplenote.xcodeproj/xcshareddata/xcschemes/Simplenote.xcscheme` is generated. 

I'm not sure why we don't have the file in the repository, though I remember I had some issues related to this in the past. 

@jleandroperez do you think we can add the file for the Simplenote scheme to the repo?

loremattei: Also, I noticed that now the lane actually switches to building the `Simplenote-AppStore`, which is not what we want 😅

jleandroperez: @loremattei I've checked... AND [found this](https://stackoverflow.com/questions/53034303/should-xcshareddata-be-checked-into-git-repo-for-xcode-project).

As per Xcode 9.3, it's okay to check in the `xcshareddata` folder. Feel free to push it (happy to do it on my end, don't wanna crash your PR without asking for permission first!!).

Thank you for doing this, gentlemen!!

loremattei:
@jleandroperez @jkmassel I pushed the scheme. This should be ready for another look. 
Thank you!

PR #428, opened 2019-12-23
Updates Secrets Path

jleandroperez:
### Fix
I've spotted, pretty much by accident, an incorrect path. Simplenote macOS was pulling secrets from the iOS credentials path.

In this PR we're fixing such mapping.

### Test
- [x] Verify the app builds correctly
- [x] Verify that you can log into your account!

### Release


jleandroperez:
Thank you Bummy!!

PR #435, opened 2019-12-30
Restoring Trash Menu

### LINKED ISSUE #249: Empty trash arrow went missing
I think there used to be a little arrow that would appear on hover near the trash button. You can still click where it used to show, and the empty trash popup appears:

screen shot 2018-10-10 at 9 58 49 am

I would vote to just update the context menu when you right-click on the trash to have the empty trash option there.

jleandroperez:
### Fix
In this PR we're restoring the specific **NSMenu** (we always had) that was displayed whenever the user right-clicked over the **Trash** label.

Closes #249

cc @bummytime 
Thank you Bummy!!

### Test
1. Launch Simplenote
2. Trash a Note
3. Right click over the **Trash** label

- [x] Verify a Menu shows up, and the option **Empty Trash** is enabled
- [x] Verify that clicking **Empty Trash** effectively removes all of the trashed notes
- [x] Verify that the **Empty Trash** action is disabled whenever there are no Trashed notes

### Release
`RELEASE-NOTES.txt` was updated in d15eab1 with:

> Fixed a bug in which the Trash couldn't be emptied with a right click action

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,3 +1,7 @@
+1.9.0
+-----
+- Fixed a bug in which the Trash couldn't be emptied with a right click action
+
 1.8.1
 -----
 - Fixed a bug that caused notes with specific emoji sequences to crash the app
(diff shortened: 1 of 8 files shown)


jleandroperez:
Thank you Bummy!!

PR #437, opened 2020-01-02
MainMenu IB Cleanup

jleandroperez:
### Details:
As per Xcode 11, autolayout is no longer optional at the top nib level. Meaning that it's enabled, by default, everytime you save an IB file.

In this PR we're:

- *Manually* setting everything to **Autosizing** (that's the way it's currently setup!).
- Implemented Autolayout in the NotesList TableViewCell.
- Nuked two IB unwired entities.
- And we're adjusting few IB frames, that are currently causing warnings.

cc @bummytime (Thanks in advance sir!!)

### Test:
- [x] Verify the main UI looks as expected, and can be resized
- [x] Verify the Notes List's width can be properly adjusted
- [x] Verify the (i) menu looks great
- [x] Verify the Versions Slider works as expected

### Release:


jleandroperez:
Thank you sir!! Checklists bugo addressed via #439 

🔥

PR #446, opened 2020-01-14
Update Fastlane to 2.140.0

jkmassel:
### Fix
Fixes a warning in Github about a vulnerable Ruby dependency. This is a minor version update and shouldn't break anything with fastlane.

### Test
Ensure all automated tests pas.

### Review
Just needs a quick 👍 from one person

### Release


Gemfile
```diff
@@ -5,7 +5,7 @@ source 'https://rubygems.org' do
   gem 'xcpretty-travis-formatter'
   gem 'octokit', "~> 4.0"
   gem 'dotenv'
-  gem "fastlane", "~> 2.137"
+  gem 'fastlane', '~> 2'
```
loremattei: @jkmassel will this work well on CI? I mean: when install is run on CI, if a newer version than the one in `Gemfile.lock` is available, the checkout on the CI of the `Gemfile.lock` will change and the check that we have to make sure that it doesn't change will make the build fail. 

Since we have to update the `Gemfile.lock` anyway in order to make things work, isn't it a better solution to lock the version in the `Gemfile` as well?

jkmassel: Hmmm – that seems unexpected. The `.lock` file is supposed to ensure that dependencies are locked to specific versions on `bundle install` – updating the `.lock` file should require `bundle update`. To test this, you could create a branch from this one, and change `fastlane (2.140.0)` in `Gemfile.lock` to `fastlane (2.139.0)`, then `rm -rf vendor && bundle install`. It should show in the output that it's installing `2.139.0` and not dirty the repo, but LMK if not?

Whenever we update `fastlane`, we basically always want the most recent compatible version without breaking changes (vs just taking the version that fixes our little banner warning) – that's what this version constraint expresses, and it frees us from having to look up the version each time. Also, this change would enable us to have a CI job that automatically opens PRs for new `fastlane` versions – that might help us stay a bit more up to date and reduce noise in our build output 😄 

I'm ok to do it either way, this is just very slightly easier when doing updates. LMK your thoughts? :)

loremattei: Oh! You're right. With `bundle`, `npm` and `pod` having slightly different behaviours and commands for this, I got confused. 

The automatic tool to update the version sounds great to me :-)

PR #463, opened 2020-03-05
Updating deployment target

jleandroperez:
### Fix
In this PR we're bumping up the deployment target to macOS High Sierra (10.13), and fixing several warnings that came up.

@frosty you up for a mac PR? (thank you!!)

🔥 

### Test
- [ ] Run a smoke test of the app
- [ ] Verify Tags can be rearranged (drag and drop)

### Release
`RELEASE-NOTES.txt` was updated in 392f2c9 with:
 
> Simplenote now requires macOS 10.13 or higher


jleandroperez:
- `Process kill()` is SDK related, couldn't nail down it's source yet, but it's on my list!
- Simperium error noted, will track it as well. Reordering is sync'ing, soo it's probably something odd going on with the backend!!

Thanks for the review James!!!

Issue #458, opened 2020-03-04
Spring Cleaning: Phase I


jleandroperez:
### 🔨 Debt 🔨
- [x] Drop **INAppStoreWindow** since it's not really needed anymore
- [x] Drop VSThemeManager
- [x] Drop dead assets
- [x] Decouple VersionsViewController
- [x] Decouple ShareViewController
- [x] Decouple PublishViewController
- [x] Implement MarkdownViewController
- [x] Drop AppDelegate's IBOutlets

### 🎶 New UX 🎶   
- [x] Display NSWindow's titleBar
- [x] Refresh Tags List
- [x] Refresh App Icons
- [x] Re-Engineer ToolbarView: Review Encapsulation + Ownership
- [x] Re-Engineer SearchBar: Review Encapsulation + Ownership
- [x] Implement ColorStudio Tooling
- [x] Implement New Color Scheme
- [x] Implement the new (Vectorial) Assets
- [x] Implement our New Blue!
- [x] Implement SplitViewController
- [x] Implement New Notes List
- [x] Implement Extended Content in the Notes List
- [x] Refresh UI Icons: 24x24 >> 20x20
- [x] Restore Window TitleBar
- [x] Implement Dividers: Sidebar / Toolbar
- [x] Refresh Editor
- [x] Implement TagsEditor
- [x] Implement TagsEditor Scrolling Support
- [x] Implement Untagged Notes Support #120
- [x] Implement ModifiedUI
- [x] Refresh Popovers Style
- [x] Refresh ToolbarView
- [x] Review RTL Support

### 📸  Screenshots 📷 
<img width="300" src="https://user-images.githubusercontent.com/1195260/75882567-30cd6700-5e00-11ea-864f-5bf61aa32aab.png">
<img width="300" src="https://user-images.githubusercontent.com/1195260/75882608-45116400-5e00-11ea-9623-6a3a0695bf13.png">
<img width="300" src="https://user-images.githubusercontent.com/1195260/75882664-5ce8e800-5e00-11ea-9dbc-ad2b4b782540.png">

codebykat:
User in 2995663-zen thinks the missing sidebar toggle icon is a bug and requests its return.

jleandroperez:
Officially closing **Spring Cleaning Phase 1**... 

Thank you @aerych @guarani @danielebogo @etoledom @emilylaguna for the million + 1 reviews !!!!!!

🔥🔥🔥🔥🔥

PR #464, opened 2020-03-06
Drops Custom Scroller

jleandroperez:
### Details
In this PR we're dropping a custom subclass of NSScroller, AppKit's class in charge of drawing the Scroll Bars.

The custom subclass was the source of a crash in macOS < 10.12, and the UI difference is next to impossible to tell, with the naked eye. Aiming at less custom code, always.

Cleanup part of #458 
cc @frosty James, you up for yet another macOS PR?

Thaaankyouuu

### Test
- [ ] Verify the app builds correctly
- [ ] Verify that scrolling over the Notes List does show the vertical scroll bar
- [ ] Verify that scrolling over the Editor shows the scroll bar

### Release


jleandroperez:
Thank you James!!

PR #466, opened 2020-03-11
Checklists: Undo Exceptions. Mark I

jleandroperez:
### Description:
This is the first part of a few PR(s) aiming at:

A. Encapsulating all the things Checklists
B. Refactoring the main mechanism, so that the UndoManager doesn't throw exceptions
C. Testing all the things

### Details:
- Added new Unit Testing Target. **Seriously**
- Updated the main Podfile so that 10.13 is the deployment target (matching the main project)
- Encapsulated the **NSAttributedString** to **Markdown** converter into an actual class
- Added Unit Test for the above!
- Updated CircleCI so that Unit Tests are ... effectively executed all over!
- And minor touches here and there (dead code deletion // overall cleanup)

cc @aerych (Thank you Eric!!!)
Ref. #467

### Test
1. Add a new note
2. Insert a checklist
3. Add several items (and swizzle the checked / unchecked state)
4. Relaunch the app

- [x] Verify that the note looks exactly the same way it was before relaunching
- [x] Verify the (new!) Unit Tests are green

### Release


Simplenote/NSAttributedStringToMarkdownConverter.swift
```diff
@@ -0,0 +1,33 @@
...
+//
+class NSAttributedStringToMarkdownConverter: NSObject {
+
+    /// Markdown replacement for "Unchecked Checklist"
+    ///
+    private static let unchecked = "- [ ]"
+
+    /// Markdown replacement for "Checked Checklist"
+    ///
+    private static let checked = "- [x]"
```
aerych: I want to say these were localized in the iOS app. Any issue with not localizing the `x` here?

Simplenote/SPTextView.m
```diff
@@ -15,7 +15,6 @@
 
 #define kMaxEditorWidth 750 // Note: This matches the Electron apps max editor width
 NSString *const MarkdownUnchecked = @"- [ ]";
```
aerych: In the interest of having a single source of truth, maybe we could nuke this const? It's only used one other place in this file. Maybe make NSAttributedStringToMarkdownConverter `@objc` friendly and its static members not private and reference `NSAttributedStringToMarkdownConverter.unchecked` @ line 140?

SimplenoteTests/NSAttributedStringToMarkdownConverterTests.swift
```diff
@@ -0,0 +1,66 @@
...
+
+            markdown += prefix + payload
+        }
+
+        return (document, markdown as NSString)
+    }
+
+    func documentWithoutAttachments() -> NSAttributedString {
+        let sample = """
+                     orem ipsum dolor sit amet, consectetur adipiscing elit,
```
aerych: Lorem?

jleandroperez: Ahhhhhh thanks for catching this!!

Simplenote/SPTextView.m
jleandroperez: +10000 for canonical source of truth!! Hoooowever (apologies I didn't mention this in the PR description), the other two spots in which is is used will be removed / heavily refactored in a follow up.

Thank you for bringing this up Eric!!

jleandroperez:
Thank you Eric!!

Simplenote/NSAttributedStringToMarkdownConverter.swift
jleandroperez: Noting for the record: Chatted with Eric few days ago, the localized snippet is the input converter (which performs a case insensitive check).

Okay to leave it this way for the output converter, since nothing (should) break.

**Thank you!!!**

Issue #467, opened 2020-03-12
UndoManager: Exceptions


jleandroperez:
### Description:
The UndoManager ends up throwing an exception whenever the document being edited contains a Checklist.

1. New Document
2. Insert a new Checklist
3. Add a few lines
4. Hit CTRL + Z few times

As a result, you'll get an exception in the Console:

```
2020-03-12 12:31:48.347792-0300 Simplenote[54444:871909] [General] *** -[Swift.__StringStorage substringWithRange:]: Range {9, 11} out of bounds; string length 16
2020-03-12 12:31:48.350566-0300 Simplenote[54444:871909] [General] (
	0   CoreFoundation                      0x00007fff2e3e98ab __exceptionPreprocess + 250
	1   libobjc.A.dylib                     0x00007fff646a3805 objc_exception_throw + 48
	2   Foundation                          0x00007fff309da61d -[NSString substringWithRange:] + 223
	3   AppKit                              0x00007fff2b78397d -[NSTextStorage(NSUndo) _undoRedoAttributedSubstringFromRange:] + 145
	4   AppKit                              0x00007fff2bd141d7 -[NSUndoTyping undoRedo:] + 142
	5   Foundation                          0x00007fff30b08a99 -[_NSUndoStack popAndInvoke] + 187
```

### Details:
1. Checklists are internally represented by means of **NSTextAttachments** embedded within the main **NSTextStorage**
2. Documents stored in Simperium encode checklists as `- [ ]` or `- [x]`
3. Inserting a new checklist results in **SPTextView** injecting a `- [ ]` string, and then running a RegEx that turns such substring into a **NSTextAttachment** (for Image purposes)

What's the problem:
1. The Regex / Replacement happen at the **NSTextStorage** level, and thus, the Undo Manager is unaware of the latest changes
2. Hitting CTRL + Z attempts to remove the `- [ ]` marker (when instead we've got an attachment of length 1)

Even if we patched up the NSTextAttachment replacement, hitting CTRL + Z would end up stripping the Checklist Image, which is **Definitely** not what we want.

### Tasks:
- [x] New **NSAttributedString** to **Markdown** Converter (saving the document)
- [x] Hitting **newline** should also insert a **NSTextAttachment** _directly_
- [x] Inserting a Checklist should insert a **NSTextAttachment** _directly_
- [x] New **NSTextView.displayNoteWithContents** API: Process Lists + Reset UndoManager . This prevents List Processing from being undoable.
[...]

PR #476, opened 2020-03-30
Lists: Undo Exceptions Mark 4.1

jleandroperez:
### Fix
This is a tech debt PR:

1. We're encapsulating a RegEx inside a Swift extension, rather than an ObjC define
2. New constants as well
3. A ton of (new) Unit Tests for the new RegEx extension
4. And a ton of new Unit Tests for our Undo support in NSTextView

Ref. #467

cc @etoledom sir, may I bug you with a reeeeally easy to verify PR?
Thank you!!!

### Test
- [ ] Verify the new Unit Tests are green

### Release


jleandroperez:
Thanks for reviewing this one Eduardo!!!

PR #479, opened 2020-04-03
Color Studio: Mark 1

jleandroperez:
### Details:
In this PR we're adding ColorStudio machinery, laying out the foundation for Spring Cleaning follow ups.

@diegoreymendez sir, may I bug you with this one? 
Thanks in advance!!

Ref. #458

### Testing:
- [ ] Verify the app builds correctly
- [x] And make sure the code makes sense!

Nothing has been wired _yet_, just adding tooling!

### Release


jleandroperez:
@diegoreymendez I've spent some time refactoring this one, in order to use `xcassets`, but there's a major problem there. NSColor's initializer looks like this:

```
init?(named name: NSColor.Name)
```

`NSColor.Name` is nothing more than an alias (it's a String), meaning that we'd need to use Extensible Enums, or keep the ColorStudio enum around.

On top of that, we'd be unable to work around the failable initializers. Instead, we'd be forced to implement static API(s) that given a Color Name, we'd provide a 'fallback color' (along with an assertion, to catch the faulty scenario).

The end result is a solution less elegant than the Hex parser. Would you mind if we leave it this way?

Thank you!!

jleandroperez:
Thank you Diego!!

PR #491, opened 2020-04-08
Decoupling ShareViewController

jleandroperez:
### Details:
- Splits `ShareViewController` UI from the MainNib
- Reimplements `ShareViewController` in Swift

@diegoreymendez sir, may I bug you with this one?
Thanks in advance!

Ref. #458 

### Test
1. Launch Simplenote and open any note
2. Press the top right (i) button
3. Press over `Collaborate`

- [x] Verify the ShareViewController shows correctly
- [x] Verify Share looks great in Dark Mode

### Release


jleandroperez:
For the record, verified over here, the email is only posted whenever the recipient is already a Simplenote user.

Thank you Diegooo!!!!

PR #493, opened 2020-04-10
Unlinking ViewController(s)

jleandroperez:
### Details:
All of our ViewControllers share references to each other, set via Interface Builder. Due to a big refactor we're working on, each nib needs to be in its own file.

In this PR we're severing the ties (Part 1!)

cc @aerych Sir, you game for a surgical PR?

Ref. #458

### Test
Please run a smoke test:

- [x] Verify `add a note` works as expected
- [x] Verify that the newly added note is immediately selected (in the notes list)
- [x] Verify that editing a note causes the Notes List to immediately refresh
- [x] Verify that clicking over notes in the Notes List causes the Editor to update its contents
- [x] Verify that `Empty Trash` works as expected

### Release


jleandroperez:
Thank you Eric!!

Issue #497, opened 2020-04-15
Building with Fastlane doesn't support notarization


loremattei:
Currently, the `build_and_upload_beta` lane doesn't support notarization and, since we want to distribute notarized betas to our testers, it can't be used. 

We have to update it in order to handle notarization.

PR #498, opened 2020-04-15
Issue/notarization

loremattei:
### Fix
This PR enables `hardened-runtime` for the beta version of the App, so that it can be submitted for notarization to Apple.
For the same reason, this PR also resigns  the Sparkle binaries with the `hardened-runtime` flag.

**Note:** It looks like `fastlane` has some issues with signing Sparkle with `developer-id`, so this PR is not updating the `Fastfile`. Since we want to release a new beta of Simplenote during this Sprint, I've opened an issue [here](https://github.com/Automattic/simplenote-macos/issues/497) so that we'll look into it later. In the meantime, we have to switch back to manual builds with XCode. cc/ @mokagio. 

### Test
1. In XCode, build and archive the app.
2. Go to Organizer, select the build, click on Distribute->Developer ID->Upload.
3. Select the `Simplenote MacOS beta` profile and upload for notarization.
4. Verify that the app is successfully notarized.
5. Version `1.10.0.1` on AppCenter has been built out of this branch. Download it and verify that it can be opened on Catalina. 

### Review
Only one developer is required to review these changes, but anyone can perform the review.

### Release


loremattei:
@jleandroperez yes sir! This is that:

> For the same reason, this PR also resigns the Sparkle binaries with the hardened-runtime flag.

Basically, Sparkle was ready to be signed with `hardened-runtime`, but not actually signed. I just resigned your binary, without any other change because it was already compliant.

loremattei:
Hey @jleandroperez sir, this is ready for another look.
As discussed, I replaced the binaries with the unsigned ones and added a step to re-sign them to the build script.

Issue #344, opened 2019-06-28
Tags: Notes list for selected tag disappears after renaming tag


rachelmcr:
### Expected

I expect to be able to rename a tag, and still see a list of the notes with that tag.

### Observed

When I rename a tag, the list of notes with that tag disappears until I select a different tag and reselect the renamed tag.

### Reproduced

1. Open the tags sidebar and select a tag. Notice that the notes list shows notes with that tag.
2. Right-click and select "Rename tag."
3. Enter a new tag name and hit Enter. Result: The notes list is now empty.
4. Select a different tag and then reselect the renamed tag. Result: The notes list shows the list of notes with the renamed tag again.

![simplenote-tag-rename](https://user-images.githubusercontent.com/8658164/60339308-59af7980-99a0-11e9-829b-a866aa8a923b.gif)

MacOS 10.14.5 | 1.5.0 beta
h/t @Robertght for the original report

PR #502, opened 2020-04-16
Refreshing Tags List: Part I

jleandroperez:
### Details:
- Drops **SPTagCellView**: Replaced by **HeaderTableCellView** and **TagTableCellView**
- Replaced (few) legacy assets with their new vectorial versions
- Improved lots of snippets along the way
- And wired new IB metrics, based on the latest mockups

@aerych I'm sooooo sorry about the PR(s) flood!!!
Ref. #458

### Test
- [x] Verify pressing **All Notes** yields all of your notes onScreen
- [x] Verify pressing **Trash** filters out trashed notes
- [x] Verify Tags can be reordered via Drag + Drop
- [x] Verify Tags can be renamed via a Right Click > Rename
- [x] Verify Tags can be deleted via Right Click > Delete
- [x] Verify Tags can be renamed after clicking on them (the TextField must become active)

### Release


Simplenote/TagListViewController.m
```diff
@@ -455,45 +455,27 @@ - (id)tableView:(NSTableView *)tableView objectValueForTableColumn:(NSTableColum
...
-        Tag *tag = [self.tagArray objectAtIndex:row-kStartOfTagListRow];
-        tagView.textField.stringValue = tag.name;
+
+    if (row == kTrashRow) {
+        return [self trashTableViewCell];
     }
-    
-    return tagView;
+
+    if (row == kTagHeaderRow) {
```
aerych, line 466: I'm liking that these conditionals aren't using `else` statements.  It's a little easier to read this way I think.

jleandroperez:
Thanks for spotting those!!

### Issue 1: Notes List empty after rename
I've found a ticket already filed here: #344

Moving that one to the Spring Cleaning backlog (so that we fix it for the next release).

### Issue 2: Editing Tag Name from the Editor
I believe this behavior is on purpose: renaming a Tag from the Note Details screen should actually create a new Tag (if the new name didn't exist before).

In that case, since a Note might have multiple tags, and the Update OP is fired only once you've finished editing, picking up the right Tag that should be preselected is straightforward. And going for All Notes is neutral, you can't go wrong with that!

### Issue 3: Note Detail / Cursor
```
Side note, when I edit the note in the note detail and get the cursor, the alignment of the text shifts. It looks centered normally, but left aligned when editing. Is this a glitch or intended? (It seemed a bit odd to me.)
```

@aerych do you refer to the Tags List in the Editor area, or something else?

Thanks in advance!!

Simplenote/TagListViewController.m
jleandroperez: Now that you mention it... that was one of the first comments Mike sent over, in a code review (circa 2013!).

If the first branch of the conditional ends up in a return, it really makes no sense to add an `else`, since... you avoid the indentation, and the `else` itself doesn't add anything to the logic.

🔥

aerych:
> do you refer to the Tags List in the Editor area,

Yep. The horizontal list at the bottom of the edit area.  
When showing a tag it looks centered like this:
Screen Shot 2020-04-22 at 11 07 42 AM
And then when editing the text left aligns:
Screen Shot 2020-04-22 at 11 07 49 AM
But now I'm seeing that its an existing thing in the production app so no need to worry about it here.

jleandroperez:
Thank you @aerych !!! I was about to file an issue for the tags glitch, buuuut a new UI is in the mockups (for Spring Cleaning), so I'll probably be building that nearly from scratch, soon.

Thankyooou!!!

PR #507, opened 2020-04-20
Analytics: User Defaults

jleandroperez:
### Details:
We're storing the `Analytics Enabled` flag in Simperium. Meaning that right after signing in, if the flag was enabled in your account, it'll be turned on by default.

In this PR we're replacing such beautiful mechanism with a local UserDefaults boolean.

**Why:** As per AAPL's guidelines, a fresh install should always be off by default.

cc @aerych  (Thank you sir!!)

### Test
0. Install `develop`
1. Log into your account
2. Click over `Privacy Settings` and enable analytics
3. Install this branch on top

- [x] Verify `Analytics` are enabled
- [x] Verify that the following snippet shows up in the console: `<> Migrating Preferences flag from Simperium`
- [x] Verify that if you disable `Analytics` and relaunch, the setting sticks disabled

### Release


jleandroperez:
Definitely edgy case!! The main goal of this PR is to... stop getting review issues (AAPL).

Since the flag was being restored right after authentication, in some cases... it appeared as if it was enabled by default (which wasn't really the case).

Thank you for reviewing this one!!!

PR #510, opened 2020-04-21
Refreshing Tags List: Part III

jleandroperez:
### Fix:
In this PR we're fixing a bug that caused the Tag Name to become invisible, on edition.

### Details:
- `NSColor.controlTextColor` is an AppKit provided color, meant for Text onScreen
- Such color has a low level behavior: NSTextField's adjusts automatically, based on the background color
- Using a custom NSColor means loosing the automatic behavior, which ensures the text won't become invisible.

In this PR we're introducing a new **NSTextField** subclass (simply **SPTextField**), capable of automatically switching to a different NSColor, whenever it becomes the first responder.

Nope. There was no standard way to do this 😭 

@aerych last one of the "Tags List" series! yay!!
(Thank you!!!)

Ref. #458

### Test
1. Launch Simplenote
2. Click over any of your tags in the Tags List

- [x] Verify the tag name is visible, on selection
- [x] Verify that the tag is readable after right clicking on it, and selecting **rename**
- [x] Verify that right clicking over any other tag name renders a Menu, and does not cause the field to enter edition mode immediately
- [x] Please repeat for Light / Dark Modes

### Release


Simplenote/TextField.swift
```diff
@@ -0,0 +1,101 @@
...
+//         within `acceptsFirstResponder`.
+//
+protocol TextFieldDelegate: NSTextFieldDelegate {
+    func controlAcceptsFirstResponder(_ control: NSControl) -> Bool
+}
+
+
+// MARK: - TextField
+//
+class TextField: NSTextField {
```
aerych, line 15: SwiftUI has a [TextField](https://developer.apple.com/documentation/swiftui/textfield) struct. I know its scoped by framework, but would it be a best practice to use a different name?

Simplenote/TextField.swift
```diff
@@ -0,0 +1,101 @@
...
+
+    var activeTextColor: NSColor? {
+        if isEditing {
+            return textEditionColor
+        }
+
+        return isSelected ? textSelectionColor : textRegularColor
+    }
+
+    var innerDelegate: TextFieldDelegate? {
```
aerych, line 94: This is only used once outside its definition. Are their future plans for it or would it make sense to drop the definition and just try casting the delegate where its used?

jleandroperez: Great call sir!! Always think big, but start small.

jleandroperez:
- Renamed **TextField** > **SPTextField** to avoid a future SwiftUI collision
- Fixed a rename regression
- Dropped the `innerDelegate` property

@aerych Thank you **SO MUCH!!!**. Ready for another peek!

aerych:
Looks great @jleandroperez :)  I retested all the things.  :shipit:

jleandroperez:
Thank you sir!!

PR #511, opened 2020-04-21
Dropping SPToolbar: Mark 1

jleandroperez:
### Details:
- Moves the **SearchBar** + **NewNote** components inside the **NotesListViewController**
- **Temporarily** dropping all of the Note Editor toolbar buttons. Will be restored later!
- Replaced ivars with real properties in **NoteListViewController**
- Drops **SPToolbarView** from the main window
- Removed all of the AppDelegate <> ToolbarView interactions
- New **NSButton** tinting extension

**Note:** This is a midway PR, to keep things small. The UI is definitely not expected to be complete here.

cc @etoledom sir, may I bug you with this one?
Thanks in advance!!

Ref. #458

### Scenario: Trash
1. Log into your account
2. Trash any note that has content
3. Click over the **Trash** filter

- [x] Verify that clicking over the Editor does nothing (should not accept First Responder)
- [x] Verify that the **Add New Note** button (right hand side of the search bar) is **disabled**
- [x] Verify everything looks great in Light and Dark Modes!

### Scenario: All Notes!
1. Log into your account
2. Make sure you have at least one note
3. Click over **All Notes** (or any tags that contain at least one note)

- [x] Verify that clicking over **Add new Note** causes the app to add a new note
- [x] Verify that the Editor can become first responder

### Scenario: Placeholders
0. Log out from your account
1. Log back in 

- [x] Verify that a spinner shows up in the Notes List while your notes sync
- [x] Verify that while the UI is empty, a placeholder reading **No Notes** shows up

### Release


jleandroperez:
Thanks **A LOT** @etoledom !!! Will review all of the Theme switching glitches in a follow up.

PR #516, opened 2020-04-24
Dropping HockeyApp

jleandroperez:
### Fix
In this PR we're dropping HockeyApp, since we're relying on Sentry for crash reporting (and Sparkle for OTA updates).

@aerych May I bug you with a really REALLY cool code deletion PR?
Thanks in advance sir!

### Test
- [x] Verify the app builds and runs in develop mode
- [x] Verify the app builds and runs in release mode

### Release


jleandroperez:
Thank you Eric!!! Woooo

PR #541, opened 2020-05-21
Refreshing Notes List: Mark 1

jleandroperez:
### Description
In this PR we're rebuilding the Notes List, so that more content fits onScreen.
This PR comes with new assets, and batteries 🔋🔋🔋

cc'ing @frosty for the dark side of things, and @SylvesterWilmott for the design review
**Thank you both in advance!!!**

P.s.: This is a really cool PR!

### Details
- **New Assets:** *Pinned* + *Shared*
- **Replaced:** *SPNoteCellView* >> *NoteTableCellView*
- **Nukes:** Legacy assets 🔥🔥🔥
- **Nukes:** Legacy Note Preview Generation logic 🔥🔥

Ref. #458

### Note
Checklists aren't expected to be rendered in the Notes List.
Already addressed via #543 😄 

### Test
- [ ] Verify that whenever you edit a Note the List gets updated in real time
- [ ] Verify the Notes List reacts super to Dark Mode
- [ ] Verify the Notes List reacts well when toggling Condensed / Regular mode
- [ ] Verify Pinned Notes display a left indicator
- [ ] Verify Published Notes display a right indicator
- [ ] Verify every row's height is the same: (even empty or short notes)

### Release
`RELEASE-NOTES.txt` was updated in 4d316e3 with:

> Your Notes List will now fit more content onscreen.


Simplenote/NoteListViewController+Swift.swift
```diff
@@ -55,3 +55,24 @@ private extension NoteListViewController {
...
+//
+extension NoteListViewController {
+
+    @objc(noteTableViewCellForNote:)
+    func noteTableViewCell(for note: Note) -> NoteTableCellView {
+        note.ensurePreviewStringsAreAvailable()
+
+        let noteView = tableView.makeTableViewCell(ofType: NoteTableCellView.self)
+
+        noteView.displaysPinnedIndicator = note.pinned
```
frosty, line 70: So clean!

Simplenote/NoteTableCellView.swift
```diff
@@ -0,0 +1,161 @@
...
+    ///
+    @IBOutlet private var titleTextField: NSTextField!
+
+    /// TextField: Body
+    ///
+    @IBOutlet private var bodyTextField: NSTextField!
+
+    /// LeftImage: Pinned Indicator
+    ///
+    @IBOutlet private var leftImageView: NSImageView!
```
frosty, line 20: Nitpicks, but if this is always used for pinned and the right is always used for shared, should we just name them `pinnedImageView` and `sharedImageView`?

Simplenote/NoteTableCellView.swift
```diff
@@ -0,0 +1,161 @@
...
+            bodyTextField.isHidden
+        }
+        set {
+            bodyTextField.isHidden = newValue
+        }
+    }
+
+    /// Note's Title String
+    ///
+    var titleString: String? {
```
frosty, line 61: Again nitpicking, but this could just be `title` or `noteTitle`? We know it's a string because of the type :) 

(same for `bodyString` below)

jleandroperez: You're SO RIGHT!!! What was I thinking about!!!

Simplenote/NoteTableCellView.swift
jleandroperez: Done!! MUCH better, thank you!!

PR #547, opened 2020-05-22
Implements Top Dividers

jleandroperez:
### Details:
In this PR we're implementing **top dividers** in the **SearchBar and Toolbar** areas. This new divider will fade in along with the scroll position.

@danielebogo Sir, do you think I could bug you with another one? (LOW PRIO!!!)
**Thanks a lot!!!**

Ref. #458 

### Notes:
- We cannot implement gradual fade in as you scroll into Markdown Preview, because... WKWebView does not expose a NSScrollView instance, as per macOS 10.15.
- Boooooo.

### Test
- [x] Verify the SearchBar's divider shows up at launch
- [x] Verify the Editor gradually fades in a Top Divider, as you scroll the document
- [x] Verify this works super in Dark / Light Mode!
- [x] Verify that toggling Markdown Preview causes the top divider to immediately show

### Release


Simplenote/NoteEditorViewController+Swift.swift
```diff
@@ -100,10 +122,47 @@ extension NoteEditorViewController {
...
+                                               name: NSView.boundsDidChangeNotification,
+                                               object: scrollView.contentView)
+    }
+
+    @objc
+    func clipViewDidScroll(sender: Notification) {
+        refreshTopDividerAlpha()
+    }
+
+    func refreshTopDividerAlpha() {
```
frosty, line 149: Nitpick, but I think this and the following var could be private.

jleandroperez: **Definitely** private. Thank you!!

jleandroperez:
Thank you @danielebogo and @frosty !!! (You both are aaawesome!!)

PR #548, opened 2020-05-27
Interface: Adjusting Metrics

jleandroperez:
### Details:
In this PR we're fine tunning the top area's metrics:

- **Toolbar:** 48pt height
- **SearchBar:** 26pt height

Plus we're making the Window "fullContentSize", and wiring `contentLayoutGuide` to deal with Fullscreen <> Windowed transitions.

@aerych Sir!! May I bug you with this one?
I swear figuring out all of the AppKit workarounds took around 7 hours each.

Ref. #458

### Test
- [x] Verify the app looks great!
- [x] Verify that entering / leaving fullscreen preserves the Tag List content insets
- [x] Verify that Markdown Preview does not look cut off
- [x] Run a smoke test on the top area interface (Search Bar / Toolbar buttons)

### Release

## AppKit Workarounds
This PR contains two major workarounds for layout issues, which took a massive amount of time and blood to figure out.

Both are documented in the code itself, and adding some extra details, for future reference.

### Workaround A: Markdown Preview
**Problem:**
>WebKit assumes its **origin.y** to always **be zero** whenever it's embedded in a NSWindow with the **fullContentSize** mask.

>This causes the rendered content to be literally cut off. As per macOS 10.15, the internal ScrollView is no longer exposed, so there is no public way to fix this.

**Solution:** 
> We've wrapped the WebView inside a NSScrollView (documented in **MarkdownViewController**). This shields us from the effects of NSWindow's contentLayoutRect.

### Workaround B: Tags List Additional Insets
**Description:**
> Displaying a Title Bar means that switching from Windowed <> Fullscreen modes requires updating the contentInsets (increasing / decreasing).

> NSScrollView / NSClipView provides a mechanism to deal with this:`automaticallyAdjustScrollviewInsets`. 

**Problem:** 
> We need to inject "_Additional contentInsets_" in order to match the Tags / Notes top padding.

**Solution:** 
> We've subclassed `ClipView`, and implemented a mechanism resembling the one found in **UIScrollView** (`adjustedContentInset` / `contentInset`).


jleandroperez:
Thank you Eric!!!

PR #578, opened 2020-06-30
Share Popover Style

jleandroperez:
### Details:
This PR adjusts the Share Popover, so that it reacts to NSAppearance changes.

@etoledom Sir!! May I annoy you, yet again?
Thanks in advance!!

Ref. #458 

### Test
1. Launch Simplenote
2. Click over the top right Icon `( ... )` > `Collaborate`

- [x] Verify a Popover shows up at the bottom center of the screen
- [x] Verify that switching to Light / Dark theme, while the popover is onScreen immediately causes the Popover to refresh (and match the new style!)

### Release


jleandroperez:
Woooo **really interesting** scenario!!! 

To us the NSPopover instantiation is 100% handled by NSViewController's API, I wonder if the SDK is internally reusing the NSPopover instance, and not recalculating its position.

Will try to address that one in a follow up PR. **Thanks a lot @etoledom !!!!**

PR #582, opened 2020-07-01
Merge hotfix 1.11.1 into develop

mokagio:
This should be merged only after Apple approves the build and we ship it.


mokagio:
I was late on marking this as ready to review because GitHub had a different opinion than my local on whether there were conflicts on the release notes file.

My local was able to automerge, but on GitHub's end there were still conflicts. 

I changed the file manually adding a new line that was due anyway between the last line of 1.12 and `1.11.1` and now everything seems fine.

PR #588, opened 2020-07-03
RTL Support: Tags Editor

jleandroperez:
### Fix
In this PR we're fixing the way the Tags Editor renders in a RTL environment.

@aerych May I bug you with a really cool PR?
Thanks in advance sir!!

Ref. #458 

### Details:
1. Fixed **TagAttachmentCell**'s text frame.width, so that the text is perfectly centered
2. **TagsEditor / IB:**  forced **LTR** and enabled **Mirror when in RTL** 

### Test
1. In Xcode
  A. Click **Edit Scheme** > **Options**
  B. Set **Application Language = Right to Left**
2. Launch Simplenote + Log In

- [x] Verify the Tags Editor is pinned to the right hand side of the screen 🔥

### Release
`RELEASE-NOTES.txt` was updated in f71c7f4 with:
 
> Fixed Tags Editor RTL support

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -5,6 +5,7 @@
 - Implements Untagged Notes Filter
 - New Note Metrics Interface
 - Relocated Editor Actions into the new More Actions Menu
+- Fixed Tags Editor RTL support
 
 1.11.0
 ------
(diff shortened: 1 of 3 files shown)


jleandroperez:
Thank you Eric!! Wooooo!!

PR #594, opened 2020-07-13
Updates release notes

jleandroperez:
### Details:
In this PR we're dropping RTL references from the release notes, since the app doesn't support Persian localizations (yet, that's our next project!)

@aerych May I bug you with a really really super quick one?
Thanks in advance!!

### Test
- [x] Verify the CI is green!

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -5,7 +5,6 @@
 - Implements Untagged Notes Filter
 - New Note Metrics Interface
 - Relocated Editor Actions into the new More Actions Menu
-- Fixed Tags Editor RTL support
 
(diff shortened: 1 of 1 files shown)


jleandroperez:
Thank you Eric!!

jleandroperez:
/cc @mokagio Sir!! no need for a new build, but we could use new Release strings!

Thank you!!

mokagio:
>  but we could use new Release strings!

That's ok, I think. We send the info to App Store Connect only when we wrap up the release, so no harm done. I'll see if I can update App Center and TestFlight now.

jleandroperez:
Thank you Gio!!!

PR #597, opened 2020-07-14
Fix: Negative word count

### LINKED ISSUE #596: Note metrics show -1 Words for new note
### Expected
When I create a new note, I expect the note metrics to show the number of words as 0.

### Observed
The note metrics show the number of words as -1.

### Reproduced

1. Create a new note.
2. Without adding any content, click the (i) icon in the top right.
3. In the note metrics popover that opens, notice that Words is -1.

Screenshot 2020-07-14 14 30 54

Make|Model|macOS Version|App Version
-|-|-|-
Apple|Macbook Pro|10.15.5|1.12.0 beta

jleandroperez:
### Fix
In this PR we're fixing a glitch that caused negative word counts to show.

@danielebogo Sir!! May I bug you with a really quick PR?
Thanks in advance!!

Closes #596

### Test
1. Add a new (and empty) Note
2. Click over the (i) icon

- [ ] Verify the word count is zero!

### Release

### DIFF
--- a/Simplenote/NoteMetrics.swift
+++ b/Simplenote/NoteMetrics.swift
@@ -28,9 +28,10 @@ struct NoteMetrics {
     init(notes: [Note]) {
         let contents = notes.compactMap({ $0.content }).reduce("", +)
         let dateProviderNote = notes.count == 1 ? notes.first : nil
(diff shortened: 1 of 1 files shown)


jleandroperez:
Thank you Daniele!!

PR #601, opened 2020-07-15
Localization Script!

jleandroperez:
### Details:
In this PR we're replicating the work done by @jkmassel [in Simplenote iOS](https://github.com/Automattic/simplenote-ios/pull/535).

@jkmassel since this is your kid, may I bug you with a fun review?
Thanks in advance!!

**P.S:** Apologies about the size of the PR, we've added Xcode support for the same list of languages allowed in iOS, and things went out of control!

/cc @mokagio first one out!

Ref. #8

### Test
1. Edit [this line](https://github.com/Automattic/simplenote-macos/blob/develop/Simplenote/Note.m#L186)
2. Run `./Scripts/localize.py` on the root directory of the project

- [x] Verify the new string is picked up!

### Release


jleandroperez:
Thank you @mokagio !!!

PR #605, opened 2020-07-22
NIB(s) Localization Support

jleandroperez:
## Description:
In this PR we're extending the Localization toolchain in order to support Xcode's standard NIB localization mechanism.
This translates into:

1. Support for **strings extraction** from `MainMenu.xib`
2. Support for **downloading MainMenu strings**
3. Enhanced failsafe mechanism to deal with missing GlotPress Strings

@jkmassel Sir, can I bug you with a fun review?

/cc @mokagio @loremattei 

Thank you all!

Ref. #8 

### localize.py:
This script now also extracts strings from **MainMenu.nib**, and converts them into UTF-8.

### update-translations.rb
We're now **also** downloading strings from the [Main Menu](https://translate.wordpress.com/projects/simplenote/macos/main-menu/) GlotPress subproject.

### fix-translations:
Missing translations, when downloaded from GlotPress, would look like this:

> "Some Text" = "";

The script `fix-translations` works by setting a `Translation.Value = Translation.Key`, so that the app doesn't display a blank value:

> "Some Text" = "Some Text";

#### Problem: 
NIB translation keys don't follow the same principle. Instead, they contain keypaths:
> "58.title" = "Some Title";

The side effect is: missing NIB translations (in GlotPress) look like this:
> "58.title" = "";

Running the standard `fix-translations` script would yield:
> "58.title" = "58.title";

And this is definitely not human readable 😞 

#### Solution:
In this PR we're enhancing `fix-translations`, so that the actual `fallback value` is extracted from a "Fallback Translations Map". That means: the script will extract the proper "Fallback Translation" from another file.

--

## Test: Localize!
1. Checkout commit 763e2de
2. Run `./Scripts/localize.py`

- [x] Verify `Simplenote/en.lproj/Localizable.strings` has an updated string
- [x] Verify `Simplenote/en.lproj/MainMenu.strings` has an updated string

## Test: Update Strings
1. Run `./Scripts/update-translations.rb`

- [x] Verify `Simplenote/pt-BR.lproj/MainMenu.strings` gets new Portuguese strings (an OSS contributor already did this! yay!)
- [x] Verify (lots of) `Localizable.strings` files get updated
- [x] Verify that `MainMenu.strings` files **don't contain empty Localizations.** Strings that aren't in Glotpress should carry the English localization!

## Release


Scripts/update-translations.rb
```diff
@@ -86,20 +90,35 @@ def copy_comment(f, trans_strings, value)
...
+
 langs.each do |code,local|
   lang_dir = File.join('Simplenote', "#{local}.lproj")
   puts "Updating #{code}"
   system "mkdir -p #{lang_dir}"
   
-  # Backup the current file
+  # Backup the current file(s)
   system "if [ -e #{lang_dir}/Localizable.strings ]; then cp #{lang_dir}/Localizable.strings #{lang_dir}/Localizable.strings.bak; fi"
+  system "if [ -e #{lang_dir}/MainMenu.strings ]; then cp #{lang_dir}/MainMenu.strings #{lang_dir}/MainMenu.strings.bak; fi"
```
mokagio, line 102: Suggestion for the future, this could also be done via the Ruby `File.exists?` and `FileUtils.copy` APIs.

Scripts/fix-translation.swift
```diff
@@ -0,0 +1,95 @@
...
+///
+func replaceEmptyTranslations(in filename: String, fallbackMap: [String: String]) throws {
+    var encoding = String.Encoding.utf8
+    let contents = try String(contentsOfFile: filename, usedEncoding: &encoding)
+    var output = ""
+
+    contents.enumerateLines { line, _ in
+        let fixed = stringByFixingEmptyValue(in: line, fallbackMap: fallbackMap)
+        output.append(fixed)
+        output.append("\n")
```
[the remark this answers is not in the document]
jleandroperez: `enumerateLines` is actually not passing the `\n` for each line, so we're good (we need it!).

**Thank you** for checking!!!

Scripts/update-translations.rb
jleandroperez: Thank you sir!!! taking note, fixing in a follow up.

jleandroperez:
Thanks **a million** for reviewing this PR @mokagio !!!

PR #612, opened 2020-07-24
Fastlane: Finalize Release

jleandroperez:
### Description:
In this PR we're enabling support for localizations in Fastlane's `finalize_release` step. This means that:
- We'll download Metadata (AppStore Strings!)
- We'll also download translated Localizable.strings
- And as per PR #605, we'll **also** download the Main Menu subproject translations.

Ref. #8

/cc @mokagio @loremattei 
Thanks in advance gentlemen!!

### Details:
- `VERSION_LONG` has been patched to match [WooCommerce style](https://github.com/woocommerce/woocommerce-ios/blob/develop/config/Version.Public.xcconfig#L4). This was causing issues when updating the `Metadata` GlotPress Key.
- Fastlane: `finalize release` now runs `ios_update_metadata` as well!
- We're bringing [download_metadata.swift](https://github.com/Automattic/simplenote-ios/blob/develop/fastlane/download_metadata.swift) from Simplenote iOS (with its URL patched!)

### Test
Testing this PR is quite tricky, because the actual scritps perform **git push**. 
For that reason, we'll need to disable few things in your local checkout. **Please:**

1. Locate **ios_git_helper.rb** in your local checkout
   - Comment [this commit OP](https://github.com/wordpress-mobile/release-toolkit/blob/bb668d63f550cb50f675470ee6a7b03ed8fb5419/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb#L99) 
   - Comment [this other commit OP](https://github.com/wordpress-mobile/release-toolkit/blob/bb668d63f550cb50f675470ee6a7b03ed8fb5419/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb#L103)
   - Comment [this push OP](https://github.com/wordpress-mobile/release-toolkit/blob/bb668d63f550cb50f675470ee6a7b03ed8fb5419/lib/fastlane/plugin/wpmreleasetoolkit/helper/ios/ios_git_helper.rb#L105)
2. Open your `fastlane/Fastfile`
   - Comment [this line](https://github.com/Automattic/simplenote-macos/pull/612/files#diff-56c5021ef06ac6cea807ae53269b6f4fR148) (disables `finalize_prechecks`!)
   - Comment [this line](https://github.com/Automattic/simplenote-macos/pull/612/files#diff-56c5021ef06ac6cea807ae53269b6f4fR150) (disables `ios_final_tag`)
3. **Turn off your wifi!!!**
4. Run `bundle exec fastlane finalize_release`

- [ ] Verify the following files were updated in your local repository, for every language:
  - [ ] **keywords.txt**
  - [ ] **description.txt**
  - [ ] **promotional.txt**
  - [ ] **release.txt**
  - [ ] **Localizable.strings**
  - [ ] **MainMenu.strings**

### Release


jleandroperez:
@mokagio Important detail about this PR is: [I've manually patched this constant](https://github.com/Automattic/simplenote-macos/pull/612/files#diff-c643f223fe693397aca8e6976a79485aR5) (otherwise the script that updates the "Download Release Notes from GlotPress" would fail).

I haven't found where is it, exactly, that `VERSION_LONG` is updated. Is that a manual step?

Thank you!!!

fastlane/download_metadata.swift
```diff
@@ -0,0 +1,129 @@
...
+        }
+
+        var subtitle: String?
+        var whatsNew: String?
+        var keywords: String?
+        var storeDescription: String?
+
+        jsonDict.forEach({ (key: String, value: Any) in
+
+            guard let index = key.firstIndex(of: Character(UnicodeScalar(0004))) else {
```
mokagio, line 67: Nice touch upgrading this to use `firstIndex(of:)` instead of the deprecated `index(of:)`.

config/Version.public.xcconfig
```diff
@@ -1,5 +1,5 @@
-VERSION_SHORT=1.12.0
+VERSION_SHORT=1.12
 
 // Public long version example: 1_12_0_1. The last 1 means the first build for
 // this version
-VERSION_LONG=11202
+VERSION_LONG=1.12.0.2
```
mokagio, line 5: I worry this change won't go well with Apple 😞 A way to test this would be to ship a new build to TestFlight.

In https://github.com/Automattic/simplenote-macos/commit/f174a5b4afd66fc59f983448780af788e9808f10 I reverted a similar change because App Store Connect rejected the build with message

> This bundle is invalid. The value for key CFBundleVersion
> [1.11.1.0] in the Info.plist file must contain a higher version than
> that of the previously uploaded version [11103].

I wonder if we may need to bump the build number even more, i.e. adding another leading 1? But maybe that wouldn't work with GlotPress?

> This was causing issues when updating the Metadata GlotPress Key.

It might be that we need to do some extra work on the plugin side.

mokagio: @jleandroperez 

> I haven't found where is it, exactly, that VERSION_LONG is updated. Is that a manual step?

Yes, it's a manual step right now 😳

fastlane/download_metadata.swift
jleandroperez: Thank you! 😃

jleandroperez:
@mokagio Missing language files would be expected (since we do have a few missing translations).

Okay, I've been over this issue for a while: 

```
This bundle is invalid. The value for key CFBundleVersion [1.11.1.0] in the Info.plist file must contain a higher version than that of the previously uploaded version [11103].
```

I'm thinking that... rather than changing the Release Toolkit, so that it accommodates this scheme, the **CLEANEST** way to address this is bumping up the major version to 2.0.0.0. That should fix the string comparison, in a single shot.

Since this next release adds localizations (and completed the UX overhaul), it'd be appropriate.

WDYT?

jleandroperez:
@mokagio checked with the team, okay with bumping up the release version.

This should also fix the `codefreeze` script that bumps up the release number (I believe!!). Please let me know if everything looks good on your end.

Reason why I'm thinking `2.00.0.0` version would work is that... I've tried comparing both strings in Swift, and indeed `11103 < 2.00.0.0` evaluates as **true**.

As far as I can tell, the comparison appears to happen based on the first digit, so it should be good.

**Thank you!!**

config/Version.public.xcconfig
```diff
@@ -1,5 +1,5 @@
-VERSION_SHORT=1.12.0
+VERSION_SHORT=2.00
```
mokagio, line 1:
```suggestion
VERSION_SHORT=2.0
```

config/Version.public.xcconfig
```diff
@@ -1,5 +1,5 @@
...
 
-// Public long version example: 1_12_0_1. The last 1 means the first build for
+// Public long version example: 2.0.0.1. The last 1 means the first build for
 // this version
-VERSION_LONG=11202
+VERSION_LONG=2.00.0.0
```
mokagio, line 5:
```suggestion
VERSION_LONG=2.0.0.0
```

jleandroperez:
@mokagio RIGHT!!! What was I thinking!!! 

I hope this change makes all the things go smooth (in theory `VERSION_LONG` will be bumped by `fastlane code_freeze` AFAIK right?).

Ready on your mark. Thank you Gio!! Wooooo 2.0 !!!

jleandroperez:
You got it ... that's **PRECISELY** where I took **Mark** from 😄 

Thank you Gio!!!

PR #613, opened 2020-07-24
Localization Script: Picking up Simperium Strings

jleandroperez:
### Fix
In this PR we're patching up the `Localize` script, so that Simperium Auth Strings are picked up.

@aerych sir!! May I bug you with this fun PR?
Thanks in advance!!

Ref. #8

### Test
1. Run `./Scripts/localize.py`

- [x] Verify the following file has changes `Simplenote/en.lproj/Localizable.strings`
- [x] Verify that the new strings are Auth-Y

### Release

### DIFF
--- a/Scripts/localize.py
+++ b/Scripts/localize.py
@@ -132,7 +132,7 @@ def localize_sources(path):
     old = original + '.old'
     new = original + '.new'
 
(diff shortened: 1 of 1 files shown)


jleandroperez:
Thank you sir!!

PR #617, opened 2020-07-28
Tags Validation

### LINKED ISSUE #616: Tags: Validation
### Details:
Every platform should validate tags with the following regex:

```
(.){1,256}
```

This collapses into: Any character is allowed (except whitespaces) / maximum string length = 256.

[iOS Sibling](https://github.com/Automattic/simplenote-ios/issues/823)

jleandroperez:
## Details:

- New standarized simperiumKey for the Tag Entity
- New 256 characters limit for (encoded) Tag Names
- Failed Tag Rename OP(s) cause the UI to refresh. So that all the things stay in sync!

cc @aerych Sir, you game for yet another Tags-Y PR?
Thank you!!!

Closes #616

---

## Scenario: Note Editor
1. Checkout commit d930280
2. Open any note
3. Click over the Tags Editor

- [x] Verify that new tags can have at most 15 characters (**this is for testing!** actual limit is 256 char)
- [x] Verify that special characters are allowed (parentheses, comma)
- [x] Verify that pressing a space "commits" the text you've typed

> **IMPORTANT:** Length is enforced against **encoded** tag name. Using non alphanumeric characters will yield a much lower limit!

## Scenario: Tags Editor
1. Checkout commit d930280
2. Right click over any of your tags (in the Tags List)

- [x] Verify that pressing **Delete** effectively removes the new Tag
- [x] Verify that pressing **Rename** allows you to edit the Tag Name (**up to 15 characters**, for testing purposes!)
- [ ] Verify that during a rename OP pressing spacebar "Commits" the edit
- [ ] Verify that you can change (any random) character in a Tag's name, and swapt it for its lowercase / uppercase version

> **IMPORTANT:** Length is enforced against **encoded** tag name. Using non alphanumeric characters will yield a much lower limit!

## Scenario: Tags Editor / Collisions
1. Create `Tag A`
2. Create `Tag B`
3. Right click over `Tag B`

- [x] Verify that the app refuses to let you rename `Tag B` as `Tag A` (falls back to the original String, the second you try to "commit"

## Release
`RELEASE-NOTES.txt` was updated in 18480bf with:
 
> Updated Tags limit to up to 256 characters


Simplenote/NSString+Simplenote.swift
```diff
@@ -5,6 +5,20 @@ import Foundation
...
+    @objc
+    var byEncodingNonAlphanumerics: String? {
+        addingPercentEncoding(withAllowedCharacters: .alphanumerics)
+    }
+
+    /// Returns a Percent Encoded / lowercased and UTF normalized version of the receiver
+    ///
+    @objc
+    var byEncodingForTagSimperiumKey: String {
+        precomposedStringWithCanonicalMapping.lowercased().byEncodingNonAlphanumerics ?? self as String
```
dmsnell, line 19: this is cool, though would it not be helpful to share some terms with the other apps concerning this transformation?

we could call this `asTagHash` or just `tagHash` or `tagHashOf`

I also find that separating `byEncodingNonAlphanumerics` and its corresponding test a bit fluffy, exposed externally when they are purely internal details to the tag hash algorithm. what's the purpose of separating them?

in any case, we don't really care if `byEncodingNonAlphanumerics` does that - we care if our tag names collapse to the appropriate cross-platform tag hash and so I would have expected any test of this behavior to test that instead of an internal implementation detail

Simplenote/TagListViewController.m
```diff
@@ -182,7 +182,7 @@ - (void)loadTags
 - (Tag *)tagWithName:(NSString *)tagName
 {
     for (Tag *tag in self.tagArray) {
-        if ([tag.name isEqualToString:tagName]) {
+        if ([tag.name compare:tagName options:NSCaseInsensitiveSearch] == NSOrderedSame) {
```
dmsnell, line 185: this should be comparing the tag hashes and not raw textual comparisons because it has to normalize Unicode variations and also casing is locale-dependent and we want to make sure that our tag name collisions aren't dependent on which language is set for a given device

```objectivec
- (Tag *)tagWithName:(NSString *)tagName
{
	NSString *tagHash = [tagName asTagHash];
	for (Tag *tag in self.tagArray) {
		if ([tag.name.asTagHash isEqualToString:tagHash]) {
			…
```

SimplenoteTests/NSStringSimplenoteTests.swift
```diff
@@ -6,6 +6,17 @@ import XCTest
...
+    /// Verifies that `byEncodingNonAlphanumerics` effectively escapes all of the non alphanumeric characters
+    ///
+    func testByEncodingNonAlphanumericsPercentEncodesAllOfTheNonAlphanumericCharactersInTheReceiver() {
+        let sample = "1234567890!@#$%^&*()-_+[]';./,qwertyuiopasdfghjkl;'zxcvbnm,./🔥😂😃🤪👍🦆🏴‍☠️☝️😯"
+        let escaped = sample.byEncodingNonAlphanumerics ?? ""
+        let escapedSet = CharacterSet(charactersIn: escaped)
+        let expectedSet = CharacterSet(charactersIn: "%").union(.alphanumerics)
+
+        XCTAssertTrue(expectedSet.isSuperset(of: escapedSet))
+    }
```
dmsnell, line 18: noting that this test is mostly asserting a behavior of a built-in function which presumably already has rock-solid and comprehensive tests

jleandroperez: Oh yeah, this was mostly for myself (to ensure... the API does what I intended to actually do!), rather than manually verifying 😄

Simplenote/TagListViewController.m
jleandroperez: Tags themselves come from the backend 100% (and AFAIK aren't really localized on the device, thus, independent from the device language). 

Considering that, I'm not sure if we **need to** compare encoded tags (since the encoded string is generated starting off with the tagname we're checking already). Am I missing something? (and thank you for stepping by D!!)

Simplenote/NSString+Simplenote.swift
jleandroperez: Mmmmmmmmmmm I don't like the name `TagHash` honestly. I can't tell why, though!! 😂 

Now, strictly speaking, this API does behave like a hashing method (unless I'm missing something big, mapping should always be 1:1). And sharing terminology could be useful eventually.

I'll get this one updated, thank you for the suggestion!

dmsnell: > I don't like the name TagHash honestly. I can't tell why, though!! 😂

If you come up with a better one we can all harmonize on it. I was looking for something descriptive and terse.

> this API does behave like a hashing method

That's exactly what it's doing.

> unless I'm missing something big, mapping should always be 1:1

Almost every hashing routine will have collisions by design since hashing is a kind of collapsing of a dimension of data. If this weren't the case our MD5 and SHA hashes would have to have an infinite number of bits since they hash every possible sequence of bytes into 16 or 20 or so bytes.

a zero-collision mapping is called _perfect_ hashing while a 1:1 mapping is a _minimal perfect_ hash. these are rare but used when the input domain is finite and static and we really want to have practically instant lookup for the keys or when we need to save as much storage space as possible.

Simplenote/TagListViewController.m
dmsnell: > aren't really localized on the device, thus, independent from the device language

the tag names aren't localized but the string comparison function _is_ and for non-English languages might return different results when comparing two strings than it would have if the device were set to English because casing rules are locale-dependent

> Considering that, I'm not sure if we need to compare encoded tags

if we want to properly differentiate tags or tag names then we have to compare the hash because there are several possible and likely ways that two tag names could be equivalent based on our rules: differences in Unicode representation as well as casing.

we might have _duplicate_ tag entries in the `tag` bucket and so imagine three tag entities erroneously exist with `name` values of the following: (copy/paste is transforming the characters so I've added the code points)
 - `ṩ` (U+0073 U+0323 U+0307) URL-encoding `s%CC%A3%CC%87`
 - `ṩ` (U+0073 U+0307 U+0323) URL-encoding `s%CC%87%CC%A3`
 - `ṩ` (U+1E69) URL-encoding `%E1%B9%A9`

these are all different characters and will compare differently but only one is the right tag, the one with the `%E1%B9%A9` hash. in this example we can see how performing a scan for `name` matches can propagate the errors forward in the "fixed" applications because the overlook the hash. it could very well be that the property tag entity is missing due to previously-erroneous renames. in such a case we can consider a corrective measure to try and find a tag with a matching name and then create a new entity for it, but that does introduce the same confusing complexity that lookup was offering

Simplenote/NSString+Simplenote.swift
jleandroperez: And now I'm terribly curious as per whether we can find a way to make this API collision!

Simplenote/TagListViewController.m
jleandroperez: Actually, [isEqualToString](https://developer.apple.com/documentation/foundation/nsstring/1407803-isequaltostring) is not localized. 

Localized comparison API(s) make it explicit in their names (for instance [localizedCompare:](https://developer.apple.com/documentation/foundation/nsstring/1416999-localizedcompare)).

```
The comparison uses the canonical representation of strings, which for a particular string is the length of the string plus the UTF-16 code units that make up the string. When this method compares two strings, if the individual Unicodes are the same, then the strings are equal, regardless of the backing store.
```

I'll try to build a collision in a unit test, nevertheless!

jleandroperez: Update: I've added this to a unit test, for posterity.

Indeed, **NSString** behaves **exactly** as you've outlined. However, Swift's equality does handle all the things. Thank you Dennis!!

aerych:
Hola @jleandroperez 👋  A few things I noticed while testing:

I was not able to add more tags to notes that already had tags, unless the first tag was very small and the second tag was very small as well. Seemed like the character limit was applied to the total number of tags on a note? Try this:
Create a new note. Add a tag "A", then add a tag "B". Try to add a tag "C". 🚫 

### NOTE EDITOR

> Verify that new tags can have at most 10 characters

Sort of. If I only used a-zA-z0-9 then yes. If I added any special characters then the character count was less.  Try these examples:
b[][
b[]asd
b<><
!@#

This was true in the Note Editor and in the Tag List.

### TAG EDITOR

> Verify that pressing Rename allows you to edit the Tag Name

Same behavior with less than 10 characters when entering special characters. Otherwise this worked.

> Verify that during a rename OP pressing spacebar "Commits" the edit

I could make "Tag A" with a space by renaming an existing tag in the tag list. However, in the Note Editor, the tag showed up without the space "TagA".  So it looks like the UI lets me type spaces but they are being sanitized.

Back to you sir!

jleandroperez:
### Bug: Unable to add more tags
This one was caused by String Attachments present in the validation string (which belong to the other "accepted" Tags).

Fixed now, nice catch!!!

### Bug: Special Characters / Tag Length
Expected (**totally forgot to mention in the test steps!!!** I've edited the issue description).

TL;DR: Tag Names are used to generate the `Tag` Entity's **simperiumKey** (which has a limit of 256 characters). For that reason... we need to cap the maximum length (after encoding). 

### Bug: Tags with Spaces
I couldn't reproduce this issue directly (definitely doing something different). **BUT** I've added a guard, to make super sure spaces are not accepted 😄 


---

Issue description updated, may I bug you with another quick look?

**Thank you sir!!!!**

aerych:
> Bug: Unable to add more tags

Confirmed fixed 👍 

> Bug: Special Characters / Tag Length

Ahh... Gotcha.  I had assumed each individual tag could be up to 256 characters and assumed it was for robust support for multi-byte character sets. 

> Bug: Tags with Spaces

Sent a demo recording to show what I was seeing :)  Only this one left.  Code looks good to me otherwise!

Simplenote/TagListViewController.m
```diff
@@ -473,30 +481,33 @@ - (void)menuDidClose:(NSMenu *)menu
...
-        if ([newTagName length] > 0 && !tagAlreadyExists && ![self.tagNameBeingEdited isEqualToString:newTagName])
-            [self changeTagName:self.tagNameBeingEdited toName:newTagName];
-        
-        self.tagNameBeingEdited = nil;
+    // Force de-selection of text
+    NSTextView *textView = notification.fieldEditor;
+    [textView setSelectedRange:NSMakeRange(0, 0)];
+
+    // Send a *COPY* of the string. Otherwise the internal string will be exposed, and this may lead to
+    // weird side effects.
```
dmsnell, line 499: like what? good comment, would love an example weird side effect

jleandroperez: DM'ing you a link to the original issue, since it's in the private repository!

Simplenote/TagListViewController.m
```diff
@@ -473,30 +481,33 @@ - (void)menuDidClose:(NSMenu *)menu
...
+    NSString *newTagName    = [textView.string copy];
+    BOOL newTagNotFound     = [self tagWithName:newTagName] == nil;
```
dmsnell, line 501: this feels like we're in the rename logic. if that's the case we should note that it's possible for the new tag to exist because of the tag-hash collapsing.

it's legitimate to change the casing of a tag: rename `markdown` to `Markdown`

in those cases `[self tagWithName:@"Markdown"]` will return the tag with the name `markdown`

to get around this we probably need to check not if the new tag exists but if it's different than the existing one

```objective-c
Tag *oldTag = [self tagWithName:self.tagNameBeingEdited];
Tag *newTag = [self tagWithName:newTagName];
BOOL isProperRename = newTag == nil || newTag == oldTag;

if (oldTagWasChanged && isProperRename && newTagName.length > 0) {
	…
```

Simplenote/TagListViewController.m
dmsnell: idea:

```
// Send a *COPY* of the string to avoid "NSLayoutManager: Loop / Crash"
```

Simplenote/TagListViewController.m
jleandroperez: TOTALLY MISSED that scenario!!! ty!!

jleandroperez:
@aerych last one should be good!!! **Thank you!!!!**

Simplenote/TagListViewController.m
jleandroperez: Sold!

jleandroperez:
Thank you Eric!!!

PR #624, opened 2020-07-31
Tags List: Prevents selecting "Unselectable" cells

jleandroperez:
### Fix
In this PR we're fixing a bug in which "Unselectable" cells might end up highlighted.

@danielebogo Sir!! can I bug you with a really simple fix?
Thanks in advance!!

### Test
1. Launch Simplenote
2. Locate, in the left hand side panel, the `TAGS` header  (you must have at least one tag
3. Right click over the `TAGS` header

- [x] Verify the `TAGS` header does not become highlighted.

### Release


jleandroperez:
Thank you sir!!

PR #623, opened 2020-07-31
TagListState: rowAtIndex bounds check

jleandroperez:
### Fix
In this PR we're fixing a bounds check, responsible of a crash when dealing with right click events.

@aerych You game for an old school Out of Bounds fix?
Thank you sir!!

Closes #619

### Test
1. Log into Simplenote
2. Right click (anywhere) in the Tags List's **empty area** at the bottom
 > Yes, the trick is to right click in an empty area!

- [x] Verify Simplenote doesn't crash!

### Release
`RELEASE-NOTES.txt` was updated in d287a51 with:
 
> Fixed a crash while handling right clicks in the Tags List

PR #630, opened 2020-08-04
Bugfix: All Notes AWOL

jleandroperez:
### Fix
In this PR we're fixing a bounds check that caused the `All Notes` row to go AWOL.

This is a side effect of #623, and I totally missed the glitch!!

cc @aerych Sir, can I bug you with a quick fix?
Thank you!!

### Test
- [x] Verify the `All Notes` row is always visible
- [x] Verify you can click anywhere else, and back into `All Notes`

### Release


jleandroperez:
I broke it, so it's on me sir!! thanks for the quick review!!!

Issue #626, opened 2020-08-03
Master Issue: Advanced Search


jleandroperez:
### Details:
- [ ] `tag:` operator support
- [ ] Sort Mode dropdown
- [ ] Intermix Tags / Notes results

### Mockups
<img width="300" src="https://user-images.githubusercontent.com/1195260/89207209-c7a97580-d590-11ea-91ff-8722040b380b.png">
<img width="300" src="https://user-images.githubusercontent.com/1195260/89207218-cb3cfc80-d590-11ea-9899-4d2f2a6cd5cc.png"?
<img width="300" src="https://user-images.githubusercontent.com/1195260/89207222-cc6e2980-d590-11ea-809a-8443139f9d42.png">

riverLethe9:
Issue #33  

How is the addition of 'sort by: date created' going?

jleandroperez:
@riverLethe9 We'll be implementing a new sort bar (which will allow you to sort search results), but no specific ETA at the moment. This particular project will ship early 2021.

Thank you for reaching out!

mokagio:
Bumped this to the 2.11 milestone because there are some tasks are still unchecked and today I'm running the 2.10 code freeze.

PR #632, opened 2020-08-05
Editor: Decoupling from Notes List

jleandroperez:
### Details:
Several EditorViewController improvements / tech debt tasks:

1. Encapsulation: *Note Editor* **NO LONGER** has IBOutlet(s) referencing to components owned by the Notes List 🔥
2. Delegation: New `EditorControllerNoteActionsDelegate` and `EditorControllerTagActionsDelegate` protocols
3. We've also killed direct access to NoteListViewController via the AppDelegate outlet

@aerych The codebase is gonna end up beautiful. Promise!!
Thanks in advance!!

Ref. #626

### Test
- [x] Verify that editing a document causes the Notes List to get immediately updated
- [x] Verify that Pinning a note causes the Notes List to get immediately updated
- [x] Verify that Adding a New Note causes the new entry to get selected in the Notes List
- [x] Verify that the **Search Field** is cleared out after pressing the `+` (New Note) button
- [x] Verify that deleting a note via the `... > Move to Trash` action preserves the selected row index (it shouldn't jump to the first row!) 
- [x] Verify that deleting a note via the `Delete` keyboard key preserves the selected row index (it shouldn't jump to the first row!) 
- [x] Verify that restoring a note from Trash causes the Notes List to immediately refresh
- [x] Verify that after adding a new Tag from the Notes Editor (bottom area), the new row shows up in the Tags List (left sidebar!)

### Release

### DIFF
--- a/Simplenote.xcodeproj/project.pbxproj
+++ b/Simplenote.xcodeproj/project.pbxproj
@@ -252,6 +252,8 @@
 		B5B17F3A2425668400DD5B34 /* NSAttributedString+Simplenote.swift in Sources */ = {isa = PBXBuildFile; fileRef = B5B17F392425668400DD5B34 /* NSAttributedString+Simplenote.swift */; };
 		B5B17F3B2425668400DD5B34 /* NSAttributedString+Simplenote.swift in Sources */ = {isa = PBXBuildFile; fileRef = B5B17F392425668400DD5B34 /* NSAttributedString+Simplenote.swift */; };
 		B5B5F33D243244F100F31720 /* TextViewInputHandlerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = B5B5F33C243244F100F31720 /* TextViewInputHandlerTests.swift */; };
+		B5BB8CD424DB51670024E4C5 /* NSSearchField+Simplenote.swift in Sources */ = {isa = PBXBuildFile; fileRef = B5BB8CD324DB51670024E4C5 /* NSSearchField+Simplenote.swift */; };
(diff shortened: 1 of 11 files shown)


aerych:
> Verify that restoring a note from Trash causes the Notes List to immediately refresh

I had a little trouble testing this step.  After deleting a note I selected the Trash item in the menu, selected a note then tapped the restore icon.  However I wasn't able to tap back to view All Notes to see if it was restored.  That seems like a glitch, but was there a different way to restore that I'm overlooking? 

I have a question about the delete behavior.  Try this:
- Have a bunch of notes.
- Select a note in the upper part of the list
- Now, double tap (or right click) to get the delete option on a note that's lower in the list.
- Observe that the editor still shows the original note, but the list is showing that the lower note is selected.
- Delete the note.
- Now the selection updates both in the list and in the note editor.
This seemed a bit weird to me because I wasn't really trying to select the lower note, and it's content never appeared in the editor.  I'd have expected to stay on the first note I selected.   Thoughts?

jleandroperez:
@aerych I've spent a while on this one. Few insights!!

1. Not selecting a row when you right click on it is a bit confusing. Because you end up with a floating Menu, which will allow you to delete a note that's not the one you've got onscreen
2. Even worse... the standard NSMenu tooling freezes the Main Thread (switches the RunLoop mode to Event tracking), which is why the Editor doesn't get updated immediately.

So far, the best approach I've got involves a bit of hacking... but is 100% based on an ugly `DispatchQueue.main.async { }`, and that doesn't make me happy. Proper solution involves redoing the menus as NSPopover(s)!

If it's okay with you, I'll address that in #643 (we'll be adding new Note Actions soon-ish, too).

**Thank you!!!**

aerych:
Sounds good to me! Let's :shipit:!

jleandroperez:
Thank you sir!!

PR #645, opened 2020-08-11
Fix: Unresponsive Tags List

### LINKED ISSUE #636: Bug: Tags List non responsive
1. Delete a note
2. Shink the window
3. Click over any random tag
4. Try to click back over All Notes

As a result, All Notes will become "unclickable"

Thank you @aerych !!!

jleandroperez:
### Fix
In this PR we're fixing a bug in which the Tags List would not properly process click events.

@aerych Sir! may I bug you with yet another bugfix?

Thanks in advance!
Closes #636

### Test
1. Log into a test account with lots of tags (DM me if needed!)
2. Shrink the window so that the Tags List scrolls
3. Scroll the `All Notes` + `Trash` rows **above** the horizontal line defined by the Search Field (close to the titlebar or even below!)

- [ ] Verify that both **All Notes** and **Trash** are clickable
- [ ] If possible verify that any Tags placed in the same area (right by the Search Field) is clickable

### Release
`RELEASE-NOTES.txt` was updated in 6d05ef3 with:
 
> Fixed a bug in which the Tags List wouldn't accept mouse clicks

### DIFF
--- a/Simplenote/SPTableView.m
+++ b/Simplenote/SPTableView.m
@@ -29,19 +29,6 @@
     [super keyDown:theEvent];
 }
 
-- (BOOL)isRowClipped:(NSInteger)row
-{
-    if (row < 0) {
-        return NO;
-    }
-
-    NSRect rectOfRow = [self rectOfRow:row];
-    NSPoint rowInWindow = [self convertPoint:rectOfRow.origin toView:nil];
-    CGFloat windowHeight = self.window.contentLayoutRect.size.height;
-
-    return rowInWindow.y >= windowHeight;
-}
-
 - (NSInteger)rowForEvent:(NSEvent *)event
 {
     NSPoint location = [self convertPoint:event.locationInWindow fromView:nil];
@@ -56,10 +43,7 @@
 
 - (void)mouseDown:(NSEvent *)event
 {
-    NSInteger row = [self rowForEvent:event];
-    BOOL isRowClipped = [self isRowClipped:row];
-
-    self.disablesScrollToRow = self.disableAutoscrollOnMouseDown && !isRowClipped;
+    self.disablesScrollToRow = self.disableAutoscrollOnMouseDown;
 
     [super mouseDown:event];
 
--- a/Simplenote/SPTableView.m
+++ b/Simplenote/SPTableView.m
@@ -8,6 +8,11 @@
 
 #import "SPTableView.h"
 
+
+@interface SPTableView ()
+@property (nonatomic, assign) BOOL disablesScrollToRow;
+@end
+
 @implementation SPTableView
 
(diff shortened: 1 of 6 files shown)


jleandroperez:
@aerych sir! I've cracked a relatively clean workaround, for the jumpy UX.

Would it be possible to take another peek, just in case?

Cells that are not **fully visible** are still expected to produce the autoscroll (this is on purpose!). But for any other scenario in which the clicked cell is visible, no jump should happen.

**Thank you!!!**

Simplenote/SPTableView.m
```diff
@@ -24,24 +29,78 @@ - (void)keyDown:(NSEvent *)theEvent
...
+{
+    if (row < 0) {
+        return NO;
+    }
+
+    NSRect rectOfRow = [self rectOfRow:row];
+    NSPoint rowInWindow = [self convertPoint:rectOfRow.origin toView:nil];
+    CGFloat windowHeight = self.window.contentLayoutRect.size.height;
+
+    return rowInWindow.y >= windowHeight;
```
aerych, line 42: If we were to do `return (rowInWindow.y - rectOfRow.size.height) >= windowHeight;` seems like it might let the row be partially clipped and still avoid the jump?

aerych:
> Cells that are not fully visible are still expected to produce the autoscroll (this is on purpose!).

Why on purpose? Is there something else that happens that makes this necessary?

jleandroperez:
@aerych there's one specific scenario I was trying to address. Without that check, users might be able to click on a row, even when it's just a few pixels visible.

This felt a bit off, BUT if we were to allow such scenario, the code would get dead simple. WDYT?

Screen Shot 2020-08-12 at 13 09 21

Simplenote/SPTableView.m
jleandroperez: Replied below, thank you sir!

aerych:
> Without that check, users might be able to click on a row, even when it's just a few pixels visible.

I think its fine. Its the same behavior I would expect if I was using, say, an iPad with a mouse and clicked a UITableViewCell that was mostly scrolled off screen. 

It would also be consistent with the behavior of tags partially scrolled off the bottom of the tag list.  I can currently click to select those and the list doesn't jump.

Thoughts?

jleandroperez:
@aerych The code always speaks for itself. And this approach is **SO MUCH CLEANER**!!!

Thanks for the insight sir, I had doubts about that, to begin with 😄

aerych:
Love the change @jleandroperez!  Thanks for taking the extra time to wrangle the jumpy scroll. 
:shipit: sir!

jleandroperez:
Thank you Eric!!! Wooooo 🔥

PR #647, opened 2020-08-13
Merge 2.0 beta 1 into develop

mokagio:
We shipped this to pick up new translations and have a chance of testing them.


config/Version.public.xcconfig
```diff
@@ -2,4 +2,4 @@ VERSION_SHORT=2.0
 
 // Public long version example: 2.0.0.1. The last 1 means the first build for
 // this version
-VERSION_LONG=2.0.0.0
+VERSION_LONG=2.0.0.1
```
mokagio, line 5: _Dropping a comment here to make the change easier to spot._

fastlane/Fastfile
```diff
@@ -107,6 +107,8 @@ end
   desc "Updates a release branch for a new beta release"
   lane :new_beta_release do | options |
     ios_betabuild_prechecks(options)
+    ios_localize_project()
+    ios_update_metadata()
```
mokagio, line 111: I intentionally checked this change in.

As far as I can see, it's not something we do in WordPress or WooCommerce but I think it's useful to update the localizations as part of the beta build. In fact, that's exactly what this beta build was about.

What do you think? Is there something I'm missing? Otherwise, I could open PRs to add this change to the other projects, too.

loremattei: Yes, this may actually make sense. 

We started adding the `ios_localize` part to some projects, because sometimes changes in the beta brings in some changes to the strings. 
There's no warranty that strings pushed during a beta will be translated, but I don't see any downside with giving it a try. 

`ios_update_metadata`: this may be useful as well. We could spot issues with translation earlier and, again, I can't think about any downsides.

PR #655, opened 2020-08-25
Merge release 2.0 into develop

mokagio:
Notice 05995c, which works around the App Store Connect upload error by adding a new `BUILD_NUMBER` value in the `.xcconfig` to use for `CFBundleVersion`.


loremattei (approved):
LGTM!

Issue #663, opened 2020-09-01
Interlinking


jleandroperez:
### Tasks:
- [x] New **Copy Internal Link** Editor Action
- [x] New **Copy Internal Link** List Action
- [x] Interlinks should always get linkified
- [x] Clicking over an Internal Link should push the Target Note onscreen
- [x] Markdown Preview Interlink Support
- [x] Markdown Preview Highlight Colors
- [x] Inline Lookup Mechanism
- [ ] References UI

### Inline Lookup Mechanism
- Filtering Mechanism
  - [x] Backgrounding
  - [x] Invalidate previous search OPs
  - [x] Not Deleted
- Geometry
  - [x] Display below the keyword's first character!
  - [x] Prevent horizontal clipping (Screen Overflow)
  - [x] Prevent vertical clipping (Fullscreen)
  - [x] Relocate on Multiline?
- Interface
  - [x] Rounded Corners
  - [x] Cursor = Hand!
- Display / Hide Mechanism
  - [x] Dismiss when there are no keywords
  - [x] Dismiss on: [.mouseClick, .scrollWheel]
  - [x] Allow circular references
- User Input
  - [x] Review: Architecture. NSWindowController Ownership
  - [x] Arrows
  - [x] Interlink Insertion
- Phase II
  - [x] Display only when there are search results
  - [x] Dismiss when there are no results
  - [x] Discard notes that don't contain the keyword in the title
  - [x] BUG: Delete > Ctrl Z > Might result in AutocompleteWindow and incorrect cursor location. Main hook: KeyDown?
  - [x] Review Remote Update Scenarios
  - [x] Review performance in large accounts
  - [ ] Adapt window size to Content!
  - [ ] Display up to 5 autocomplete rows

### Mockups:
![blueprint_macos](https://user-images.githubusercontent.com/1195260/91873978-0e51c480-ec50-11ea-9e80-fadd44f6567a.png)

PR #670, opened 2020-09-09
Interlinks: Navigation Support

jleandroperez:
### Description:
This PR adds support for Interlinking Navigation. Opening `simplenote://note/KEY` will now cause the Editor to display the associated note.

@aerych Eric!! Can I bug you yet again?
Thanks in advance!!

Ref. #663

### Details:
- Linkification is now handled by SPTextView itself, whenever **displayNote** is invoked, or a Paste OP is performed.
- Few months ago, [Storage would remove every Attribute](https://github.com/Automattic/simplenote-macos/blob/1.8/External/Notepad/Storage.swift#L37) whenever the **Storage.Theme** property was set.
- [This "reset" process was overhauled](https://github.com/Automattic/simplenote-macos/blob/release/2.1/External/Notepad/Storage.swift#L180) so that only relevant attributes are removed, leaving intact any `.link` attributes.
- Because of the above, it's safe to drop explicit Linkification Invocations whenever the "Font / Markdown" flags are toggled.

### Note:
Please make sure you don't have `/Applications/Simplenote` installed, otherwise it'll capture the `simplenote://` scheme!

### Scenario: Instant Linkification
1. Right click over **Note A** > **Copy Internal Link**
2. Click over **Note B**
3. **Paste!**

- [x] Verify that the newly pasted link is **immediately linkified**

### Scenario: Navigation + Selected Tag
1. Right click over **Note A** > **Copy Internal Link**
2. Click over **Note B**
3. **Paste!**
4. Add both **Note A** and **Note B** to **Tag 1**
5. Click over **Tag 1**
6. Select **Note B**

- [x] Verify that clicking over the **Link to Note A** leads to the associated note
- [x] Verify that **Tag 1** remains selected

### Scenario: Navigation + All Notes
Repeat the steps from the previous scenario, but in Step 4 **only add Note B to Tag 1**

- [x] Verify that clicking over the **Note A** link causes the app to switch to **All Notes**
- [x] Verify that the Tags List **scrolls** to the All Notes row
- [x] Verify, also, that clicking over the **Note A** link causes the target note to be displayed

## Release
`RELEASE-NOTES.txt` was updated in 225da65 with:

> Clicking over a Note Link now causes the editor to display the referenced note #670


aerych (approved):
Hola @jleandroperez 👋 
Changes look and work great.  Tested as described and confirmed all expected behaviors. 
:shipit: sir!

jleandroperez:
Thank you Eric!!!

PR #671, opened 2020-09-09
Linkifying on Paste ©

jleandroperez:
### Description:
In this PR we're explicitly linkifying the document, whenever the user performs a **Paste** OP.

Plus, since we were in the neighborhood, few tech debt notes:

- Linkification is now handled by the **NSTextView.displayNote** API
- All of the Editor explicit Linkification / **checkTextInDocument** invocations have been dropped.

Ref. #663 

### Details:
Few months ago, [Storage would remove all of the NSAttributedString Styles](https://github.com/Automattic/simplenote-macos/blob/1.8/External/Notepad/Storage.swift#L37) whenever the **Storage.Theme** property was set.

[This "reset" process was overhauled](https://github.com/Automattic/simplenote-macos/blob/release/2.1/External/Notepad/Storage.swift#L180) so that only `.font, .foreground, .paragraphStyle` attributes are removed, leaving intact any `.link` attributes.

Because of the above, it's safe to drop explicit Linkification Invocations whenever the "Font / Markdown" flags are toggled.

### Test
[TBD]

### Release

### DIFF
--- a/Simplenote/NSTextView+Simplenote.swift
+++ b/Simplenote/NSTextView+Simplenote.swift
@@ -128,6 +128,39 @@ extension NSTextView {
 }
 
 
+// MARK: - Linkification
+//
+extension NSTextView {
+
+    /// Asynchronously Processess Links in the Document
+    ///
+    @objc
+    func processLinksInDocumentAsynchronously() {
+        DispatchQueue.main.async(execute: processLinksInDocument)
+    }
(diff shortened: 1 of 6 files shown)


jleandroperez:
Merging into PR #670 since the footprint of these changes are minimum, and without them, #670 is a bit shaky.

PR #672, opened 2020-09-10
Switch configuration to trunk

loremattei:
This PR updates `.configure` to use trunk instead of main.
It also updates `fastlane` to the latest version and `.configure` to the current file format. 

### Test
1. `bundle install`
2. Run `bundle exec fastlane run configure_apply`
3. Verify that the tool succeeds. If you already had the latest configuration, you shouldn't have any changes.
4. Delete all the files in the `files_to_copy` section of `.configure` from your project directory.
5. Run `bundle exec fastlane run configure_apply` again and verify that the file you deleted have been recreated.
6. Build the app and make sure it works as expected. 

### Review
Only one developer is required to review these changes, but anyone can perform the review.

### Release


mokagio (approved):
Tested, works as described ✅ 

---

One thing that surprised me a bit, though, was that when I first run `configure_apply`, it failed with:

```
[!] Decryption key could not be found
```

To me, the reason was clear: I had left my `.mobile-secrets` `HEAD` pointing to `main`, and once I checked out `trunk`, it worked fine. I'm not sure how clear this would be for someone less involved with the infra stuff as we are.

Anyways, that's more of a comment on the tool itself than this PR. Maybe something to keep in mind in the possible upcoming rewrite.

loremattei:
Ah! I see! This happens randomly. I have an issue open for this here: https://github.com/wordpress-mobile/release-toolkit/issues/166. 
There's actually code in the toolkit to pull and checkout the right branch, but I'm pretty sure there's a path in the flow where we look for the decryption key before pulling... 
Something to investigate more. 
If you noticed more details, please add them to the issue.

PR #675, opened 2020-09-18
Merge release 2.1 into develop

mokagio:
There is no build number bump here. I was expecting the `finalize_release` lane to do it, but I didn't, which is something I'll have to look into. It might have to do with the new versioning strategy here.

The lane did tag the local `HEAD` with 2.1, so I thought to just leave it as it is and open the PR like this.


jkmassel (approved):
All translations makes it easy to approve!

:shipit:

PR #678, opened 2020-09-23
Interlinks Autocomplete UI: Mark I

jleandroperez:
### Description:
This PR implements the Interlink Autocomplete UI, and takes care of all of the logic required to display the Autocomplete Window below the Editor's cursor, when appropriate.

**Note:** Actual Interlink insertion will come up in a follow up PR!

@aerych Would you be game for this one?
Thanks in advance!!

P.s.: Apologies about the size of the PR, new Storyboard included!!

Ref. #663

### Details:
For the record! We're using a NSWindow instance for Autocomplete purposes, because the Autocomplete Feature must:

A. Not interfere with Text Input in the main editor. You must be able to go on typing
B. Not be affected by the ScrollView
C. And it must also not be affected, anyhow, by anything else in the hierarchy

### Test: Interface!
1. Open any note
2. Enable Markdown
3. Type `Hello 🇮🇳 🌎 [a`  (any emoji would work!)

- [x] Verify a Window shows up below the `a` character, with 3 placeholder rows
- [x] Verify the Autocomplete Window has rounded corners (macOS +10.15)
- [x] Verify that the mouse cursor switches to *Pointing Hand*, when hovering over the Interlink Suggestion Window
- [x] Verify that Light / Dark Mode switches affect the autocomplete window

### Test: Window Vertical Positioning
1. Place the Simplenote Window at the bottom of the screen
2. Insert `[a` in the very last line of the editor

- [x] Verify that the autocomplete window shows up above the cursor, and not below (to avoid clipping!)

### Test: Window Horizontal Positioning
1. Place the Simplenote Window close to the right edge of the screen
2. Insert `[a` on the righter-most corner of the editor

- [x] Verify the Autocomplete Window is positioned in such a way that it doesn't get clipped

### Release

### Screenshots
Autocomplete


aerych:
> Apologies about the size of the PR, new Storyboard included

After the last couple of (cough) large (/cough) PRs you've reviewed for me, no apology needed amigo! 🙇‍♂️

Simplenote/InterlinkWindowController.swift
```diff
@@ -0,0 +1,87 @@
...
+        guard #available(macOS 10.15, *) else {
+            return
+        }
+
+        window?.backgroundColor = .clear
+        window?.isOpaque = false
+    }
+
+    func calculateWindowOrigin(windowSize: CGSize, positioningRect: CGRect) -> CGPoint {
+        let screenWidth = NSScreen.main?.visibleFrame.width ?? .infinity
```
aerych: I'm not too familiar with the way screen positioning works.  Do we use `.infinity` to ensure the calculated origin is waaaaaay off screen in the event that NSScreen.main? is nil as sort of a safety net?

When NSScreen.main == nil, is there any reason to not return early and avoid the extra math?  Maybe something like
```
guard let screen = NSSCreen.main else {
    return CGRect.infinite.origin
}
```
Just curious!

Simplenote/NSStoryboard+Simplenote.swift
```diff
@@ -12,13 +12,22 @@ extension NSStoryboard {
 
         return target
     }
+
+    func instantiateWindowController<T: NSWindowController>(ofType type: T.Type) -> T {
```
aerych, line 16: Love this trick

aerych (approved):
Hola @jleandroperez 👋 
Changes look good!  Tested as described and everything worked as expected. 
I had one question but it's nothing that's a blocker. Just my curiosity more than anything. 
:shipit:  when ready!

Simplenote/InterlinkWindowController.swift
jleandroperez: Actually!! I was just trying to neutralize the "Overflow in horizontal coordinates" check. 
But you're right!! it's just best to do a simple unwrap, if possible.

Old school frame calculations, in Mac the (0, 0) is at the bottom left corner of the UI.

A negative Y, or a "Max X" exceeding the screen width imply that the window wouldn't be visible, so we either move the popup to compensate.

Thanks for spotting this!!

Simplenote/NSStoryboard+Simplenote.swift
jleandroperez: Thank you sir!!

jleandroperez:
Thank you for reviewing this one Eric!!!

PR #680, opened 2020-09-28
Interlinks Autocomplete UI: Mark II

jleandroperez:
### Details:
- Reimplements **SPWindow** in Swift: Adds support for the **ParentWindowDelegate** protocol
- Implements **AutocompleteWindow**

This new **NSWindow** subclass receives events from its parent **Window** entity, and allows us to redirect specific events (such as Arrows / Esc / Return) over to the Autocomplete helper window.

That way, the main Editor keeps the focus / firstResponder, and the Autocomplete UI reacts when needed.

@aerych sir!!! Can I trouble you with a **REALLY FUN** PR?
Thanks a looot!!

🔥 

Ref. #663

### Test
1. Open any note and enable Markdown
2. Type `[keyword` with `keyword` being a word contained within any random note

- [x] Verify the **Autocomplete Window** now shows the title of notes containing `keyword` in their contents
- [x] Verify that **Right / Left click** in the Parent Window (Editor) dismisses the Autocomplete Window
- [x] Verify that pressing **ESC** dismisses the Autocomplete Window
- [x] Verify that pressing **Arrow Up / Arrow Down** let you change the highlighted Autocomplete Row
- [x] Verify switching to **Light / Dark / System** themes cause the main app to instantaneously refresh
- [x] Verify switching to **Light / Dark / System** themes cause the Autocomplete Window to instantaneously refresh

### Release


jleandroperez:
I've clicked entirely by accident the big Merge button in the Part 3 PR. Reverted, apologies about the noise!!

Simplenote/InterlinkViewController.swift
```diff
@@ -55,12 +83,28 @@ private extension InterlinkViewController {
...
+
+    func setupResultsController() {
+        resultsController.limit = Settings.maximumNumberOfResults
+        resultsController.onDidChangeContent = { [weak self] _, _ in
+            self?.tableView.reloadAndPreserveSelection()
+        }
+    }
+
+    func refreshResultsPredicate(for keyword: String) {
+        resultsController.predicate = NSCompoundPredicate(andPredicateWithSubpredicates: [
```
aerych, line 101: Should this be modified to exclude the note being edited? Otherwise it will always appear in the search results, and can be a false positive match.

aerych (approved):
Howdy @jleandroperez 👋 
Tested as described and it worked as expected.  Code looks good.
I had one question about the results but otherwise :shipit: sir! :)

Simplenote/InterlinkViewController.swift
jleandroperez: We're gonna have to change the Predicate itself, before shipping (I still haven't figured out how!). Lookup should happen at the Note Title level, soooo the glitch itself will be gone soon-ish!!

Problem there is: there's a single String field at the coredata level, so it's nowhere near straightforward.

Thank you for bringing that up sir!!

jleandroperez:
Thanks a looot for reviewing this one sir!!

PR #688, opened 2020-10-02
Interlinks Autocomplete UI: Mark III

jleandroperez:
### IMPORTANT
- [x] Ship SimplenoteInterlinks Package 1.1.0 before merging!!

### Details:
In this PR we're enhancing the Interlink Autocomplete Window so that **return** or **double click** inserts the selected note's Interlink.

@aerych Sir!! May I bug you yeeet again?
Thanks in advance!!!

Ref. #663
Ref. [Simplenote Interlink PR 2](https://github.com/Automattic/SimplenoteInterlinks-Swift/pull/2)

### Test
1. Open any note and enable Markdown support
2. Type `[keyword` (with `keyword` being a word present in any other note)
3. Select any note that shows up in your search reuslts

- [x] Verify that double clicking over any row inserts a **Simplenote Interlink™**
- [x] Verify that pressing return over any row **also** inserts a **Simplenote Interlink™**
- [x] Verify the newly inserted link is linkified (underlined, should be clickable!)
- [x] Verify that clicking over such link ends up opening the target note

### Release
`RELEASE-NOTES.txt` was updated in acd1da5 with:

> Typing `[` will now allow you to pick Interlinking Notes right from the editor


Simplenote/InterlinkViewController.swift
```diff
@@ -32,6 +32,10 @@ class InterlinkViewController: NSViewController {
         return ResultsController(viewContext: mainContext, sortedBy: [sortDescriptor])
     }()
 
+    /// Closure to be executed whenever a Note is selected. The Interlink URL will be passed along.
+    ///
+    var onInsertInterlink: ((String) -> Void)?
```
aerych, line 37: This is me being a little picky, but could we include a mention that the callback is one-time use only?  If at some point in the future interlinks were ever used in a way where the owner didn't dismiss the view as part of the call back block the control becoming suddenly [partially] unresponsive might be surprising.  Hopefully a note about it would help a future coder.

Alternatively, would it make sense for the owner to be responsible for nullifying the callback since it handles the assignment?

Simplenote/SPTableView.m
```diff
@@ -25,6 +25,14 @@ - (void)keyDown:(NSEvent *)theEvent
             return;
         }
     }
+
+    // Nü Event Handling Mechanism
```
aerych, line 29: The umlaut makes it metal

aerych (approved):
Hola @jleandroperez 👋 
Looking good from here.  Test scenarios all pass.  Unit tests work for me locally (we might need to teach CI to fetch packages to make it happy?)
I had one nitpicky code question but otherwise :shipit:

Simplenote/SPTableView.m
jleandroperez: Long live Nü Metal!!

Simplenote/InterlinkViewController.swift
jleandroperez: On second thoughts, let's drop the "reset callback" on selection (since the callback is weak anyways!!).

The ideal flow involves, as you mentioned, the owner that assigns the callback actually nullifying the closure. But that code doesn't look cute LOL 😂 !!

jleandroperez:
Thank you @aerych !!!

PR #696, opened 2020-10-19
Interlink References

### LINKED ISSUE #663: Interlinking

jleandroperez:
### Details:
- Rebuilds the Metrics UI, based on NSTableView
- Implements new NSTableViewCell subclasses: `[MetricTableViewCell, ReferenceTableViewCell, SeparatorTableViewCell]`
- Metrics: We're now displaying inbound note references

@eshurakov Mind taking a look at this one Evgeny?
Thank youuu!

Ref. #663 

### Test: Multiple Selection
1. Log into your account!
2. Right click over **Note A** in your notes list: click over `Copy Internal Link`
3. Click over **Note B** and paste the contents of your clipboard
4. CMD + Click over **Note A** + **Note B** in your list
5. Click over the top right `( i )` icon

- [x] Verify only 5 rows show up:
- [x] Verify only `Words` and `Characters` are populated
- [x] Verify that editing either `Note A` or `Note B` in a secondary device (simplenote.com on the web works!) causes the UI to refresh

### Test: Single Selection
1. Log into your account!
2. Right click over **Note A** in your notes list: click over `Copy Internal Link`
3. Paste the contents of your clipboard on **at least** 4 notes (so that 4 notes reference back to Note A!)
4. Click over **Note A**
5. Click over the top right `( i )` icon

- [x] Verify that all of the inbound links (References!) you pasted in step 3 show up at the bottom of the Metrics UI
- [x] Verify that you can scroll **all of the contents** of the Metrics UI (every single row should move!)

### Release

---

### Screenshots
Screen Shot 2020-10-21 at 15 22 11

Screen Shot 2020-10-21 at 15 22 24

### DIFF
--- a/Simplenote/MetricsViewController.swift
+++ b/Simplenote/MetricsViewController.swift
@@ -279,6 +279,7 @@ private extension MetricsViewController {
     func dequeueHeaderCell(from tableView: NSTableView, text: String) -> NSView {
         let headerCell = tableView.makeTableViewCell(ofType: HeaderTableCellView.self)
         headerCell.title = text
+        headerCell.titleColor = .simplenoteTextColor
         return headerCell
     }
 
--- a/Simplenote/MetricsViewController.swift
+++ b/Simplenote/MetricsViewController.swift
@@ -7,30 +7,21 @@ import SimplenoteFoundation
 //
 class MetricsViewController: NSViewController {
 
-    /// Section Headers
+    /// Enclosing ClipView
     ///
-    @IBOutlet private(set) var informationTextLabel: NSTextField!
-    @IBOutlet private(set) var referencesTextLabel: NSTextField!
+    @IBOutlet private weak var clipView: NSClipView!
 
-    /// Modified: Left Text / Right Details
+    /// Our main TableView
     ///
-    @IBOutlet private(set) var modifiedTextLabel: NSTextField!
-    @IBOutlet private(set) var modifiedDetailsLabel: NSTextField!
+    @IBOutlet private weak var tableView: NSTableView!
 
-    /// Created: Left Text / Right Details
+    /// NSPopover instance that's presenting the current instance.
     ///
-    @IBOutlet private(set) var createdTextLabel: NSTextField!
-    @IBOutlet private(set) var createdDetailsLabel: NSTextField!
-
-    /// Words: Left Text / Right Details
-    ///
-    @IBOutlet private(set) var wordsTextLabel: NSTextField!
-    @IBOutlet private(set) var wordsDetailsLabel: NSTextField!
-
-    /// Characters: Left Text / Right Details
-    ///
-    @IBOutlet private(set) var charsTextLabel: NSTextField!
-    @IBOutlet private(set) var charsDetailsLabel: NSTextField!
+    private var presentingPopover: NSPopover? {
+        didSet {
+            refreshStyle()
+        }
+    }
 
     /// Notes whose metrics should be rendered
     ///
@@ -44,23 +35,17 @@ class MetricsViewController: NSViewController {
 
     /// Entity Observer
     ///
-    private let observer: EntityObserver
+    private lazy var observer = EntityObserver(context: mainContext, objects: notes)
 
     /// ResultsController: In charge of CoreData Queries!
     ///
-    private lazy var resultsController: ResultsController<Note> = {
-        return ResultsController<Note>(viewContext: mainContext, sortedBy: [
-            NSSortDescriptor(keyPath: \Note.content, ascending: true)
-        ])
-    }()
+    private lazy var resultsController = ResultsController<Note>(viewContext: mainContext, sortedBy: [
+        NSSortDescriptor(keyPath: \Note.content, ascending: true)
+    ])
 
-    /// NSPopover instance that's presenting the current instance.
+    /// Rows to be rendered
     ///
-    private var presentingPopover: NSPopover? {
-        didSet {
-            refreshStyle()
-        }
-    }
+    private var rows = [Row]()
 
 
     // MARK: - Lifecycle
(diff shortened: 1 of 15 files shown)


Simplenote/MetricsViewController.swift
```diff
@@ -7,41 +7,52 @@ import SimplenoteFoundation
...
+    /// ResultsController: In charge of CoreData Queries!
     ///
-    private var presentingPopover: NSPopover? {
-        didSet {
-            refreshStyle()
-        }
-    }
+    private lazy var resultsController = ResultsController<Note>(viewContext: mainContext, sortedBy: [
+        NSSortDescriptor(keyPath: \Note.content, ascending: true)
+    ])
```
eshurakov, line 51: Just a thought, should the order match the order that is set in the list? (last modified / alphabetically)

Simplenote/MetricsViewController.swift
```diff
@@ -82,16 +89,22 @@ class MetricsViewController: NSViewController {
...
         observer.delegate = self
     }
+
+    func setupResultsControllerIfNeeded() {
+        guard mustSetupResultsController, let plainInterlink = notes.first?.plainInterlink else {
+            return
+        }
+
+        resultsController.predicate = NSPredicate.predicateForNotes(exactMatch: plainInterlink)
+        try? resultsController.performFetch()
```
eshurakov, line 106: We probably need to setup a listener here and update the table? 🤔

eshurakov (commented):
Looks great! 

Nitpicking: I noticed that last modified date is formatted quite differently. Wonder if we want to make it more similar and for example drop time in the list but keep the date format the same.

Also we use different words for the same thing: "Date Updated", "Modified", "Last Modified".

Screenshot 2020-10-22 at 07 37 05

Simplenote/MetricsViewController.swift
```diff
@@ -138,21 +143,186 @@ private extension MetricsViewController {
...
-        wordsDetailsLabel.stringValue = String(metrics.numberOfWords)
-        charsDetailsLabel.stringValue = String(metrics.numberOfChars)
+            .metric(title: NSLocalizedString("Created", comment: "Note Creation Date"),
+                    value: metrics.creationDate),
+
+            .metric(title: NSLocalizedString("Words", comment: "Number of words in the note"),
+                    value: String(metrics.numberOfWords)),
+
+            .metric(title: NSLocalizedString("Characters", comment: "Number of characters in the note"),
+                    value: String(metrics.numberOfChars))
```
eshurakov: Maybe we can format numbers using a decimal formatter. If the number is big, it looks better and easier to read with delimiters.

jleandroperez: Awesome idea, done!! thank you!

Simplenote/MetricsViewController.swift
jleandroperez: 🤦  I totally forgot about that. Done, thank you!!

Simplenote/MetricsViewController.swift
jleandroperez: Let's skip that one and keep it alpha, for a bit at least! 

We've got missing sort options available in other platforms, it'd be neat to address that first.

jleandroperez:
@eshurakov Q for you!!

```
noticed that last modified date is formatted quite differently. Wonder if we want to make it more similar and for example drop time in the list but keep the date format the same.
```

I can't find where else is the last modified date showing up. Is it possible that was iOS?

Regarding the wording change, I'd stick to the wording in the mockups if you don't mind! It feels a bit redundant to include the word `date` (when the text on the right hand is a date).

Other than that, ready for another peek sir! thaaank you!!

eshurakov:
> I can't find where else is the last modified date showing up. Is it possible that was iOS?

Sorry @jleandroperez, what I meant is that in the note information popover `modified` date is formatted as "21 Oct 2020 at 10:10" and `last modified` date for each reference is formatted as "21/10/2020".
So I was wondering if it makes sense to use format that is closer to each other. For example, format the date in second case as "21 Oct 2020" without time.

eshurakov (approved):
🚀 
🚢 :shipit:

jleandroperez:
Ahhhhh thank you @eshurakov !!! totally missed that. Love the idea, updated!

🔥

PR #707, opened 2020-11-04
Extract all build settings to xcconfigs

mokagio:
For my 2020 Q4 HACK week project, I decided to look into extracting all of the project configurations into `xcconfig` files.

## Rationale

Having the build settings in dedicated files brings a few advantages:

- When the configs change, the resulting diff is cleaner and easier to review
- We can add comments in the `xcconfig`s, which helps adding context to configurations that might not be standard
- Extracting configurations also helps in case we'll move towards project generators like [Tuist](https://tuist.io/) or better build systems like [Bazel](https://bazel.build/) to help us better manage the project's complexity and build performance.

## The Process

I started with Simplenote macOS because it's the one with the simplest build configurations and target setup.

To extract the configs, I used [BuildSettingsExtractor app](https://buildsettingextractor.com/) made by @dempseyatgithub. The app uses a neat structure for the `xcconfig`s it generates: the project and all of the targets have each a shared `xcconfig` which is imported by the build configuration files.

The app also adds helpful documentation comments for each setting on which they are available. These comments, while helpful, also add a lot of lines to the diff. I'm conflicted between keeping them or not. I opened #709 to show this diff but without comments, keen to know what you think of it. 

### DRYing the setup

I attempted to DRY the configs setup a bit by extracting a couple of common settings I noticed while reading the files: the default code signing and the Swift optimization setting for the Debug builds.

I'm sure there are other opportunities to extra configurations in composable units, but I time-boxed my experiment.

## Future proofing

To avoid new settings making it into the project and being missed during code review, I added a Run Script build phase that runs the [`VerifyNoBS`](https://github.com/olofhellman/VerifyNoBS) script by @olofhellman. This script parses the project file and fails the build if there are build settings.

## To Test

This PR touches all the targets and all the build configurations. I'm not sure how to test it thoroughly, but here's what I did:

- Run the unit tests
- Archive both schemes, and validated the App Store one with ASC via the Xcode UI
- Made sure there are no new warnings

CI also builds and tests the Simplenote scheme and as you can see it's ✅ 

To test that the build fails when adding a new build settings to the project, I added a new build setting to the project and tried to build 😄 

Screen Shot 2020-11-06 at 12 26 05 pm

## Feedback Please

I'm keen to know what you think of this approach. I think it's worth pursuing because of the benefit listed above.

The many config files setup seems cumbersome at first, but I believe it's granularity and consistency will make it easier to maintain in the long run. I also feel that if we find more opportunities to DRY the leaf config files, it will become easier to wrap our head around them.

Other questions:

- Is there anything I might have missed in the setup or in testing it that might bite us later down the track?
- What about the documentation in the files, keep the comments? No comments? _Some_ comments?

Thanks 😄


Simplenote.xcodeproj/project.pbxproj
```diff
@@ -757,7 +772,7 @@
 		26F72A7D14032D2900A7935E = {
 			isa = PBXGroup;
 			children = (
-				8C902F8B22D3EDB30018D654 /* config */,
+				3FAA553325526BBE001534D4 /* config */,
```
mokagio, line 775: This is just a change from a pure "group" (the icon in the Project Navigator has a little triangle in the bottom right corner) to one that has a folder attached to it.

config/Project-Debug.xcconfig
```diff
@@ -0,0 +1,84 @@
...
+// * Fastest, Smallest: Optimize for size. This setting enables all `Faster`
+// optimizations that do not typically increase code size. It also performs further
+// optimizations designed to reduce code size. [-Os]
+// * Fastest, Aggressive Optimizations: This setting enables `Fastest` but also enables
+// aggressive optimizations that may break strict standards compliance but should work
+// well on well-behaved code. [-Ofast]
+// * Smallest, Aggressive Size Optimizations: This setting enables additional size
+// savings by isolating repetitive code patterns into a compiler generated function.
+// [-Oz]
+GCC_OPTIMIZATION_LEVEL = 0
```
mokagio, line 61: This is what I mean regarding the comments being useful but also noisy 😳 

This information is available [online](https://developer.apple.com/library/archive/documentation/DeveloperTools/Reference/XcodeBuildSettingRef/1-Build_Setting_Reference/build_setting_ref.html#//apple_ref/doc/uid/TP40003931-CH3-SW90) and in the Xcode Help:

Screen Shot 2020-11-06 at 1 59 01 pm

Not having the comment would be a bit of extra friction to get to the meaning of a setting, but if it makes the file easier to digest and work with, it might be worth it.

config/Swift-Optimization-Debug.xcconfig
```diff
@@ -0,0 +1,7 @@
+// Swift Optimization Level
+//
+// * None: Compile without any optimization. [-Onone]
+// * Optimize for Speed: [-O]
+// * Optimize for Size: [-Osize]
+// * Whole Module Optimization: [-O -whole-module-optimization]
+SWIFT_OPTIMIZATION_LEVEL = -Onone
```
mokagio, line 7: I called this file "-Debug" instead of "-Onone" because I feel it explains better that it contains the optimization value for a debug build.

I think there might be other settings we can move here, too.

This naming is a bit inconsistent with `Code-Sign-Mac-Developer.xcconfig` above. In that case, though, I felt being explicit was better. Even more so, this value is used in the Simplenote Release config (because it doesn't override the value from Shared)

mokagio:
Asked review from @Automattic/release-managers as well as the Simplenote Apple devs to make sure we get everyone's input on this.

olofhellman:
Glad to see VerifyNoBS is useful!

olofhellman:
One may want to consider allowing for local override of some settings via an optional .xcconfig file.  For example, the case where a developer not on your development team wants to compile and sign the app, but doesn't want to manually edit the DEVELOPMENT_TEAM setting, or for the case of having a build machine that uses a different DEVELOPMENT_TEAM, or some alpha / beta release special settings, or who knows what.  The NetNewsWire project handles this by #including an optional .xcconfig file like this:

#include? "../../SharedXcodeSettings/ProjectSettings.xcconfig"

So developers (or build machines) can have their own .xcconfig file outside the repo and have it part of the build without any changes.

jkmassel (commented):
I gave this a read-over and I generally like this approach.  I've included a few thoughts below:

1. I think I strongly prefer having the comments in the files. It's a lot of noise now, but it makes it far clearer what's going on. I _do_ wish that we didn't have to specify _all_ of them though :( 

2. I'm concerned about Xcode upgrades – often they'll add new compiler directives (something like `GCC_WARN_UNUSED_FUNCTION`) – will Xcode add them to the project, or be smart enough to put them in the `xcconfig` file?

3. Regarding Xcode upgrades – will the `.swift` script continue to work as Apple changes the `.xcproj` file format? I'm also wondering if this should be a linter check rather than a build step, but that's relatively minor.

4. I like the idea presented to have a user-definable set of options for code signing by OSS developers. 

5. The tree here seems fairly complex. I wonder about possibly combining build settings between the `AppStore-Shared`, `Project-Shared`, and `Simplenote-Shared` files? 

6. Do we need to enforce `.xcconfig` files for test and secrets targets? (if so can we make the secrets stuff _not_ be a target?) 😅

---

In general I like this, but as you mentioned, it seems there's maybe still some duplication that could be cleaned up? IMHO it would make sense to do that prior to merging this, though I definitely see value in this approach.

dempseyatgithub:
> 2. I'm concerned about Xcode upgrades – often they'll add new compiler directives (something like `GCC_WARN_UNUSED_FUNCTION`) – will Xcode add them to the project, or be smart enough to put them in the `xcconfig` file?

When Xcode suggests project changes, they are added to the project file. It is not smart enough to add them to xcconfig files. Typically, the handful of changes introduced can be copied from the Xcode project editor and pasted into the xcconfig file. Or, you could run BuildSettingExtractor again and generate throwaway xcconfig files in a temporary location. The generated xcconfig files will only contain the settings that are in the project. You can then copy and paste the settings into the corresponding xcconfig files in your project. (This way comments are generated for the new settings as well.)

In either case, the final step is to delete the settings from the project file.

Note all of this can be done at once, so the final commit just shows the new settings being added to the xcconfig files, and the update of the version check in the Xcode project file.

The VerifyNoBS script is a decent way to check that no changes were accepted that were left in the project file.

> 3. Regarding Xcode upgrades – will the `.swift` script continue to work as Apple changes the `.xcproj` file format? I'm also wondering if this should be a linter check rather than a build step, but that's relatively minor.

The format of the project.pbxproj file has remained very stable for over a decade. For example, BuildSettingExtractor uses the same code handling project file versions back to Xcode 3.2. So, unless there is a big upheaval, I would imagine the script would continue to work.

> 5. The tree here seems fairly complex. I wonder about possibly combining build settings between the `AppStore-Shared`, `Project-Shared`, and `Simplenote-Shared` files?

BuildSettingExtractor generates these with the notion that having the structure in place to put shared settings at the project level (for use in all build configurations in all targets) and the target level (for use in all build configurations in a single target) makes it easier to put a build setting where it should be since since the friction of adding additional xcconfig files, understanding how to include one from another, matching the file naming scheme, are already taken care of.

But it's completely up to team preference whether to keep that structure as-is or modify it.

config/Project-Debug.xcconfig
AliSoftware: Yeah I agree that comments _that_ long will likely be too noisy for little benefits. It would be nice if it was only including the first line instead of the whole shebang.

But at that point I think having such lengthy comments is not worth it imho: we could always refer to either the official –even if not fully complete– Apple documentation, or to de-facto references like https://pewpewthespells.com/blog/buildsettings.html (†).  
I think it's better to use comments for manually documenting why we introduced some esoteric build setting for non-obvious specific cases, rather than using them to show the raw definition of the setting which can always be found somewhere else anyway.

_(†) Sam also has a couple of other articles [in her blog](https://pewpewthespells.com/blog/), about the Xcode build system, and how to use Build Settings efficiently, etc, that could be worth referencing btw._

AliSoftware:
> Regarding Xcode upgrades – will the .swift script continue to work as Apple changes the .xcproj file format? I'm also wondering if this should be a linter check rather than a build step, but that's relatively minor.

As James said, the Xcodeproj format has been stable for a while. A lot of tools, including `CocoaPods` and our `Xcodeproj` internal gem it relies on, and Tuist and the equivalent `Xcodeproj` Swift library it relies on. But more than that, the `VerifyNoBS.swift` script doesn't rely on the internal structure of the pbxproj file, but instead just does regex parsing of the text content (which will only work if the pbxproj file is in ASCII-Plist format, but it usually is for 99.9% of cases including ours so that's ok). So even if the format introduces other keys or moves those around in the pbxproj file, as long as the key names don't change (quite unlikely) we should be good

We could also implement a similar functionality as `VerifyNoBS` with other solutions, one being to use the `Xcodeproj` gem to parse the pbxproj file (as a tree/object, not as a text file with regex), or even as a simple Plist (`pbxproj` format is an ASCII Plist after all, so `NSDictionary(contentsOfFile: pbxprojFile)` would work too) to inspect the structure and warn about any `buildSettings` entry defined there. It might actually be a bit faster than running regexes on every line of the `pbxproj`? (just wondering, not claiming, didn't test).

It would make us lose the information about the line number where those settings would appear in the `pbxproj` textual content tho, so would not allow us to do inline warnings attached to a specific line (is it really useful given that if such a violation appears we'll likely edit it out from Xcode's UI not by editing the text in `pbxproj` manually?), but otoh it would allow us to provide context for those build settings (e.g. use the dictionary representation to know what the `name` of the `XCBuildConfiguration` or target those settings are attached to) which might be more useful to fix it from Xcode UI?

config/Simplenote-Debug.xcconfig
```diff
@@ -0,0 +1,13 @@
...
+// Import the CocoaPods xcconfig first, so we can override what's needed
+#include "../Pods/Target Support Files/Pods-Automattic-Simplenote/Pods-Automattic-Simplenote.debug.xcconfig"
+
+#include "Simplenote-Shared.xcconfig"
+#include "Swift-Optimization-Debug.xcconfig"
+
+// Preprocessor Macros
+//
+// Space-separated list of preprocessor macros of the form `foo` or `foo=bar`.
+GCC_PREPROCESSOR_DEFINITIONS = $(inherited) COCOAPODS=1
```
AliSoftware, line 13: This should be removed. It's supposed to already be provided by the CocoaPods's generated xcconfig instead (included on line 5)

config/Project-Shared.xcconfig
```diff
@@ -0,0 +1,300 @@
...
+// declaration.
+GCC_WARN_UNUSED_VARIABLE = YES
+
+// Runpath Search Paths
+//
+// This is a list of paths to be added to the `runpath` search path list for the image
+// being created. At runtime, `dyld` uses the `runpath` when searching for dylibs whose
+// load path begins with `@rpath/`. See [Dynamic Library Programming
+// Topics](https://developer.apple.com/library/content/documentation/DeveloperTools/Conceptual/DynamicLibraries/000-Introduction/Introduction.html).
+LD_RUNPATH_SEARCH_PATHS = $(inherited)
```
AliSoftware, line 244: Can be removed since it doesn't add anything.

config/Project-Shared.xcconfig
```diff
@@ -0,0 +1,300 @@
...
+
+// Library Search Paths
+//
+// This is a list of paths to folders to be searched by the linker for libraries used by
+// the product. Paths are delimited by whitespace, so any paths with spaces in them need
+// to be properly quoted.
+LIBRARY_SEARCH_PATHS = $(inherited)
```
AliSoftware, line 251: Can be removed since it doesn't add anything.

config/Simplenote-AppStore-Shared.xcconfig
```diff
@@ -0,0 +1,170 @@
...
+// load path begins with `@rpath/`. See [Dynamic Library Programming
+// Topics](https://developer.apple.com/library/content/documentation/DeveloperTools/Conceptual/DynamicLibraries/000-Introduction/Introduction.html).
+LD_RUNPATH_SEARCH_PATHS = $(inherited) @loader_path/../Frameworks
+
+// Library Search Paths
+//
+// This is a list of paths to folders to be searched by the linker for libraries used by
+// the product. Paths are delimited by whitespace, so any paths with spaces in them need
+// to be properly quoted.
+LIBRARY_SEARCH_PATHS = $(inherited)
```
AliSoftware, line 120: Can be removed since it doesn't add anything.

config/Simplenote-AppStore-Shared.xcconfig
```diff
@@ -0,0 +1,170 @@
...
+
+// Other C Flags
+//
+// Space-separated list of additional flags to pass to the compiler for C and Objective-C
+// files. Be sure to backslash-escape any arguments that contain spaces or special
+// characters, such as path names that may contain spaces. Use this setting if Xcode does
+// not already provide UI for a particular C or Objective-C compiler flag.
+OTHER_CFLAGS = ${inherited}
```
AliSoftware, line 128: Can be removed since it doesn't add anything.

config/Simplenote-Shared.xcconfig
```diff
@@ -0,0 +1,182 @@
...
+// load path begins with `@rpath/`. See [Dynamic Library Programming
+// Topics](https://developer.apple.com/library/content/documentation/DeveloperTools/Conceptual/DynamicLibraries/000-Introduction/Introduction.html).
+LD_RUNPATH_SEARCH_PATHS = $(inherited) @loader_path/../Frameworks
+
+// Library Search Paths
+//
+// This is a list of paths to folders to be searched by the linker for libraries used by
+// the product. Paths are delimited by whitespace, so any paths with spaces in them need
+// to be properly quoted.
+LIBRARY_SEARCH_PATHS = $(inherited)
```
AliSoftware, line 132: Can be removed since it doesn't add anything.

config/Simplenote-Shared.xcconfig
```diff
@@ -0,0 +1,182 @@
...
+
+// Other C Flags
+//
+// Space-separated list of additional flags to pass to the compiler for C and Objective-C
+// files. Be sure to backslash-escape any arguments that contain spaces or special
+// characters, such as path names that may contain spaces. Use this setting if Xcode does
+// not already provide UI for a particular C or Objective-C compiler flag.
+OTHER_CFLAGS = ${inherited}
```
AliSoftware, line 140: Can be removed since it doesn't add anything.

config/Swift-Optimization-Debug.xcconfig
AliSoftware: Not sure I get the point of this extra `xcconfig` file. What's the goal of having that setting isolated here instead of declaring that setting directly in `Project-Debug.xcconfig` to apply it to the Debug config for all targets?

That's actually what you did with it for Release (`SWIFT_OPTIMIZATION_LEVEL = -O` in `Project-Release.xcconfig`), so not sure why it's different for debug.

config/Simplenote-AppStore-Debug.xcconfig
```diff
@@ -0,0 +1,36 @@
+// Originally generated by BuildSettingExtractor
+// https://github.com/dempseyatgithub/BuildSettingExtractor
+
+// Import the CocoaPods xcconfig first, so we can override what's needed
+#include "../Pods/Target Support Files/Pods-Automattic-Simplenote-AppStore/Pods-Automattic-Simplenote-AppStore.debug.xcconfig"
+
+#include "Simplenote-AppStore-Shared.xcconfig"
+#include "Swift-Optimization-Debug.xcconfig"
+#include "Code-Sign-Mac-Developer.xcconfig"
```
AliSoftware, line 9: For other configurations/targets, this one is `#include`'d in `Simplenote-Shared.xcconfig` and `SimplenoteTests-Shared.xcconfig`. So why only include it in `Simplenote-AppStore-Debug` here and not in the shared one like for the others?

AliSoftware (commented):
Finally found time to do a very thorough review of all those changes.

Here are my main feedback + some details about how I automated checking the changes made by this refactoring PR. Hopefully those "automated diff" tips techniques could be re-used to also review the other PRs we'll do for the similar xcconfig extractions in other repos.

## Comments in xcconfig

I'm convinced we should remove those comments which describe the build setting in a general way that can already be found online (as I explained in a comment below) like you did in #709.

That would also make the life easier for anyone trying to understand this new `xcconfig` structure/organization we introduce here, because it's way easier to understand the purpose of each `xcconfig` file if it only contains 2-3 lines overriding some build settings than if it contains 60 LoC with those interesting settings flooded in a sea of comments (maybe the fact that I'm reading those files from GitHub so without syntax highlighting doesn't help, but still)

## Automated diff techniques used to review this PR

### Comparing the output with `xcodeproj config-dump`

Instead of James' tool, I used the `xcodeproj config-dump` command line to try and do a similar extraction, then compared the results with what you got (after some renaming of the files so that they use the same names for `opendiff` to compare the right things together, and after removing comments and empty lines from each for easier `diff`).

_The `xcodeproj` command is shipped as part of the `xcodeproj` gem (which you probably already have installed as it's a dependency of `cocoapods`)._

Comparing the output xcconfigs that tool produces with what you had in #709 allowed me to spot those couple of `$(inherited)` entries and other comments I left.

### Comparing the resolved settings before/after

Another way I used to review those changes in some automated way is that I asked `xcodebuild` to dump the build settings in a JSON file, and ran that for all 3 schemes both in `develop` and in this PR's branch, then compared the resulting JSON. Since they matched (`diff` was empty except for `BUILD_NUMBER`/`VERSION_LONG` changes from `Version.public.xcconfig` which are expected) I'm pretty confident that this refactor won't break anything or change any compilation behavior.

```sh
$ mkdir config-diff

# Dump resolved settings before this PR's changes (develop)

$ git checkout develop

$ for scheme in Simplenote SimplenoteSecrets SimplenoteSecrets Simplenote-AppStore; do
for> xcodebuild -showBuildSettings -json -workspace Simplenote.xcworkspace -scheme $scheme >"config-diff/${scheme}.develop.json"
for> done

# Dump resolved settings after this PR's changes

$ git checkout extract-xcconfigs

$ for scheme in Simplenote SimplenoteSecrets SimplenoteSecrets Simplenote-AppStore; do
for> xcodebuild -showBuildSettings -json -workspace Simplenote.xcworkspace -scheme $scheme >"config-diff/${scheme}.pr.json"
for> done

# Print the diff for each

$ for scheme in Simplenote SimplenoteSecrets SimplenoteSecrets Simplenote-AppStore; do
for> diff "config-diff/${scheme}.pr.json" "config-diff/${scheme}.develop.json"
for> done
```

## Post-PR Cleanup

Some build settings might benefit to move around to higher levels. Especially, some settings were already probably not configured at the right level in the `pbxproj` before the extraction, but one of the side benefits of the extraction is it now allows us to see those inconsistencies more clearly.

For example, everything that should be applied to all targets in the Debug configuration should be in `Project-Debug.xcconfig`, including e.g. `DEBUG_INFORMATION_FORMAT` (that I've seen in e.g. `Simplenote-AppStore-Debug` but should probably be applied at the `Project-Debug` level instead)

If we clean up those and move some settings to higher levels where they were always due, ideally the end result would be that `{Scheme}-{Debug|Release|Shared}.xcconfig` files will likely end up only containing mostly `#include`s to just assemble the various parts (especially the right `xcconfig` from CocoaPods + the `Shared` xcconfig… and that's mostly it.

But those inconsistencies were already in the wrong place to begin with before the extraction, and might require a different discussion and separate review, so I think those cleanups should only be done in a separate PR on top of this one once this one is merged, to keep this one only about the automated extraction (and leaves the project in the exact same config state in the once once all is resolved, see `xcodebuild -showBuildSettings`) and keep the other one just about cleaning up.

mokagio:
Closing as stale.

PR #709, opened 2020-11-05
Extract all build settings to xcconfigs (without documentation)

mokagio:
Same as #707, but with no documentation comments to make it see if it's less noisy.


mokagio:
Closing as stale.

PR #715, opened 2020-11-13
Upgrade CocoaPods and remove the specs repo

jkmassel:
Uses the CocoaPods CDN instead of checking out the entire specs repo in order to speed up pod installation

### Fix
Removes the CocoaPods specs repo from the project in order to enable CI improvements across all projects. Required an update to CocoaPods at the same time.

### Test
Ensure all tests pass

### Review
Take a quick look at the code – the `Podfile` and `Podfile.lock` changes are fairly straightforward – the rest is the CocoaPods upgrade.

### Release


Gemfile
```diff
@@ -1,6 +1,6 @@
 source 'https://rubygems.org' do
   gem 'rake'
-  gem 'cocoapods', '1.6.1'
```
mokagio, line 3: This will make upgrading easier 👍 

For the usage we do of CocoaPods, staying on `~> 1` makes sense 👌

Simplenote.xcodeproj/project.pbxproj
```diff
@@ -84,7 +84,7 @@
 		37D4DD6A20B3574D00C225EA /* WPAuthHandler.m in Sources */ = {isa = PBXBuildFile; fileRef = 37D4DD6820B3574C00C225EA /* WPAuthHandler.m */; };
 		37F742EB202A382400A47D3A /* AboutViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 37F742EA202A382400A47D3A /* AboutViewController.swift */; };
 		37F742EC202A382400A47D3A /* AboutViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 37F742EA202A382400A47D3A /* AboutViewController.swift */; };
-		3FAC7B25254BBCA600B47276 /* SimplenoteInterlinks in Frameworks */ = {isa = PBXBuildFile; productRef = 3FAC7B24254BBCA600B47276 /* SimplenoteInterlinks */; };
```
mokagio: This happens to me at times, but I don't think it should be committed. I'm not 100% sure how it happens, a guess would be something to do with how the gem CocoaPods uses to touch the project file.

If I recall correctly, either a clean build from Xcode or a new CocoaPods run should restore this to its original value.

The reason I'm saying this should not be committed is that I'm pretty sure from personal experience that Xcode will at some point revert this change. So, better avoid noise and confusion to whoever will pull these changes.

Simplenote.xcodeproj/project.pbxproj
```diff
@@ -1590,26 +1590,11 @@
 			files = (
 			);
 			inputFileListPaths = (
-			);
```
mokagio, line 1593: TIL about `xcfilelist`. I never noticed it before.

Simplenote.xcodeproj/project.pbxproj
mokagio: Update. Of course, now that I said "Xcode will revert it at some point" I'm not able to trigger that change 😳 .

I still think this is incorrect. Here's two data points in favor of this hypothesis:

- [Tuist](https://tuist.io/, an open source Xcode project generator, has [a type for `XCRemoteSwiftPackageReference`](https://github.com/tuist/XcodeProj/blob/82bf5efcaa27e94ed8c761c1eb3e397b6dea82b9/Sources/XcodeProj/Objects/SwiftPackage/XCRemoteSwiftPackageReference.swift) but no type for `RemoteSwiftPackageReference`
- Someone asked something similar to this [on StackOverflow](https://stackoverflow.com/questions/59830330/swiftlint-overriding-project-settings-related-to-spm)

mokagio (commented):
Other than for the project file thing, this looks good.

Simplenote.xcodeproj/project.pbxproj
AliSoftware: So, I can confirm that the format that Xcode uses to generate the pbxproj might sometimes slightly differ from what CocoaPods, Tuist, XcodeGen or others generate, because Xcode's `pbxproj` generator is private implementation of course…

The usual recommendation I'd do is indeed to modify something in the Xcode project (e.g. move a file in one of the groups in the file navigator to reorder the files, then move it back to its original position) then save the project via Xcode UI to restore what Xcode generates at the end. No real magic solution for this unfortunately.

Simplenote.xcodeproj/project.pbxproj
AliSoftware: They are very useful and a welcome addition from Xcode 11!

_FYI, SwiftGen has migrated to them now too, so that you can make the `swiftgen` Script Build Phase only run if one of the file listed in the `xcfilelist` has changed, and SwiftGen itself is able to generate that `xcfilelist` text file from its config file and thus from the input files the codegen ends up depending on. All that without SwiftGen having to modify the `pbxproj`, just an external `xcfilelist`… very useful!_

Gemfile
AliSoftware: `~> 1` doesn't really do anything (might as well remove the version constraint completely) because it means "any version >=1".

The `~>` operator means "you're only allowed to pick a version where the **last** digit is bigger". So e.g. `~> 1.6` would mean "1.6 or 1.7 or anything 1.x where x>=6; but not 2.x". `~> 1.2.3` would mean "any 1.2.x where x>=3, but not 1.3 nor newer". So here `~> 1` just means "any x where x>1", which is basically… any version of cocoapods (now that it has been out of its 0.x phase for ages)

Not sure if what we would prefer in our case is `~>1.9` (any 1.x that is at least 1.9 but not 2.x) or just nothing (which is what `~> 1` is in practice, = any version). I think no constraint at all makes more sense, because with a `Gemfile` and a `Gemfile.lock` in place, we only update the gems like CocoaPods explicitly (`bundle update cocoapods`) and it will stay locked on the version in the lockfile otherwise so we should be good.

Simplenote.xcodeproj/project.pbxproj
```diff
@@ -84,7 +84,7 @@
 		37D4DD6A20B3574D00C225EA /* WPAuthHandler.m in Sources */ = {isa = PBXBuildFile; fileRef = 37D4DD6820B3574C00C225EA /* WPAuthHandler.m */; };
 		37F742EB202A382400A47D3A /* AboutViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 37F742EA202A382400A47D3A /* AboutViewController.swift */; };
 		37F742EC202A382400A47D3A /* AboutViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 37F742EA202A382400A47D3A /* AboutViewController.swift */; };
-		3FAC7B25254BBCA600B47276 /* SimplenoteInterlinks in Frameworks */ = {isa = PBXBuildFile; productRef = 3FAC7B24254BBCA600B47276 /* SimplenoteInterlinks */; };
+		3FAC7B25254BBCA600B47276 /* BuildFile in Frameworks */ = {isa = PBXBuildFile; productRef = 3FAC7B24254BBCA600B47276 /* SwiftPackageProductDependency */; };
```
AliSoftware: 🤔  A mention of a generic "BuildFile" name here is usually the sign of a corruption in the pbxproj (typically a file that was referenced by a UUID in this Build Phase but the file and its UUID were removed from the Objects List at the top of the pbxproj but the UUID pointing to it wasn't – probably the result of conflict resolution gone bad).

In such cases, there is usually a "BuildFile" entry with a generic document icon when you open the corresponding build phase in Xcode's UI, or even sometimes there's just an empty line (that mostly happen when the entry is `(null)` in the pbxproj tho, not `BuildFile`)

Simplenote.xcodeproj/project.pbxproj
```diff
@@ -210,12 +210,12 @@
...
+		B5609AEA24EEC9910097777A /* BuildFile in Frameworks */ = {isa = PBXBuildFile; productRef = B5609AE924EEC9910097777A /* SwiftPackageProductDependency */; };
 		B5609AEC24EEE7200097777A /* SPBucket+Simplenote.swift in Sources */ = {isa = PBXBuildFile; fileRef = B5609AEB24EEE7200097777A /* SPBucket+Simplenote.swift */; };
 		B5609AED24EEE7200097777A /* SPBucket+Simplenote.swift in Sources */ = {isa = PBXBuildFile; fileRef = B5609AEB24EEE7200097777A /* SPBucket+Simplenote.swift */; };
-		B5609AF024EF171D0097777A /* SimplenoteFoundation in Frameworks */ = {isa = PBXBuildFile; productRef = B5609AEF24EF171D0097777A /* SimplenoteFoundation */; };
-		B5609AF224EF17270097777A /* SimplenoteFoundation in Frameworks */ = {isa = PBXBuildFile; productRef = B5609AF124EF17270097777A /* SimplenoteFoundation */; };
+		B5609AF024EF171D0097777A /* BuildFile in Frameworks */ = {isa = PBXBuildFile; productRef = B5609AEF24EF171D0097777A /* SwiftPackageProductDependency */; };
+		B5609AF224EF17270097777A /* BuildFile in Frameworks */ = {isa = PBXBuildFile; productRef = B5609AF124EF17270097777A /* SwiftPackageProductDependency */; };
```
AliSoftware: ditto, we should double-check those in Xcode UI, those feels wrong and result of corrupted merge conflict resolution or similar

Gemfile
jkmassel: I locked this to `1.9 or greater` for now – thanks for the reminder 😁

Simplenote.xcodeproj/project.pbxproj
jkmassel: Nice fix - TIL about that trick.

Simplenote.xcodeproj/project.pbxproj
jkmassel: No merge conflict, just CocoaPods breaking things. Your fix above definitely resolved it though which was super.

Simplenote.xcodeproj/project.pbxproj
mokagio: Neat

mokagio (approved):
Thanks for addressing the Xcode project stuff. Everything looks good now.

Issue #718, opened 2020-11-18
Big Sur: UX Refresh


jleandroperez:
### Tasks:
- [x] Drop Titlebar. YAY! #719
- [x] Refresh Colors #733
- [x] Big Sur Icons #745
- [x] Tags: Update Initial padding.top #719
- [x] TableRowView: New Styles #721
- [x] Notes: Replace SearchBar with Section Title #722
- [x] Notes: Top Blur! #724
- [x] Notes: Row Separators #758
- [x] Toolbar: Top Blur! #724
- [x] Toolbar: Implement Search #722
- [x] Toolbar: Implement Sidebar Action #737 
- [x] Toolbar: Search Bar Expansion Animation #741
- [x] Toolbar: New Ellipsis Outline #720
- [x] Bug: Info Sheet Overlap #725
- [x] Bug: Limit Notes List Title width #726
- [x] Bug: Editor not opening notes at the top position #728
- [x] Bug: Tags List getting clipped at 120pt #727
- [x] Bug: Catalina / Notes List not reaching the upper edge #731
- [x] Bug: Disable Desktop Tinting in the Notes List / White Mode #732
- [x] Bug: Fix collision between Title / Semaphore #734
- [x] Bug: Window Semaphore's Height #746
- [x] Bug: Multiple Selection #754
- [x] Bug: Tag List Cursor #774
- [x] Bug: Scroller appears cut off (main editor) #777
- [x] Bug: SplitView Dimple Artifact #779
- [x] Bug: Semaphore Buttons not 100% aligned with regards to the title (macOS Catalina) #787

### Mockups:

<img width="300" src="https://user-images.githubusercontent.com/1195260/99596465-11e62380-29d5-11eb-968f-e64b59539bb9.png"> <img width="300" src="https://user-images.githubusercontent.com/1195260/99596481-17dc0480-29d5-11eb-8b1d-ee1ec60f21a5.png"> <img width="300" src="https://user-images.githubusercontent.com/1195260/99596482-190d3180-29d5-11eb-86b1-41b1aa3e77ba.png">

<img width="300" src="https://user-images.githubusercontent.com/1195260/99596478-17436e00-29d5-11eb-8d4c-d5d6a4f9dcbb.png"> <img width="300" src="https://user-images.githubusercontent.com/1195260/99596477-1579aa80-29d5-11eb-8b6a-5211d911f5ad.png">

jleandroperez:
[...]

PR #719, opened 2020-11-18
Main Window: Drops titlebar

jleandroperez:
### Fix
In this PR we're dropping the main window's titlebar:

- Updated Top Bar's height to 52pt
- Tags List: New Top Separator

**Known Issues:**
Collapsing the Tags List is causing a collision between the Window's semaphore and the Search Bar. We're fixing that soon!

cc @eshurakov (Thank youuuu!!)
Ref. #718

### Test
1. Log into Simplenote!

- [x] Verify the app looks great!

### Release
`RELEASE-NOTES.txt` was updated in 393af7c with:

> Removed Main Window's Titlebar #719

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,5 +1,6 @@
 2.5
 -----
+- Removed Main Window's Titlebar #719
 
 2.4
 ---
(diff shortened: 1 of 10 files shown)


eshurakov (approved):
Looks great without a titlebar!

One thing that I noticed and not sure if it is related to this PR: even though there is plenty of space, some labels in the sidebar are truncated.

Screenshot 2020-11-23 at 20 44 35

jleandroperez:
Thank you for finding that bug @eshurakov !!! (and thank you for cracking the fix!!!).

For the record: sending bugfixes in a PR, on top of all the changes, to avoid storyboard merge conflicts.

Again, thank you sir!

PR #727, opened 2020-11-25
Taglist: Fixing Resize

jleandroperez:
### Fix
In this PR we're (aiming!!) at fixing a bug in which the Tag List cells would not resize.

Ref. #718

### Test
1. Launch the app in macOS Catalina. Specifically in Evgeny's mac!

- [ ] Verify the Tags List looks good!

### Release


jleandroperez:
@eshurakov Mind taking a second look please? 

**I THINK** it should be good now. Thank you!!!

eshurakov:
@jleandroperez I checked and unfortunately tag names are still clipped :-(

jleandroperez:
@eshurakov I'm a bit lost on this one 😄 

I've compared (side by side) IB's layout settings from 2.4 vs current... and reverted back to `Layout > Autosizing Mask`.

Flying a bit blind over here!! would it be possible to check again?.

Thank you!!!

eshurakov:
@jleandroperez It's a mystery why it doesn't work on my machine.
I've tried different ways to fix this and there are two solutions that worked for me. I'm not proficient in macOS so not sure if it breaks somethings else :-)
1. Add `sizeLastColumnToFit()` to `ensureStyleIsFullWidth` method of `NSTableView` extension.
2. I went to `Main.storyboard` and changed `Tags Column` width to 150. After that I changed the width of the following views to 150 as well: `view`, `Background Visual Effects View`, `Scroll View`, `Header Visual Effects View`, `Header Separator View`.

loremattei:
Hey there! 2.5 is now in code freeze. If you want this PR to make it to 2.5 (and I guess we want to :-) ), please, feel free to update it to target the release branch and to ping me to build a new beta.

jleandroperez:
@eshurakov Thank you for checking this out!!

I'm afraid I can't repro the original issue on my end (with a debugger!!). I ended up implementing both of your suggestions. 

`sizeLastColumnToFit` behaves, pretty much, like a `sizeToFit`, so your idea is right on spot.

Mind taking another look? Thanks in advance!!

eshurakov (approved):
It woooorks! 🎉 

:shipit:  :-)

jleandroperez:
Thanks everyone!!

PR #734, opened 2020-12-02
Preventing Title Overlap

jleandroperez:
### Fix
In this PR we're adding a new Layout Constraint that prevents the **Notes List Title** from overlapping with the Window's semaphore (Close / Minimize / Zoom buttons).

@eshurakov Almost there!! Thanks in advance!!

Ref. #718

### Test: LTR
1. Launch the app in **Left to Right** mode

- [x] Verify the Notes List title looks great when the three panels are visible (Tags List / Notes List / Editor)
- [x] Verify that when the Tags List is closed, the Notes List title stops sliding as soon as the Window Semaphore is reached
- [x] Verify that going into fullscreen mode causes the Title to reach the edge of the screen (the buttons aren't really visible!)

### Test: RTL
1. Launch the app in **Right to Left** mode

- [x] Verify that collapsing the Tags List causes the Notes List title to avoid the Window Semaphore area
- [x] Verify the Notes List title has no padding when you go fullscreen

### Release

### Screenshots

LTR|RTL
-|-
LTR | RTL


eshurakov (approved):
Looks good! Tested it on Catalina 👍 

:shipit:

jleandroperez:
Thank you Evgeny!!!

jleandroperez:
@loremattei Sir!! Would it be possible to get a new Simplenote 2.5 build?

**Thank youuuu!!**

PR #743, opened 2020-12-08
Fix for archive failure

mokagio:
### Fix
This PR updates the pods to versions that compile under Xcode 12.2 after upgrading to Tracks 0.6.0. (I think the Xcode version is the reason why the archiving fails)

### Test
Checkout this branch, run `bundle install && bundle exec pod install`, and Archive both Simplenote and Simplenote-AppStore targets.

### Review
Only one platform developer is required to review these changes, but anyone can perform the review.

### Release

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org' do
   gem 'rake'
-  gem 'cocoapods', '~> 1.8'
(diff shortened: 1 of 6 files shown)


mokagio:
Closing.

I picked up a simpler workaround for this, d013e6f5f86a0ec25ee925baa8a2bd76e4ebf0b3, as part of #749.

PR #752, opened 2020-12-11
Debt: Initial Setup Flow

jleandroperez:
### Fix
In this PR we're porting over to Swift the Welcome Note flow, and wiring Options usage (NSUserDefaults wrapper).

cc @eshurakov Slowly swifting the AppDelegate!

Thank you sir!!

### Test: First Install
1. Delete the macOS Welcome Note from your Simplenote account
2. Make sure you empty the Trash bin
3. In ZSH run `defaults delete com.automattic.SimplenoteMac.beta SPFirstLaunch`
4. Relaunch the app

- [x] Verify the Welcome Note shows up and it's properly localized in your local machine's language

### Test: Deletion
1. Launch the app
2. Delete the macOS Welcome Note from your Simplenote account
3. Make sure you empty the Trash bin
4. **Relaunch!**

- [x] Make sure the **Welcome Note** does not show up again!

### Release


Simplenote/SimplenoteAppDelegate+Swift.swift
```diff
@@ -78,6 +78,41 @@ extension SimplenoteAppDelegate {
...
 
+// MARK: - Welcome Note
+//
+extension SimplenoteAppDelegate {
+
+    @objc
+    func configureWelcomeNoteIfNeeded() {
+        if Options.shared.initialSetupComplete {
+            return
+        }
```
eshurakov, line 89: No `guard`? 🤔

jleandroperez:
### Update:
For the record, @eshurakov brought up a potential Sync'ing glitch caused by the way the Welcome note is handled (Fixed ID).

Following that up in this P2 reference: `p2XJRt-2y3-p2`

I'll patch the mechanism in a follow up, let's keep this PR as is, since it's a direct port of the current behavior.

**Thank you!!**

eshurakov (approved):
Looks good!

When I removed the `SPFirstLaunch` key from user defaults and ran the app, there is a spinner that never goes away.

Screenshot 2020-12-29 at 20 48 26

Simplenote/SimplenoteAppDelegate+Swift.swift
jleandroperez: There was an article about this one (can't find it anywhere!!) but the idea is: in some scenarios, a simple `if` might be more readable:

```
guard !Options.shared.initialSetupComplete else {
    return
}
```

Our brains "needs to parse" **NEGATION [CONDITION], ELSE**. In this scenario, a simple `if { }` feels super readable and natural, didn't wanna force things!

jleandroperez:
Thank you @eshurakov !! 

RE: Spinner, expected (missed mentioning that!!). The First Launch involves a reindex OP, which, when messing with User Defaults, never really happens.

🔥

Simplenote/SimplenoteAppDelegate+Swift.swift
eshurakov: I see, makes sense! Thx :-)

PR #755, opened 2020-12-14
Notes List: New Sort Modes

### LINKED ISSUE #33: Note List Sorting: Not consistent with other clients
<a href="https://github.com/roundhill"></img></a> **Issue by [roundhill](https://github.com/roundhill)**
_Friday Dec 19, 2014 at 16:54 GMT_

---

We have added alpha sort, but users might want to sort Ascending/Descending or by created date.

jleandroperez:
### Fix
In this PR we're implementing new Sort Modes in the Sort List, so that we match iOS, Electron and Android:

- Alphabetically: A-Z / Z-A
- Creation Date: Newer / Older
- Modification Date: Newer / Older

@SylvesterWilmott Matching Electron's Menu. WDYT? (Thanks in advance!!)

cc @eshurakov (Thankyooou)

Closes #33
Ref. #626

### Test: Migration / Updated
1. Install **develop** in your mac and authenticate
2. Click over `View > Note Sorting > Date Updated`
3. Install this branch on top

- [x] Verify `Note Sorting > Modified: Newest` is preselected for you

### Test: Migration / Alpha
1. Run this in bash
  ```defaults delete com.automattic.SimplenoteMac.beta notesListSortMode
   defaults write com.automattic.SimplenoteMac.beta kAlphabeticalSortPreferencesKey 1
  ```
2. Run this branch on top

 - [x] Verify `Note Sorting > Alphabetically: A-Z` is preselected

### Test: Sort Modes
1. Run this branch in your mac
2. Launch iOS as well!

- [x] Verify the setting `Note Sorting > Alphabetically: A-Z` matches iOS
- [x] Verify the setting `Note Sorting > Alphabetically: Z-A` matches iOS
- [x] Verify the setting `Note Sorting > Created: Newest` matches iOS
- [x] Verify the setting `Note Sorting > Created: Oldest` matches iOS
- [x] Verify the setting `Note Sorting > Modified: Newest` matches iOS
- [x] Verify the setting `Note Sorting > Modified: Oldest` matches iOS

### Release
`RELEASE-NOTES.txt` was updated in 536257b with:
 
> New Notes List Sort Modes #33

### Screenshots
Before|Now
-|-
Screen Shot 2020-12-14 at 5 31 09 PM|Screen Shot 2020-12-15 at 12 20 01 PM


SylvesterWilmott:
I'm not sure if I like the "Reversed" option because it would also require you to open the menu twice in order to select Reversed Date Created for example. Can we list all 6 options in the same way we have on iOS instead?

jleandroperez:
@SylvesterWilmott I had doubts as well (just went with the Electron approach to try and match all the things!!). But having one item per actual Sort Mode is way more convenient.

PR Updated (and new Screenshot attached), thank youuuu!!

SylvesterWilmott (approved):
Thanks @jleandroperez

jleandroperez:
Thanks for checking this one Sly!!

Simplenote/SortMode.swift
```diff
@@ -0,0 +1,91 @@
...
+            return .noteSortAlphaDescMenuItem
+        case .createdNewest:
+            return .noteSortCreateNewestMenuItem
+        case .createdOldest:
+            return .noteSortCreateOldestMenuItem
+        case .modifiedNewest:
+            return .noteSortModifyNewestMenuItem
+        case .modifiedOldest:
+            return .noteSortModifyOldestMenuItem
+        }
```
eshurakov, line 89: Idea: what if we create a map and use it from both methods instead of doing switch?
```
sortModeToItem = [
    alphabeticallyAscending: noteSortAlphaAscMenuItem,
    alphabeticallyDescending: noteSortAlphaDescMenuItem,
    ....
]
```

eshurakov (approved):
Looks good!

:shipit:

Simplenote/SortMode.swift
jleandroperez: In this kind of mapping it's best to avoid maps, since whenever a new case is added (if ever) you'll get a compiler warning for free for all of the missing cases.

In the Map realm, instead, the compiler goes silent, and you're dealing with a runtime problem instead.

Thank you sir!!

PR #763, opened 2020-12-18
Implementing NotesListController: Mark II

jleandroperez:
### Fix
In this PR we're building NotesListController, which wraps up access to NSFetchedResultsController.
Plus:

- Builds **MockupStorage**: unit testing CoreData stack
- **NotesListController** gets **a million** Unit Tests. 
- Updates the **Testing macOS Deployment Target** to 10.15, in order to fix an annoying warning :o

🏳️🏳️ About 70% of the code in here are related to Unit Testing :: Apologies about the magnitude of this PR!! 🏳️🏳️
 
Ref. #626

cc @eshurakov (Thank yooouuu!!)

### Test
- [ ] Verify the tests are green
- [ ] Verify the code makes sense!

### Release


Simplenote/NotesListController.swift
```diff
@@ -0,0 +1,213 @@
+import Foundation
+import CoreData
+import SimplenoteFoundation
+import SimplenoteSearch
+
+
+// MARK: - NotesListController
+//
+class NotesListController: NSObject {
```
eshurakov, line 9: I think that correct English here is `NoteListController` 🤔 
But not totally sure :-)

Simplenote/NotesListController.swift
```diff
@@ -0,0 +1,213 @@
...
+
+    /// Returns the index of a given note (if any)
+    ///
+    func indexOfNote(withSimperiumKey key: String) -> Int? {
+        for (index, note) in notesController.fetchedObjects.enumerated() where note.simperiumKey == key {
+            return index
+        }
+
+        return nil
+    }
```
eshurakov: Maybe `return notesController.fetchedObjects.firstIndex(where: { note.simperiumKey == key })` ?

Simplenote/NotesListController.swift
```diff
@@ -0,0 +1,213 @@
...
+    @objc
+    var numberOfNotes: Int {
+        notesController.numberOfObjects
+    }
+
+    /// Returns all of the Retrieved Notes
+    ///
+    var retrievedNotes: [Note] {
+        notesController.fetchedObjects
+    }
```
eshurakov: Maybe just `var notes: [Note]`?

Simplenote/NotesListController.swift
```diff
@@ -0,0 +1,213 @@
...
+        return index >= .zero  && index < fetchedObjects.count ? fetchedObjects[index] : nil
+    }
+
+    /// Returns the Fetched Note with the specified SimperiumKey (if any)
+    ///
+    func note(forSimperiumKey key: String) -> Note? {
+        notesController.fetchedObjects.first { note in
+            note.simperiumKey == key
+        }
+    }
```
eshurakov: Maybe we can use `indexOfNote(withSimperiumKey:)` and `note(at index:)` here to get the note?

Simplenote/NotesListController.swift
```diff
@@ -0,0 +1,213 @@
...
+    ///
+    var searchSortMode: SortMode = .alphabeticallyAscending {
+        didSet {
+            guard case .searching = state, oldValue != searchSortMode else {
+                return
+            }
+
+            refreshSortDescriptors()
+        }
+    }
```
eshurakov, line 85: Curious, do we really need two different variables for sort modes?

Simplenote/NotesListController.swift
```diff
@@ -0,0 +1,213 @@
...
+        }
+        set {
+            guard let keyword = newValue, !keyword.isEmpty else {
+                state = .results
+                return
+            }
+
+            state = .searching(keyword: keyword)
+        }
+    }
```
eshurakov, line 61: Wondering if we have to have `state`, `filter` and `searchKeywords`. NoteListController represents a list of notes which is possible to sort and filter. For that purpose I think it should be enough to have only `var filter: NotesListFilter` and `var sortMode: SortMode`. This way we would make this class simpler and more straightforward to use.
For example combining `NoteListState` and `NoteListFilter` we can have:
```
enum NoteListFilter: Equatable {
    case everything
    case deleted
    case untagged
    case tag(name: String)
    case keywords(keywords: String)
}
```

Simplenote/NotesListController.swift
```diff
@@ -0,0 +1,213 @@
...
+                return
+            }
+
+            refreshPredicates()
+        }
+    }
+
+    /// Refreshes the Internal State, so that Search Results matching the specified Keyword will be filtered out.
+    /// - Important: It's up to the caller to invoke `performFetch`!!
+    ///
```
eshurakov, line 44: > Important: It's up to the caller to invoke `performFetch`!!

Why is that? :-) Shouldn't it be automatically triggered because `state` will change?

Simplenote/NotesListController.swift
```diff
@@ -0,0 +1,213 @@
...
+                return
+            }
+
+            refreshSortDescriptors()
+        }
+    }
+
+    /// Relays back willChangeContent Events
+    ///
+    var onWillChangeContent: (() -> Void)?
```
eshurakov, line 89: How will we use `onWillChangeContent`? Just wondering if we need it.

SimplenoteTests/MockupStorage.swift
```diff
@@ -0,0 +1,115 @@
...
+                guard let error = error else {
+                    return
+                }
+
+                fatalError("☠️ [MockupStorage] Unable to regenerate Persistent Store! \(error)")
+            }
+
+            NSLog("💣 [MockupStorage] Stack Destroyed!")
+        }
+    }
```
eshurakov: Will it be easier to nuke the `persistentContainer` var and create a new one? I think it will be less code :-)

SimplenoteTests/NotesListControllerTests.swift
```diff
@@ -0,0 +1,495 @@
...
+
+    override func setUp() {
+        super.setUp()
+        noteListController = NotesListController(viewContext: storage.viewContext)
+        noteListController.performFetch()
+    }
+
+    override func tearDown() {
+        super.tearDown()
+        storage.reset()
```
eshurakov: I'm actually not sure if we even need to reset. It seems that all stored vars are recreated by Xcode between tests.
I added this to `MockupStorage` and ran `NotesListControllerTests`
```
init() {
        print("\(self)")
}
```

It was printed multiple times, so that a new MockupStorage was created for each test.

SimplenoteTests/NotesListControllerTests.swift
```diff
@@ -0,0 +1,495 @@
+import XCTest
+import SimplenoteFoundation
+@testable import Simplenote
+
+
+// MARK: - NotesListControllerTests
+//
+class NotesListControllerTests: XCTestCase {
```
eshurakov, line 8: I think these tests are great as an integration tests, but I don't think strictly speaking we can call them unit tests. 

I will try to explain why I think this way. 

Right now we're testing not only the logic of `NoteListController`, but also it's collaborators:
* `ResultsController` and in the end CoreData that is used to fetch data
* `NoteListState` which is used to generate predicates

So along with `NoteListController` we also test that core data fetches the right entities and `NoteListState` produces correct predicates. But is it what we really want to test here? I would say no. I hope that Core Data is already tested and when provided with a correct predicate will return correct entities.

If we test `ResultsController` and `NoteListState` separately and we know that they work, then when testing `NoteListController` we can fully mock them and test that during certain code path certain methods are invoked.

Example of some tests I would write to test the behaviour of `NoteListController`:
* When we init `NoteListController`, then `onWillChangeContent` and `onDidChangeContent` is set in `ResultsController`
* When `ResultsController` calls `onWillChangeContent` / `onDidChangeContent`, then respective callbacks from `NoteListController` are invoked. The actual content `objectsChangeset` is not relevant here because `NoteListController` doesn't do anything with it, we just want to make sure that it is passed along.
* When filter is set on `NoteListController`, then we check that we get predicates from `NoteListState` and set them on `ResultsController`. Again for `NoteListController`  it is not really important which filter is chosen, because it's job is to take the predicates for this filter from the state and set it to `ResultsController`

I think this way we can improve by only testing behaviour of one object at a time (if possible).

What do you think? I can elaborate more on how I see this :-)

eshurakov (commented):
Sending some comments, let me know what you think!

Simplenote/NotesListController.swift
jleandroperez: I tend to avoid APIs that use comparator blocks (because, unless Swift optimizes, they should have a performance hit).

Nevertheless, compact wins, updated! Thanks for the suggestion!

Simplenote/NotesListController.swift
jleandroperez: Done, ty!

Simplenote/NotesListController.swift
jleandroperez: Both Sort Modes will map to different User Defaults properties (this will be wired by the Sort Bar).

It's up to the ListController which one should be applied (otherwise we'd need to check on the state of the ListController before assigning the property).

Simplenote/NotesListController.swift
jleandroperez: If you don't mind, let's keep this name, beeeecause fixing the other 3 PRs is gonna be a nightmare
(AND) if singular is the actual correct english version, I should also rename on iOS 😂

SimplenoteTests/MockupStorage.swift
jleandroperez: You're right, simplified!

SimplenoteTests/NotesListControllerTests.swift
jleandroperez: AND You're right, simplified as well!!

Simplenote/NotesListController.swift
jleandroperez: We'll use this to preserve the selected indexes / notes, before applying an update.

Tricky, I know!! [Ref. Here!](https://github.com/Automattic/simplenote-macos/pull/764/files#diff-973bc837cd1ca5e240a9782bed8d5e75e5c51c4e9adce43d248f8844aed6cc7eR197)

Simplenote/NotesListController.swift
jleandroperez: 100% on purpose, in order to avoid a duplicate `performFetch()`.

**Will be revisited** up next, I'm merging NoteListFilter + NoteListState, as per your suggestion.

Thank you for spotting this!!

Simplenote/NotesListController.swift
jleandroperez: (Discussed over Slack!!). **BEAUTIFUL idea**, following up in a PR aiming at the tail of this chain, in order to avoid breaking everything that comes up next!!

eshurakov (approved):
🚢 :shipit: 🚀

jleandroperez:
For the record:

- NoteListController renamed in #780
- NoteListFilter + NoteListState merged in #781 
- Unit Tests decoupling in progress!

Thank you for reviewing this one Evgeny!!

PR #765, opened 2020-12-22
Implementing NotesListController: Mark IV

jleandroperez:
### Details:
1. **NoteListViewController** is now 100% Swift based
2. Updated **NoteListViewController**'s method visibility
3. Dropped a ton of `@objc` modifiers
4. **TagListViewController** no longer tampers with the NoteList's TableView selected index

cc @eshurakov 🔥 
Thank you Evgeny!!

Ref. #626

### Test: Regressions
- [ ] Verify that during LogIn a spinner shows up in the Notes List (and gets hidden once sync is done)
- [ ] Verify that the Editor isn't editable (and the Simeperium logo shows up) while you're Logging In
- [ ] Verify that right after launch the Note List Title is correctly set
- [ ] Verify that switching to Light / Dark Mode preserves your selected **Tag** / **Note**
- [ ] Verify the **New Note** toolbar action is disabled when visualizing the Trash
- [ ] Verify the **New Note** toolbar action is disabled when visualizing anything that's not Trash
- [ ] Verify that pressing the **New Note** toolbar action effectively adds a new Note
- [ ] Verify clicking over any Tag causes the NoteList Title to update
- [ ] Verify changing the `View > Note Sorting` mode (to any value) causes the Notes List to immediately react
- [ ] Verify changing the `View > Note Display` mode (to any value) refreshes the way the Cells are rendered
- [ ] Verify the NoteList Header (blur on top, title area!) fades in as you scroll
- [ ] Verify that pressing **backspace** whenever any note is highlighted moves such document to the Trash
- [ ] Verify you can Delete Tags from the Tags List, and no console exception shows up

### Test: TagList Reload
1. Select any `Tag A` and add a new `Note B`
2. Click over the Tags Editor area
3. Type a new (random) tag name and hit **return** to confirm

- [ ] Verify the new Tag immediately shows up in the Tags List
- [ ] Verify the `Tag A` remains selected
- [ ] Verify the `Note B` remains onscreen!

### Test: TagList Sort Mode
1. Select any tag from your list
2. Toggle the TagList Sort Mode (`View > Tag Sorting > Sort Alphabetically`)

- [ ] Verify the Tag List gets immediately reordered
- [ ] Verify the Notes List matches the contents of the newly selected Tag

### Release


Simplenote/TagListViewController+Swift.swift
```diff
@@ -87,6 +112,12 @@ extension TagListViewController {
         let absoluteOffSetY = scrollView.documentVisibleRect.origin.y + clipView.contentInsets.top
         return min(max(absoluteOffSetY / SplitItemMetrics.headerMaximumAlphaGradientOffset, 0), 1)
     }
+
+    private func notifyTagsListFilterDidChange() {
+        let isViewingTrash = selectedNotesFilter == .deleted
+        let name: NSNotification.Name = isViewingTrash ? .TagListDidBeginViewingTrash : .TagListDidBeginViewingTag
```
eshurakov, line 118: Wondering if we need separate notifications here? 🤔

eshurakov (approved):
Yesss, :shipit:

Simplenote/TagListViewController+Swift.swift
jleandroperez: Can be definitely simplified, but since we had to patch the Editor as well, I kept the change as small as possible.

I'll be revisiting this, if FRC goes as expected, the TagList should get the same refactor! 🔥

jleandroperez:
Thank you @eshurakov !!!

PR #768, opened 2020-12-24
Fixing Unexpected TagsList Jump

### LINKED ISSUE #767: Unexpected Tags List Jump
### Details:
1. Log into an account that has a Tags List long enough that it requires scrolling
2. Click over any Note
3. In the Tags Editor area (below the text editor) type a new random tag name
4. Scroll all the way down in the Tags List
5. Attempt to click any tags at the bottom

As a result, the tags list will unexpectedly jump, and a tag (different from the one you tried to click) will end up highlighted.

### Demo
Bug

jleandroperez:
### Fix
In this PR we're fixing a bug that prevented users from selecting any Tags at the bottom of the screen, right after editing the Note's Tags.

cc @eshurakov (Thank youuuuu!!!)
Fixes #767

### Details
This bug was caused by this check:

```
[note hasTag:appDelegate.selectedTagName] == false
```

When the `All Notes` filter is selected, the `selectedTagName` will be null.
Since I was in the neighborhood, figured, Swifting is a neat idea!

### Test: Jumpy Scenario
1. Log into an account that has a Tags List long enough that it requires scrolling
2. Click over **All Notes** and select any notes in the list
3. In the Tags Editor area (below the text editor) type a new random tag name
4. Scroll all the way down in the Tags List
5. Click over any tags at the bottom of the list

- [ ] Verify the Tag you've clicked actually gets highlighted!
- [ ] Verify the Tag you've added in (3) shows up in the Tags List

### Test: Dropping a Tag
1. Click over any Tags that contain a note
2. Select any note
3. Remove the selected Tag Name from the Tags Editor

- [ ] Verify the app immediately jumps over to **All Notes**
- [ ] Verify the same note remains onscreen

### Release
`RELEASE-NOTES.txt` was updated in 247b24f with:

> Fixed a bug that caused an unexpected jump in the Tags List #767

### Demo: Buggy Behavior
Bug

### Demo: Fixed Behavior
Fixed

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -9,6 +9,7 @@
 - Notes List now display row separators #758
 - Fixed a bug that affected Multiple Selection in Big Sur #753
 - Notes List will now animate content changes #764
+- Fixed a bug that caused an unexpected jump in the Tags List #767
  
 2.5
 -----
--- a/Simplenote/NoteEditorViewController+Swift.swift
+++ b/Simplenote/NoteEditorViewController+Swift.swift
@@ -137,6 +137,18 @@ extension NoteEditorViewController {
     var isSelectingText: Bool {
         noteEditor.selectedRange().length != .zero
     }
+
+    /// Simperium 🖖
+    ///
+    var simperium: Simperium {
+        simplenoteAppDelegate.simperium
+    }
+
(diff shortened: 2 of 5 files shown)


eshurakov (approved):
🚀 
:shipit:

jleandroperez:
Thank yooouuuu!!

PR #772, opened 2020-12-28
Empty Document: Line Height

### LINKED ISSUE #771: Empty State Typing Attributes
### Steps
1. Add a new note
2. Notice the cursor's height
3. Type anything

As a result, the cursor's height will match the Heading Style font size. Deleting the new character will cause the cursor's height to shrink back again.

**Expected:**
Empty State's cursor height should be constant at all times.

### Demo
Demo

jleandroperez:
### Fix
In this PR we're overriding the **typingAttributes** NSTextView API, so that when the document being edited is 100% empty, we use the Heading attributes.

As a result, the cursor's height (on empty documents) will match the first line's font size, at all times.

@eshurakov Spamming you with PRs!! 😂 
Closes #771

### Test
1. Add a new note
2. Notice the cursor's height
3. Type anything

- [ ] Verify the line height in (2) and (3) never changes!

### Release
`RELEASE-NOTES.txt` was updated in e5aa797 with:

> Fixed the Line Height when editing an empty document #771

### DIFF
--- a/External/Notepad/Storage.swift
+++ b/External/Notepad/Storage.swift
@@ -194,6 +194,17 @@ class Storage: NSTextStorage {
 }
 
 
+// MARK: - Typing Attributes
+//
+extension Storage {
+
+    @objc
+    var typingAttributes: [NSAttributedString.Key: Any] {
+        backingStore.length == .zero ? theme.headlineStyle.attributes : theme.bodyStyle.attributes
+    }
+}
+
+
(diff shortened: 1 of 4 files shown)


eshurakov (approved):
Looks good to me! 👍 

:shipit:

jleandroperez:
This particular one annoyed me since forever. Thanks for checking!!

Issue #762, opened 2020-12-18
Default Line Length


jleandroperez:
### Details:
By default, the line length should be set to **full**

1. User had the line lenght **explicitly set** to full OR narrow  >> stays the same
2. Otherwise the Line Length should be **full width**

PR #775, opened 2020-12-29
Editor: Drops direct frame manipulation

jleandroperez:
### Fix
In this PR we're dropping direct Frame Manipulation in the EditorViewController.

**Note:** 
- The current codebase clips, effectively, 10 points on the right hand side
- New changes drop such clipping
- YES gentlemen, the screenshots look slightly different
- As a user, I would expect the (new) UX rather than the old one

cc @SylvesterWilmott 
cc @eshurakov 

Thanks in advance!!

Ref. #762

### Test
- [ ] Verify the Editor looks great when `View > Line Length > Full` is toggled
- [ ] Verify the Editor also looks great with `View > Line Length > Narrow` enabled

### Release

### Screenshots

Old Narrow|New Narrow
-|-
Old Narrow|New Narrow

Old Full|New Full
-|-
Old Full|New Full

### DIFF
--- a/Simplenote/NoteEditorViewController.m
+++ b/Simplenote/NoteEditorViewController.m
@@ -78,7 +78,6 @@ static NSString * const SPMarkdownPreferencesKey        = @"kMarkdownPreferences
 {
     [super viewDidLoad];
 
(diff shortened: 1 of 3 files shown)


eshurakov (approved):
Looks good to me!

:shipit:

PR #785, opened 2020-12-30
Fixing Unexpected Jumpy Scroll

### LINKED ISSUE #536: NSScrollView Scrolls on Resize
## Details:
This appears to be an ancient and well known **NSScrollView bug**, with [several mentions in StackOverflow.](https://stackoverflow.com/search?q=nsscrollview+resize+scrolling)

Whenever the main NSScrollView is resized, under some scenarios, there'll be an (unwanted) vertical scroll.

## Steps:
1. Insert a new document (sample document at the bottom of this issue!)
2. Shrink the main window to its minimum size
3. Move the cursor all way down in the document
4. Toggle Focus mode back and forth

As a result, the Scroll Offset will jump.

## Investigation:
An (extensive) investigation has led me to the mechanism we've got to constraint the TextView's width [by means of textContainerInset.](https://github.com/Automattic/simplenote-macos/blob/develop/Simplenote/SPTextView.m#L40)

This is triggering an internal mechanism that results in a call to `NSScrollView.scroll(clipView:to)`.

### Potential Fixes Explored:

1. Moving this logic to Autolayout.
  **Problem:** The right scroller ends up in the middle of the screen.

2. Overriding the `NSScrollView.scroll(clipView:to)` API so that it can perform a NO-OP, whenever it's called during a resize cycle. 
  **Problem:** Nothing apparent, but... it's a hack!

3. Updating the textContainer size (to an insane width), and disabling tracking. 
  **Obvious Problem:** The TextView won't adapt to updated width!

### Demo
https://user-images.githubusercontent.com/1195260/103379519-a360af00-4ac4-11eb-80c3-159188534aff.mov

jleandroperez:
### Fix
In this PR we're implementing the mechanism that gets the TextView's "Narrow Width" mode:

- **NSTextContainer.widthTracksTextView** is now disabled. This was the source of a jumpy behavior
- We're manually updating the TextContainer's size
- Insets are now being applied in a `layout` API, rather than in the `draw` override

@eshurakov WOW third PR of the year!! 😂 
Closes #536 

### Test
1. Open a long document
2. Press CMD + T repeatedly, at different scroll offsets

- [ ] Verify the Stroll Offset is always preserved
- [ ] Repeat for `Line Length > Full` and `Narrow` please!!

### Release
`RELEASE-NOTES.txt` was updated in 72a59f1 with:

> Fixed a bug that caused the Editor Scroll Offset to jump #785

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -14,7 +14,8 @@
 - Notes List will now animate content changes #764
 - Fixed a bug that caused an unexpected jump in the Tags List #767
 - Fixed a bug that caused the Tags List to go blank upon tag rename #344
- 
+- Fixed a bug that caused the Editor Scroll Offset to jump #785
+
 2.5
 -----
 - Removed Main Window's Titlebar #719
--- a/Simplenote/NoteEditorViewController+Swift.swift
+++ b/Simplenote/NoteEditorViewController+Swift.swift
@@ -108,6 +108,9 @@ extension NoteEditorViewController {
 
         noteEditor.textContainerInset  = targetContainerInset
         container.containerSize        = targetContainerSize
+
+        /// Note: Disabling `widthTracksTextView` fixes jumpy Scroll Offsets on resize
+        /// Ref. https://github.com/Automattic/simplenote-macos/issues/536
         container.widthTracksTextView  = false
     }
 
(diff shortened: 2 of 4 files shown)


eshurakov (approved):
Looks good to me!

In full line length mode the scroll offset is not always exactly the same, but quite close with and without the sidebar. So I think it's good to go :-) 

:shipit:

jleandroperez:
Thank you @eshurakov !!!

PR #790, opened 2021-01-05
Notes: Sort Date

jleandroperez:
### Details
In this PR we're enhancing the Notes List so that when you're in Search Mode, and the selected Sort Mode is either **Created** / **Modified**, the associated Date will be prepended to the Note Body.

cc @eshurakov 
Thank you sir!!

Ref. #626 

### Test
1. Log into your account
2. Press the top right 🔍 icon
3. Click anywhere in the Sortbar (top of the Notes List)

- [x] Verify that selecting **Creation: Newest / Oldest** causes the Creation date to be prepended to the Note's Body (in the Notes List Area)
- [x] Repeat for **Modified: Newest / Oldest**
- [x] Verify that exiting the Search Mode removes the **Date** prefix

### Release
`RELEASE-NOTES.txt` was updated in 12770fd with:

> Search Results will now display the Creation or Modification date when appropriate #790

### Screenshot

Screenshot

### DIFF
--- a/Simplenote/Note+Simplenote.swift
+++ b/Simplenote/Note+Simplenote.swift
@@ -5,6 +5,21 @@ import Foundation
 //
 extension Note {
 
+    /// Returns the Creation / Modification date for a given SortMode
+    ///
+    func date(for sortMode: SortMode) -> Date? {
+        switch sortMode {
+        case .alphabeticallyAscending, .alphabeticallyDescending:
+            return nil
+
+        case .createdNewest, .createdOldest:
+            return creationDate
+
+        case .modifiedNewest, .modifiedOldest:
+            return modificationDate
+        }
+    }
+
     /// Given a collection of Tag Names, this API will return the subset that's not already associated with the receiver.
     ///
     func filterUnassociatedTagNames(from names: [String]) -> [String] {
--- a/Simplenote/Note+Simplenote.swift
+++ b/Simplenote/Note+Simplenote.swift
@@ -5,21 +5,6 @@ import Foundation
 //
 extension Note {
 
-    /// Returns the Creation / Modification date for a given SortMode
-    ///
-    func date(for sortMode: SortMode) -> Date? {
-        switch sortMode {
-        case .alphabeticallyAscending, .alphabeticallyDescending:
-            return nil
-
-        case .createdNewest, .createdOldest:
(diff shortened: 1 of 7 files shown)


Simplenote/Note+Simplenote.swift
```diff
@@ -5,6 +5,21 @@ import Foundation
...
+        case .alphabeticallyAscending, .alphabeticallyDescending:
+            return nil
+
+        case .createdNewest, .createdOldest:
+            return creationDate
+
+        case .modifiedNewest, .modifiedOldest:
+            return modificationDate
+        }
+    }
```
eshurakov: I don't think that the logic to select a date based on the sort mode belongs in the domain model. 
In my opinion it is something specific to how UI works and should be somewhere in the UI layer.

I have several ideas:
1. Keep it in the Note extension, make the extension private and move it to `NoteListViewController`
2. Move the code to `bodyPrefix` in `NoteListViewController`
3. Create a separate object that would format the date based on sortMode. For example, like this:

```
struct NoteListPrefixFormatter {
    func prefix(from note: Note, for sortMode: SortMode) -> String? {
        let date: Date? = {
            switch sortMode {
            case .alphabeticallyAscending, .alphabeticallyDescending:
                return nil

            case .createdNewest, .createdOldest:
                return note.creationDate

            case .modifiedNewest, .modifiedOldest:
                return note.modificationDate
            }
        }()

        if let date = date {
            return DateFormatter.notesFormatter.string(from: date)
        }

        return nil
    }
}

extension NoteListViewController {
    func bodyPrefix(for note: Note) -> String? {
         guard isSearching else {
             return nil
         }

        return NoteListPrefixFormatter().prefix(from: note, for: listController.sortMode)
     }
}
```
 
Because `NoteListPrefixFormatter` is a struct and doesn't have large instance variable, it can be created on stack and it's basically free to create and destroy it like this.

eshurakov (commented):
Looks good! Sending you some comments.

I also had this strange issue, but don't see what could cause it here. Maybe it is already fixed somewhere? :-)

Screenshot 2021-01-07 at 16 27 11

eshurakov (approved):
Looks great! 

:shipit:

jleandroperez:
Thank you @eshurakov, loooved those changes!!

PR #798, opened 2021-01-13
[Platform] Update Tools and fix nokogiri vulnerability

AliSoftware:
### Fix

This fixes [the following security vulnerability from nokogiri](https://github.com/Automattic/simplenote-macos/security/dependabot/Gemfile.lock/nokogiri/open), by updating to [the latest version of the Release Toolkit](https://github.com/wordpress-mobile/release-toolkit/releases/tag/0.12.0).

It also updates `fastlane` and `cocoapods` to their latest version, as well as `.ruby-version` to `2.6.4`

### Test

* `bundle install` and check gems are consistent and the bundle works.
* Optionally, test some release lanes or ReleaseToolkit actions (`fastlane run …`)

### Review

Only one release manager is required to review these changes, but anyone can perform the review.

### Release


.ruby-version
```diff
@@ -1 +1 @@
-2.5.5
+2.6.4
```
mokagio, line 1: Did you choose 2.6.4 for consistency with other repos / the [release toolkit](https://github.com/wordpress-mobile/release-toolkit/blob/0fc876691762aca8a8f7645f30dd525d5f4b3605/.ruby-version) value?

I was experimenting in another repo and I used the latest Ruby 2.x, [2.7.2](https://www.ruby-lang.org/en/news/2020/10/02/ruby-2-7-2-released/). Everything worked fine. This is totally unrelated with this PR, but I guess it raises the question of how to (or whether to) keep the Ruby version of our project in sync 🤔 

Aside, Ruby 3.0.0 has been released, too 😱

mokagio (approved):

.ruby-version
AliSoftware: Yup that's exactly it, just for consistency with other repos, (and thus to a version of ruby I already had installed on my Mac 😅). I agree it would be nice to update to even newer version on all repos at one point but also to find a way to make it consistent across repos

PR #802, opened 2021-01-20
Update project configuration and pods to build for Apple Silicon

mokagio:
### Fix
Jorge noticed that the app wasn't built for Apple Silicon, a massive fail on my behalf :flushed:

This PR updates a bunch of things, all unfortunately required to make the build happen.

- CocoaPods to version 1.10.1 because that has proper support for `.xcframework`s, required by...
- Sodium to version 0.9.1, which has support for Apple Silicon and is included in
- Tracks, which also includes
- Sentry (via @jkmassel's PR https://github.com/Automattic/Automattic-Tracks-iOS/pull/166/) because the current version has a macro to exclude iOS specific code when not building for arm64, which no longer works because we're now building for arm64 on macOS, too

Finally, the `CrashLogging` implementation had to be tweaked, too, in order to compile with the Sentry-related changes in Tracks.

### Merge Thoughts

~~I based this PR off `release/2.6` hoping that it can make it in there, but **it would be appropriate to wait till an official version of Tracks has been released before merging this PR.**~~

~~That might mean delaying shipping version 2.6 to the users for a few days, or longer in case there are issues with the Sentry migration.~~

~~If the work on Tracks takes longer, we can shift the Apple Silicon support to version 2.7 and have a more thorough testing of it.~~

This is now based off `release/2.7`.

### Test

Checkout this code, run `bundle install && bundle exec pod install`, build then archive the Simplenote-AppStore scheme. Run the app and make sure it works. ~~Do the same for the Simplenote scheme.~~ _See #803 for that_.

Screen Shot 2021-01-20 at 3 07 57 pm

### Review
I'd like a review from a MIE and from someone with an M1 Mac to verify it builds fine there.

### Release
`RELEASE-NOTES.txt` was updated in 36f3b2b with:

> Add Apple Silicon build support


Gemfile
```diff
@@ -1,7 +1,6 @@
 source 'https://rubygems.org' do
   gem 'rake'
-  gem 'cocoapods', '~> 1.8'
-  gem 'cocoapods-repo-update', '~> 0.0.3'
+  gem 'cocoapods', '>= 1.10'
```
mokagio, line 3: CocoaPods 1.10.1 is required to properly build the `xcframework`s that Sodium (a Tracks dependency) uses to support Apple Silicon.

The repo update plugin blocked the dependency resolution and it should anyways no longer necessary because CocoaPods uses the CDN by default.

Podfile
```diff
@@ -6,16 +6,13 @@ use_frameworks!
...
-
-
 # Main
 #
 abstract_target 'Automattic' do
 
   # Automattic Shared
   #
-  pod 'Automattic-Tracks-iOS', '~> 0.6'
+  pod 'Automattic-Tracks-iOS', git: 'git@github.com:Automattic/Automattic-Tracks-iOS', branch: 'update/sentry-to-v6'
```
mokagio: **It would be appropriate to wait till an official version of Tracks has been released before merging this PR.**

That might mean delaying shipping version 2.6 to the users for a few days, or longer in case there are issues with the Sentry migration. 

Alternatively, we can shift the Apple Silicon support to version 2.7 and have a more thorough testing of it.

Simplenote.xcodeproj/project.pbxproj
```diff
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 52;
+	objectVersion = 54;
```
mokagio, line 6: This and the `Xcode 12.0` change below are so that CocoaPods can keep using a file to provide the frameworks list as part of its integration. Otherwise, it will list them directly in the project, making for noisier diffs.

Simplenote.xcodeproj/project.pbxproj
```diff
@@ -1762,7 +1762,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "plist=\"${TARGET_BUILD_DIR}/${INFOPLIST_PATH}\"\nyear=$(date \"+%Y\")\ncopyright=\"Copyright © $year Automattic. All rights reserved.\"\n\n# Set the version numbers in the buildPlist\n/usr/libexec/PlistBuddy -c \"Set :NSHumanReadableCopyright $copyright\" $plist\n";
+			shellScript = "plist=\"${TARGET_BUILD_DIR}/${INFOPLIST_PATH}\"\nyear=$(date \"+%Y\")\ncopyright=\"Copyright © $year Automattic. All rights reserved.\"\n\n# Set the version numbers in the buildPlist\n#\n# Temporarily disabled! It's failing the build after updating to support Apple Silicon.\n#\n# Unfortunately, a number of changes were required to make the build pass before getting to this point,\n# so there is no clear responsible for this failure yet.\n# /usr/libexec/PlistBuddy -c \"Set :NSHumanReadableCopyright $copyright\" $plist\n";
```
mokagio: You'll have to scroll horizontally a bit to see the change. Here's a formatted version:

```diff
  plist="${TARGET_BUILD_DIR}/${INFOPLIST_PATH}"
  year=$(date "+%Y")
  copyright="Copyright © $year Automattic. All rights reserved."

  # Set the version numbers in the buildPlist
- # /usr/libexec/PlistBuddy -c "Set :NSHumanReadableCopyright $copyright" $plist
+ #
+ # Temporarily disabled! It's failing the build after updating to support Apple Silicon.
+ #
+ # Unfortunately, a number of changes were required to make the build pass before getting to this point,
+ # so there is no clear responsible for this failure yet.
+ # /usr/libexec/PlistBuddy -c "Set :NSHumanReadableCopyright $copyright" $plist
```

I'd be curious to see if this happens for you, too.

Simplenote.xcodeproj/project.pbxproj
```diff
@@ -2409,7 +2409,6 @@
 				SDKROOT = macosx;
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
 				SWIFT_VERSION = 5.0;
-				VALID_ARCHS = x86_64;
```
mokagio, line 2556: This appears to be a legacy build setting: if you create a new macOS with Xcode 12.0, it won't be there.

I opted for removing it instead of updating it to include `arm64`.

Simplenote/TagListState.swift
```diff
@@ -108,7 +108,7 @@ extension TagListState {
     ///
     @objc
     func tag(atIndex index: Int) -> Tag? {
-        guard case let .tag(tag) = rowAtIndex(index) else {
+        guard let row = rowAtIndex(index), case let .tag(tag) = row else {
```
mokagio, line 111: The compiler required `rowAtIndex(_:)` `Optional` return value to be unwrapped, although I don't understand why this failure appeared now, since it's not related with Tracks 🤔

Podfile
jleandroperez: Good on my end! We can switch when it's available!

Simplenote.xcodeproj/project.pbxproj
jleandroperez: Apparently it defaults to `Standard Archs` so it's all good! thank you!!

Simplenote/TagListState.swift
jleandroperez: Absolutely a mystery to me as well, this change is unrelated to anything else. I suspect it's perhaps a glitch in the compiler (since one of the changes is the supported archs right?).

No big deal, it's a valid change!! thank you Gio!!

jkmassel:
I'm happy to review this since most of it is the Tracks update, but if it's ok with y'all I'll wait until the Tracks PR lands – at that point I'd love to test this a bunch to ensure we're good to go!

jleandroperez:
@mokagio code changes look good!! But I'm having issues archiving the `AppStore` target on my end. Will poke you later today!!

```
/Users/lantean/Developer/Automattic/simplenote-macos/Simplenote/Note.h:9:9: error: module 'Simperium_OSX' not found
@import Simperium_OSX;
        ^
```

mokagio:
@jleandroperez and I worked on this together earlier.

Archiving worked on my machine but not on his. We noticed he's on Big Sur and I'm on still on Catalina so we thought the issue was due to that. I tried on my other Mac with Big Sur and... bang! I got the same issue: building works but archiving doesn't 🤔 

Only problem was... I wasn't sure how to fix it.

After some poking around I noticed that I was archiving for "My Mac". Once I switched to "Any Mac (Apple Silicon, Intel)" the archive worked 🎉 🤦 

I went back to my main Catalina machine and – _you guess it_ – the destination was Any Mac. If I switch it to My Mac, the archive fails.

Screen Shot 2021-01-21 at 2 15 34 pm


Here's an excerpt from the [Xcode documentation](https://developer.apple.com/documentation/xcode/distributing_your_app_for_beta_testing_and_releases#3405659):

> In the main window of your Xcode project, choose a target and a build-only device or real device from the Scheme toolbar menu. If the target is a simulator, you can’t create an archive. Then, choose Product > Archive to build the target, and create the archive that will appear in the Archives organizer.

It's super easy thing to overlook: on iOS, Xcode disables the Archive menu item when the destination is a Simulator, but that's not the case for Mac (I'm guessing because "My Mac" counts as a real device?)

I'm still unsure why the error is that the Simperium module can't be found with destination "My Mac" 🤔 But, unless archiving for "Any Mac" still fails for you @jleandroperez, I think we can consider that issue closed.

mokagio:
> I'm happy to review this since most of it is the Tracks update, but if it's ok with y'all I'll wait until the Tracks PR lands – at that point I'd love to test this a bunch to ensure we're good to go!

Sounds good to me 👍

jleandroperez (approved):
Thank you @mokagio, that was it, `Any Mac`!!!

Good to go sir!!

:shipit: :shipit: :shipit:

mokagio:
Moved this to 2.7 after discussing it with @jleandroperez. Since the Tracks work is still in flux, it would be unsafe to ship it to the users. M1 owners can still enjoy Simplenote because of Rosetta.

Podfile
jkmassel: I've switched this over

.ruby-version
```diff
@@ -1 +1 @@
-2.5.5
+2.6.4
```
jkmassel: Do we need to use a newer Ruby version for this repo?

Gemfile.lock
```diff
@@ -297,4 +300,4 @@ DEPENDENCIES
   xcpretty-travis-formatter!
 
 BUNDLED WITH
-   2.1.4
+   2.2.5
```
jkmassel, line 304: Did we mean to change this?

Simplenote/CrashLogging.swift
```diff
@@ -7,19 +7,22 @@ import AutomatticTracks
 /// using `SNCrashLoggingDataProvider`.
 @objc(CrashLogging)
 class CrashLoggingShim: NSObject {
+
+    private static var sharedCrashLogging: CrashLogging?
```
jkmassel: Instead of the singleton here, could we instead move this to the `AppDelegate` and explicitly initialize it there?

All of the calls into this shim happen from the AppDelegate anyway, so there they could reference the instance, and we could get rid of all the `static` business going on here?

Simplenote.xcodeproj/project.pbxproj
jkmassel: This doesn't happen to me locally [or in CI](https://app.circleci.com/pipelines/github/Automattic/simplenote-macos/1830/workflows/9ab7856b-4399-4f2e-8b2a-2605425b52f4/jobs/1784), so AFAICT it's safe to put back, but LMK if not?

jkmassel (changes requested):
I updated this PR to use Tracks 0.8.0 prior to testing – the changes are all in 6f95235cf776b9ca2f8ea8ddf37a97ee2550e624. While I was there, I also dropped the old Pods workaround.

A few things that caught my eye:
- Some (possibly?) unintended changes to the Ruby setup
- Ideally we wouldn't introduce another singleton object in the CrashLogging shim 😅

.ruby-version
mokagio: No we don't. I updated this to keep it in line with the rest of our projects (b79d8d2587810cf59124da7d65eb6863d124cdf8).

Do you think it's a problem? An out-of-scope change on a PR that should be as little as possible given it targets the release branch?

Gemfile.lock
mokagio: I did it intentionally, yes. The reason being that Bundler 2.2.x fixes an issue that may result in confusing changes to the `.bundle/config` file (more details in this Simplenote iOS PR https://github.com/Automattic/simplenote-ios/pull/1059).

Do you think it's a problem? An out-of-scope change on a PR that should be as little as possible given it targets the release branch?

Simplenote/CrashLogging.swift
mokagio: Great suggestion, thank you for the extra context. I'll admit I went with this option simply because it allowed me to keep the change in this file without touching more of the app.

I'll give your suggestion a try, as I'm not happy with singletons either.

Simplenote/CrashLogging.swift
```diff
@@ -8,21 +8,32 @@ import AutomatticTracks
 @objc(CrashLogging)
 class CrashLoggingShim: NSObject {
 
-    private static var sharedCrashLogging: CrashLogging?
+    private let crashLogging: CrashLogging
```
mokagio: @jkmassel from your suggestion, I got the idea of fully removing this class and calling its methods directly in the application delegate.

That's not possible because `CrashLogging` is a Swift type that's not visible to the Objective-C `AppDelegate`.

This is an attempt to remove the `static` component while staying within the Objective-C constraints. I created a private instance of the Swift `CrashLogging` that this shim calls via its instance methods visible to the Objective-C code.

.ruby-version
AliSoftware: I personally would keep that change, as I agree it's nice to use the same ruby version across all our tooling.
I did a similar update in the past on another PR and repo (can't remember which one, but we discussed it in PR comments with Gio at that time).

The benefits of using the same ruby across our repos whenever possible is to avoid having to `rbenv install` and `bundle install` all the gems again (duplicating all the downloads and size on disk for the gems for usually no main benefit), while I don't really see any major drawback.

Simplenote/CrashLogging.swift
AliSoftware: Wondering if we couldn't make the main `CrashLogging` Swift type be `@objc`-compatible in the first place?

mokagio: Thank you for the suggestion. 🙇 

`CrashLogging` comes from Tracks, so updating it would require another round trip. That's not a huge issue, just a possible time sink that might prevent us from landing this in `release/2.7`.

I'd be reluctant to go down that path, though, because I fear a rabbit hole. @jkmassel you have  more experience with that code, so maybe you can  tell me if I'm being overly cautious?

Maybe we should update the app delegate to be Swift, instead 😛 I'm only half joking. Obviously that would be a bigger piece of work, but it might a good time investment to modernize the codebase?

.ruby-version
jkmassel: kk works for me if we're making things _more_ consistent!

Gemfile.lock
jkmassel: I'm good with it! Just wanted to be sure it was intentional!

Simplenote/CrashLogging.swift
jkmassel: Thanks for looking at this @mokagio. @AliSoftware – we could certainly do that, though Simplenote is the only Objective-C (iOS and macOS) App Delegate (and ideally IMHO wouldn't own the Tracks `CrashLogging` class directly). 

FWIW it wouldn't be that hard to do, just need to conform to `NSObject`.

Simplenote/SimplenoteAppDelegate.m
```diff
@@ -163,7 +165,7 @@ - (void)handleGetURLEvent:(NSAppleEventDescriptor *)event withReplyEvent:(NSAppl
 
 - (void)setupCrashLogging
 {
-    [CrashLogging startWithSimperium: self.simperium];
+    self.crashLogging = [[CrashLogging alloc] initWithSimperium:self.simperium];
```
jkmassel, line 168:
```suggestion
    self.crashLogging = [[CrashLogging alloc] initWithSimperium:self.simperium];
    [self.crashLogging start];
```

We need to call `start` on the `CrashLogging` object in order to initialize the SDK internally.

jkmassel (approved):
Thanks for addressing the feedback!

I did an end-to-end test with Sentry, which didn't work for me (which lines up with what you'd mentioned!). We'll need to address that (I've left a suggestion), and to test, I'd recommend also adding `-force-crash-logging 1` to both of the app schemes (though switched off by default) to make it possible to test from debug builds.

Then I think this is ready!!

jkmassel (changes requested):
Accidentally approved instead of requesting changes 🤦

Simplenote/SimplenoteAppDelegate.m
mokagio: 🤦‍♂️

AliSoftware: If it's a common pattern it could have been nice to add a `static func start(with simperium: Simperium)` function in the lib that would be just a wrapper around `init(simperium:)` + `start()` just for nicer API… bwah, too late and not worth it now ^^

Simplenote/SimplenoteAppDelegate.m
```diff
@@ -47,6 +47,8 @@ @interface SimplenoteAppDelegate () <SPBucketDelegate>
 @property (strong, nonatomic) SPUStandardUpdaterController      *updaterController;
 #endif
 
+@property (strong, nonatomic) CrashLogging                      *crashLogging;
```
AliSoftware, line 50: spaaaaace (Portal 2 meme)

Simplenote/TagListState.swift
AliSoftware: I think that's a change in a recent Swift compiler update?

Some other possibilities to fix it would have been to use `?` to match with the proper optional lhs type:
```swift
        // '?' shorthand syntax for matching `Optional.some`
        guard case let .tag(tag)? = rowAtIndex(index) else {
        // with the `let` as close as possible to the variable you want to bind (which I personally prefer, but 🤷 )
        guard case .tag(let tag)? = rowAtIndex(index) else {
```

Which are both valid shorthands for the following more expanded forms of `Optional.some` (since that's an non-nil optional we're trying to match with here) if that helps see it better:
```swift
        // Fully expanded form with fully qualified names
        guard case Optional.some(TagListRow.tag(let tag)) = rowAtIndex(index)
        // Explicit Optional case
        guard case .some(.tag(let tag)) = rowAtIndex(index)
```

I think in previous versions of Swift the compiler would allow automatically lifting the operands to optionals if needed but that got removed to avoid ambiguity found in some cases.

AliSoftware (approved):
Was not able to test on an actual M1, so could only test on Intel and by reviewing code in GitHub UI, but LGTM 👍

Simplenote/SimplenoteAppDelegate.m
mokagio: 😆

mokagio:
Adding the `start` call solved the issue.

Here's the demo crashes I sent

Screen Shot 2021-01-29 at 6 04 52 am

_The bottom one, with the blue dot next to it, is not mine, it's a test I guess @jkmassel sent. It actually confused me for a while because I though it was how Sentry reported its force-crash; I kept reloading its detail page and wondering why mine weren't appearing._

Here's the detail from one
Screen Shot 2021-01-29 at 6 05 28 am

Notice how it was triggered by `aboutAction:` and how the release is `42` (bottom right)

mokagio:
@jkmassel asking a new review from you, but reverting this to Draft to make sure I merge it only _after_ #803 is sorted out.

jkmassel (approved):

jleandroperez:
Awesome work gentleman!!! 🎉

PR #639, opened 2020-08-10
Update Sparkle SUFeedURL from Hockey to App Center URL

mokagio:
### Fix
The previous URL redirects to this one. This change will save us a
redirect when checking for updates, and also bring the app up-to-date
with the App Center standards.

### Test

To test this, I changed the version to something old, like 1.0, and also enabled Sparkle for the "Simplenote" target in the Debug build configuration.

```diff
diff --git a/Simplenote.xcodeproj/project.pbxproj b/Simplenote.xcodeproj/project.pbxproj
index 4404404..b7f1530 100644
--- a/Simplenote.xcodeproj/project.pbxproj
+++ b/Simplenote.xcodeproj/project.pbxproj
@@ -2179,6 +2179,7 @@
 				GCC_PREPROCESSOR_DEFINITIONS = (
 					"$(inherited)",
 					"COCOAPODS=1",
+					"SPARKLE_OTA=1",
 				);
 				GCC_WARN_UNDECLARED_SELECTOR = NO;
 				INFOPLIST_FILE = "Simplenote/Simplenote-Info-Hockey.plist";
diff --git a/config/Version.public.xcconfig b/config/Version.public.xcconfig
index 212df1d..179a37f 100644
--- a/config/Version.public.xcconfig
+++ b/config/Version.public.xcconfig
@@ -1,5 +1,5 @@
-VERSION_SHORT=2.0
+VERSION_SHORT=1.0

 // Public long version example: 2.0.0.1. The last 1 means the first build for
 // this version
-VERSION_LONG=2.0.0.0
+VERSION_LONG=1.0.0.0
```

Something that surprised me, though, was how the version it proposed me was 1.12.0 instead of 2.0.0.

I wonder if that is due to the different version format in the [appcast XML](https://api.appcenter.ms/v0.1/public/sparkle/apps/ede9c493-4694-6c51-a1db-1db167b49892):

```xml
<enclosure sparkle:version="2.0.0.0" sparkle:shortVersionString="2.0" url="https://appcenter-filemanagement-distrib2ede6f06e.azureedge.net/0e3db9c2-3387-4efb-bc76-a07fe805c776/Simplenote.zip?sv=2019-02-02&sr=c&sig=jC%2FRjmryhD%2BziK%2F548Nnb3RO%2BtjlNj5wc6halNNaW2Y%3D&se=2020-08-11T03%3A48%3A34Z&sp=r" length="6632120" type="application/octet-stream"/>
```

vs

```xml
[...]

PR #803, opened 2021-01-20
Update Sparkle to version 1.24.0 to support Apple Silicon

mokagio:
**Update: Superseded by #824** 

---

This builds on top of #802 and only upgrades Sparkle to allow the internal Simplenote target to build for Apple Silicon.

I moved it in an isolated PR because of the huge diff. You'll notice there is not change other than the Sparkle files, so if the Simplenote target builds, runs, and archives as normal for you, that should be enough.


AliSoftware:
I'll wait for https://github.com/Automattic/Automattic-Tracks-iOS/pull/166 then #802 before considering testing this one, but in the meantime, quick question: any reason why we include Sparkle manually and commit it into git (`External/*`), as opposed to include it via CocoaPods or something?

I don't know if Sparkle provides a podspec (even if it doesn't we might consider creating a local `.podspec` file if needed), but if they were to at least distribute it in a way that would avoid us having to commit it, it would (1) reduce the repo clone size (2) make diffs easier (3) reduce noice in PRs like this one. 🤷

mokagio:
> any reason why we include Sparkle manually and commit it into git (`External/*`), as opposed to include it via CocoaPods or something?

The setup pre-dates me. I'm not sure if there was a specific reason or if maybe CocoaPods wasn't supported at the time.

> I don't know if Sparkle provides a podspec (even if it doesn't we might consider creating a local .podspec file if needed), but if they were to at least distribute it in a way that would avoid us having to commit [...]

Totally. I didn't go down that route to keep this PR focused on making the build work vs. changing integration method. While that might not have interfered with anything, I feel like I already changed enough infrastructure already 😅

mokagio:
Moved to 2.7 as per [this comment in the base PR](https://github.com/Automattic/simplenote-macos/pull/802#issuecomment-765034695).

jkmassel (changes requested):
This was actually a breaking change, but it wasn't detected as such because we only built the Sparkle libraries for release builds previously, so CI didn't catch it. I've added two commits: 

- One to cause the build to fail due to the change
- One to fix Sparkle

Because it is a breaking change, IMHO this is a good opportunity to adopt Sparkle as a CocoaPod (since we'll have to test it extensively anyway).

Interested to hear your thoughts since you've already looked at it @AliSoftware.

Simplenote.xcodeproj/project.pbxproj
```diff
@@ -1855,19 +1809,6 @@
...
-		};
```
mokagio, line 1871: The CocoaPods should make this unnecessary. See https://github.com/sparkle-project/Sparkle/issues/709.

Simplenote/Simplenote-Info-Hockey.plist
```diff
@@ -53,7 +53,7 @@
 	<key>NSPrincipalClass</key>
 	<string>SPApplication</string>
 	<key>SUFeedURL</key>
-	<string>https://rink.hockeyapp.net/api/2/apps/ede9c49346946c51a1db1db167b49892</string>
+	<string>https://api.appcenter.ms/v0.1/public/sparkle/apps/ede9c493-4694-6c51-a1db-1db167b49892</string>
```
mokagio, line 56: See #639. Without this, the update check would fail with 

```
[Sparkle] Error: An error occurred in retrieving update information. Please try again later. An error occurred while parsing the update feed. (URL (null))
[Sparkle] Error: An error occurred while parsing the update feed. (null) (URL (null))
[Sparkle] Error: Line 6: Opening and ending tag mismatch: hr line 5 and body
 (null) (URL (null))
```

mokagio:
🤔 I'm getting the update dialog, the new version is downloaded correctly, but then I end up stuck on the "Install and Relaunch" dialog

Screen Shot 2021-01-27 at 4 25 29 pm

Simplenote.xcodeproj/project.pbxproj
jkmassel: Yep, that's a huge win for sure!

jkmassel:
> I end up stuck on the "Install and Relaunch" dialog

I wonder if it has anything to do with code signing? Something should show up in `Console.app` if gatekeeper is blocking a process (I think...)

mokagio:
@jkmassel you might be onto something there.

Yesterday, I looked into `Console.app` but couldn't find anything. I found that suspicious, so today I did some reading to learn how to use the tool properly and was able to find the logs Simplenote macOS emits.

image


Among those, I found these:

```
Error registering stapled ticket: Error Domain=NSOSStatusErrorDomain Code=-67048 "(null)"
Warning: could not remove quarantine metadata from /Users/gio/Library/Containers/com.automattic.SimplenoteMac.beta/Data/Library/Caches/com.automattic.SimplenoteMac.beta/org.sparkle-project.Sparkle/Simplenote (Autoupdate).app: Error Domain=NSPOSIXErrorDomain Code=1 "Operation not permitted" UserInfo={NSLocalizedDescription=Failed to remove file quarantine on Simplenote (Autoupdate).app.}. This may not be a problem, the update process will continue.
```

I might run out of time to follow this up today, but it's a promising lead to get to the bottom of the issue.

mokagio:
Closing in favor of #824.

PR #481, opened 2020-04-06
Updating Sparkle

jleandroperez:
### Description
In this PR we're updating Sparkle to the [(experimental) branch 2.x](https://github.com/sparkle-project/Sparkle/tree/2.x).

Plus we're integrating Sparkle's codesign tooling, rather than calling `codesign` ourselves, which is expected to solve our notarization issues.

### Note
**No, it's not possible to integrate Sparkle 2.x via cocoapods**

### Details: Building Sparkle
For the record, in order to build Sparkle you'd need to:

1. Check out [branch 2.x](https://github.com/sparkle-project/Sparkle/tree/2.x)
2. Switch your local ruby setup to 2.6 (and make sure you've got cocoapods installed)
3. Run `make release` in Sparkle's folder root

### Test
- [ ] Verify the app builds correctly
- [ ] Verify the Autoupdater works as expected

### Release
These changes do not require release notes.

PR #383, opened 2019-08-30
Fixing OTA

jleandroperez:
### Fix
In this PR we're implementing a beta version of Sparkle, which has sandboxing support.

Since this version isn't distributed over Cocoapods, i'm afraid we were forced to just add the files.

[Ref. here.](https://github.com/sparkle-project/Sparkle/issues/1182)

### Test
1. Ship a build of this branch over HockeyApp
2. Download such build in your mac!
3. Bump up the version number (for testing purposes)
4. Ship yet another build over HockeyApp

- [ ] Verify the OTA mechanism works well

PR #824, opened 2021-01-29
Manually update Sparkle to support Apple Silicon

mokagio:
In #803, we discussed using CocoaPods to fetch Sparkle. That's not possible because we need support for Sandboxing and Notarization and the version distributed via CocoaPods doesn't include it.

To support Sandboxing and Notarization, we need to use the WIP version 2.x branch. Unfortunately, the only way to use that version of Sparkle is to fetch it and build it manually (see `./Scripts/update-sparkle.sh`).

We cannot do like we've done many times with other libraries and point the `Podfile` to the the `2.x` branch because there the Pods [depends on a vendored framework](https://github.com/sparkle-project/Sparkle/blob/4767ef60ece7de7404f402c23177f0810bf72418/Sparkle.podspec#L27) which needs to [be built](https://github.com/sparkle-project/Sparkle/blob/4767ef60ece7de7404f402c23177f0810bf72418/Makefile#L16) every time. At least, that's my current understanding. I did try to just switch the branch and CocoaPods didn't generate any framework or fetch any source file

image

For future reference, the commit I checked out was `4767ef60ece7de7404f402c23177f0810bf72418` (this information is in the update commit, too).

I considered using a submodule and scripting the build process, but decided against it because it would add complexity to the Git setup for a relatively uncommon operation – we don't need to update Sparkle often.

Having said that, I still wrote a script to perform the update, because it's not as straightforward as a drag-and-drop.

For more background on Sparkle Sandboxing and Notarization, see:

- https://sparkle-project.org/documentation/sandboxing/
- https://github.com/Automattic/simplenote-macos/pull/383
- https://github.com/Automattic/simplenote-macos/pull/481

### Test

I tested this by:

1. Change the version (`config/Version.public.xcconfig`) to be `2.6 (42)`
2. Archive this branch
3. Submit it for notarization (Distribute > Developer ID > Upload)
4. Once notarized (leave the window open and check after a few minutes), export the `.app` and copied it to the `/Applications` folder
5. Launch Simplenote

The Sparkle update should appear a few moments after and it should be possible to update to version 2.7

### Review

Keen to hear what @jkmassel has to say.
Also @jleandroperez because you worked on using Sparkle 2.x, so let me know if my setup looks coherent with the original.

### Release


Scripts/Build-Phases/sparkle-code-sign.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
```
mokagio, line 1: This was inlined in the project file. I merely extracted it to a dedicated file for clarity.

Scripts/update-sparkle.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
```
mokagio, line 1: I'll admit, this script took me longer than just manually copying those files. It might be wasted time because the Sparkle folks look close to releasing 2.x. Still, it wasn't at all clear to me that we were using version 2.x, so I think this might save us time in case we'll have to update Sparkle manually again.

Scripts/update-sparkle.sh
```diff
@@ -0,0 +1,62 @@
...
+
+#
+# Builds a new version of the Sparkle framework and copies the necessary files
+#
```
mokagio, line 5: I'll admit, this script took me longer than just manually copying those files. It might be wasted time because the Sparkle folks look close to releasing 2.x. Still, it wasn't at all clear to me that we were using version 2.x, so I think this might save us time in case we'll have to update Sparkle manually again.

Scripts/Build-Phases/sparkle-code-sign.sh
jkmassel: This is a great improvement – will definitely help clarify the diff! ✨

Simplenote.xcodeproj/project.pbxproj
```diff
@@ -1541,7 +1541,7 @@
 				26F72A8514032D2A00A7935E /* Frameworks */,
 				26F72A8614032D2A00A7935E /* Resources */,
 				71881504155F487D001F2E30 /* CopyFiles */,
-				B51D44911945FEFD00496889 /* ShellScript */,
+				B51D44911945FEFD00496889 /* Sparkle XPC Code Sign For Sandbox */,
```
jkmassel, line 1544: Yaay for more descriptive build phase names!

Scripts/update-sparkle.sh
jkmassel: > it wasn't at all clear to me that we were using version 2.x,

Yep, same – this will definitely be an improvement there!

readme.md
```diff
@@ -21,6 +21,16 @@ We use a few tools to help with development. To install or update the required d
 Simplenote for iOS uses [CocoaPods](http://cocoapods.org/) to manage third party libraries.
 Third party libraries and resources managed by CocoaPods will be installed by the `rake dependencies` command above.
 
+#### Why isn't Sparkle fetched with CocoaPods
```
jkmassel, line 24: This is great – thanks for adding it!!

Scripts/Build-Phases/sparkle-code-sign.sh
```diff
@@ -0,0 +1,58 @@
...
+if [ "${CONFIGURATION}" != "Release" ]; then
+    echo "NOT Signing in debug!"
+    exit 0
+fi
+
+# Reset: Sparkle's Codesign fails when attempting to re-sign
+#
+echo "Nuking old XPC..."
+
+rm -r "${XPC_TARGET}"
```
jkmassel:
```suggestion
rm -rf "${XPC_TARGET}"
```

This will fail on a clean build directory because the directory doesn't exist. Using `-rf` fixes that.

jkmassel (changes requested):
I left a few notes, but this also didn't work for me – I followed the steps and was able to see that an update was available, but couldn't install it – the new version downloaded but the install didn't start, the app didn't close, and the version in my applications directory had the same version afterward.

I tried downloading from AppCenter, but that version didn't prompt for update at all (I cleared out the app container to ensure it wasn't applying the sparkle rate limiter)

Might be worth roping someone else in if this is all working on your machine?

Scripts/Build-Phases/sparkle-code-sign.sh
mokagio: Can you believe that I actually had a commit that used `-rf` then I removed it because I thought the failure I experienced when I added it was due to version `1.x` not having the XPC stuff?

🤦

mokagio:
@jkmassel 

> this also didn't work for me – I followed the steps and was able to see that an update was available, but couldn't install it – the new version downloaded but the install didn't start, the app didn't close, and the version in my applications directory had the same version afterward.

🤔 maybe it's got something to do with local code signing?

[Here's a ZIP](https://github.com/Automattic/simplenote-macos/files/5895495/Simplenote-Sparkle-Update-Demo.app.zip) with the build I made for which the update works on my end.

> I tried downloading from AppCenter, but that version didn't prompt for update at all (I cleared out the app container to ensure it wasn't applying the sparkle rate limiter)

~~That happened to me, too. I have the suspicion something broke with Sparkle recently and we only realized it now.~~

Oh, I figured out why! The Hockey app version of the URL 503s:

https://rink.hockeyapp.net/api/2/apps/ede9c49346946c51a1db1db167b49892

jkmassel (approved):
> Oh, I figured out why! The Hockey app version of the URL 503s:

Interesting – I wonder if that's permanent 🤔 

I did a build that uses [the newest recommended format](https://docs.microsoft.com/en-us/appcenter/distribution/sparkleupdates) (same as the one in `Simplenote-Info-Hockey.plist`) and it worked great – it silently updated me to 2.7 when I restarted the app.

I think this is ready to go! (after we copy the change over to the other `.plist` file, just in case) :shipit:

mokagio:
> after we copy the change over to the other .plist file, just in case

@jkmassel oh, interesting... I didn't even think to check to see if the scheme for the App Store. 

What do you think of _removing_ the Sparkle keys from that one instead? After all, the App Store build should not use Sparkle, the `SPARKLE_OTA` value is not defined so [it won't even import the framework in the `AppDelegate`](https://github.com/Automattic/simplenote-macos/blob/develop/Simplenote/SimplenoteAppDelegate.m#L25-L27).

jkmassel:
>What do you think of removing the Sparkle keys from that one instead? After all, the App Store build should not use Sparkle, the SPARKLE_OTA value is not defined so it won't even import the framework in the AppDelegate.

Yeah, we can do that – just not sure why it would be hitting https://rink.hockeyapp.net/api/2/apps/ede9c49346946c51a1db1db167b49892 at all, but based on the targets (I was running the non-App Store version) it looks like removing it should be just fine!

mokagio:
Okay. Thanks. 

I'll do that then merge.

> I was running the non-App Store version

🤔 oh, that's odd that the non-App Store one would pick that one from the App Store one `Info.plist`.

PR #830, opened 2021-02-01
Show relevant excerpts while searching

eshurakov:
### Details
In this PR we're updating note preview to show relevant excerpt while searching.

Excerpt is generated from the text around the first encountered keyword. Text before and after the keyword is trimmed and only full words are used.

Ref: #626 

This PR is mostly the same as https://github.com/Automattic/simplenote-ios/pull/965

### Screenshots

Screenshot 2021-02-01 at 11 59 35
Screenshot 2021-02-01 at 11 59 29

### Test

- [x] You can see the word you're searching for in the excerpt
    - [x] Even if it's the first word in the body
    - [x] Even if it's the last word in the body
    - [x] Even if it's somewhere in the middle of the body
- [x] Searching for words in the beginning of the text doesn't prepend "..." to the excerpt
- [x] Searching for words in the middle of the text prepends "..." to the excerpt
- [x] Diactric insensitive search. For example, if you search for Sao, it finds and highlights São.

### Review
Only one developer and one designer are required to review these changes, but anyone can perform the review.

### Release
`RELEASE-NOTES.txt` was updated in a8857f7 with:

> Searching for a note is showing relevant excerpts


SylvesterWilmott (approved):
Design looks good.

Little design note: we changed the colour of the selected state in the notes list so that the keyword highlighting was able to meet accessibility standards.

Simplenote/Note.m
```diff
@@ -161,24 +161,6 @@ - (void)ensurePreviewStringsAreAvailable
...
 }
 
-- (void)createPreview
-{
-    NSString *aString = self.content;
-    if (aString.length > 500) {
-        aString = [aString substringToIndex:500];
-    }
-
-    [aString generatePreviewStrings:^(NSString *title, NSString *body) {
```
jleandroperez, line 171: Before we forget!! Mind deleting these? (since they're no longer in use):

- NSString+Condensing
- NSStringCondensingTests

Simplenote/NoteTableCellView.swift
```diff
@@ -249,14 +266,30 @@ extension NSAttributedString {
     /// Returns a NSAttributedString representation of a given String, with the specified parameters.
     /// List Markers will be replaced by Text Attachments
     ///
-    static func previewString(text: String, font: NSFont, color: NSColor) -> NSAttributedString {
+    static func previewString(text: String,
+                              font: NSFont,
+                              color: NSColor,
+                              highlighting highlight: (keywords: [String]?, color: NSColor, font: NSFont)? = nil) -> NSAttributedString {
```
jleandroperez: I'm wondering if it's worth encapsulating these 3 properties into a struct (`struct KeywordHighlight`?), or dropping the tuple?

(Just to improve a bit readability!)

jleandroperez (approved):
@eshurakov this PR is pure gold, exactly as advertised (and similar to the iOS sibling!!).

Thank you!!!

Simplenote/Note.m
eshurakov: Good point, totally missed it.
We were using `replacingNewlinesWithSpaces` from that file, so I moved it to `String+Simplenote` extension.

Simplenote/NoteTableCellView.swift
eshurakov: At first I did it without a tuple, but we're not using highlighting in all cases, so `color` and `font` for highlighting would need to be optional. To avoid this I wrapped it in a tuple and made the tuple optional.
I think it looks pretty much like struct. Why do you think it's not very readable, because of the line length?

jleandroperez: Yeah it's just the line length. BUT it's a cosmetic detail, so TBH no need to change anything, I'd say.

A typealias maybe is what we're thinking of?

eshurakov: Great thinking! Added a typealias :-)

eshurakov:
Thank you!

pachlava:

PR #831, opened 2021-02-01
SortBar: Always visible

jleandroperez:
### Fix
In this PR we're adjusting the Search Bar, so that it's always visible.

cc @eshurakov (Thank youuu!!)
Ref. #626

### Test
- [x] Verify the Sort Bar is visible at launch (at the top of the Notes List)
- [x] Verify you can adjust the Sort Order (via the Sort Bar) at any time
- [x] Verify the Notes List's content offset is perfect at launch (nothing should get clipped!)

### Release
`RELEASE-NOTES.txt` was updated in 7c0f435 with:

> SortBar is now always visible #831

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,7 +1,10 @@
+2.8
+-----
+- SortBar is now always visible #831
+
 2.7
 -----
 - New account verification screen #800
-- SortBar is now always visible #831
 
 2.6
 -----
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,7 @@
 2.7
 -----
 - New account verification screen #800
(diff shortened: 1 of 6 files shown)


RELEASE-NOTES.txt
```diff
@@ -1,6 +1,7 @@
 2.7
 -----
 - New account verification screen #800
+- SortBar is now always visible #831
```
eshurakov: Should it be in 2.8? 🤔

eshurakov (approved):
Nice!

RELEASE-NOTES.txt
jleandroperez: RIGHT!! Nice catch!!

jleandroperez:
Thank you sir!!

pachlava:

PR #833, opened 2021-02-01
Interface: Separators

jleandroperez:
### Fix
In this PR we're adjusting the Interface's dividers. All of them!

cc @eshurakov 
Ref. #626

### Test
- [x] Verify there's a divider between the Sort Bar and the Notes List (always visible)
- [x] Verify that a divider beween Notes Title and Sort Bar shows up, when you scroll
- [x] Verify that a divider between the Editor and the Toolbar shows up, when you scroll
- [x] Verify that a divider between the Tags List and the Window's Lights shows up, when you scroll
- [x] Verify the Editor's bottom divider (between the Editor itself and the Tags Editor) extends from edge to edge
- [x] Verify that clicking over the Sort Bar displays a Menu, which allows you to change the Sort Order

### Release

### Screenshots
Old|Nu
-|-
Old|Nu

### DIFF
--- a/Simplenote/BackgroundView.swift
+++ b/Simplenote/BackgroundView.swift
@@ -24,8 +24,7 @@ class BackgroundView: NSView {
 
     /// Bottom Border: Width
     ///
-    @IBInspectable
-    var borderWidth: CGFloat = NSScreen.main?.pointToPixelRatio ?? defaultBorderWidth {
+    var borderWidth: CGFloat = defaultBorderWidth {
         didSet {
             guard borderWidth != oldValue else {
                 return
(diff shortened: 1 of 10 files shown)


eshurakov (approved):
Looks good!

One small thing that I noticed is that in light mode the separator between tag list and the traffic light becomes light. But that I guess because of blur? 
What do you think @SylvesterWilmott?

Screenshot 2021-02-03 at 10 19 41

SylvesterWilmott:
Nice catch @eshurakov It looks like that line is light or dark based on what you have behind the UI since it's a solid colour. Can we change that particular divider to black with 0.1 alpha?

jleandroperez:
Gentleman, merging this one to unlock what comes next. I'll be updating the Tags divider as requested by Sly in a follow up!!

Thank you both for checking on this one!!

pachlava:

As mentioned above, issue with separator for tags divider is sill present, but will be addressed later.

PR #839, opened 2021-02-03
Updating Sort Bar UX

jleandroperez:
### Fix
In this PR we're adjusting the Sort Bar's UX.

cc @eshurakov 
cc @SylvesterWilmott 

Ref. #626 

### Test
- [x] Verify the Sort Bar now reads `Sort By` rather than `Sort:`
- [x] Verify the top two options are `Name: A-Z` and `Name: Z-A` rather than `Alphabetically:...`
- [x] Verify the Menu `View > Note Sorting` matches the Sort Bar menu

### Release

### DIFF
--- a/Simplenote/Base.lproj/MainMenu.xib
+++ b/Simplenote/Base.lproj/MainMenu.xib
@@ -465,19 +465,19 @@
                                 <modifierMask key="keyEquivalentModifierMask"/>
                                 <menu key="submenu" title="Note Sorting" systemMenu="font" id="YpJ-fs-p4p">
                                     <items>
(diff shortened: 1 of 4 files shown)


SylvesterWilmott:
Just to confirm, do we have tracks events for changing the sorting mode?

eshurakov (approved):
Works as advertised 👍 

:shipit:

SylvesterWilmott:
It looks good!

Let's lose the down chevron. I don't think we really need it, it doesn't appear on Electron.

jleandroperez:
Thank you gentleman!!

pachlava:

PR #837, opened 2021-02-03
Merge release 2.7 beta 2 into develop

mokagio:
This adds the account verification fix from #832 and the Apple Silicon support from #802 and #824.

This is coming from a dedicated branch because GitHub said it couldn't automatically merge, even though Git was able to resolve all the conflicts automatically on my machine.

Apologies for the huge diff. That's mainly because of the Sparkle (#824) manual update and there was no way out of it.

PR #844, opened 2021-02-05
Merge finalize version 2.7 into develop

mokagio:
This has no change in the translations because the recent https://github.com/Automattic/simplenote-macos/pull/837 already contains translations updates. There mustn't have been changes since then, which is understandable since we're at the very end of the code freeze.

---

Note that the name of the branch is messed up because I'm experimenting with a Bash command combo to generate it. It has nothing to do with the content of the branch itself. 

You can verify this by [comparing `release/2.7` against this branch](https://github.com/Automattic/simplenote-macos/compare/merge/release-2.7-sparkle-via-cocoapods-into-develop...release/2.7): there are no changes.

### DIFF
--- a/config/Version.public.xcconfig
+++ b/config/Version.public.xcconfig
@@ -2,8 +2,8 @@ VERSION_SHORT=2.7
 
 // Public long version. Example: 2.0.0.0. The last 0 means the first build for
 // version 2.0.0
-VERSION_LONG=2.7.0.1
+VERSION_LONG=2.7.0.2
 
 // This is the value for the CFBundleVersion it should be incremented on every
 // build that gets distributed
-BUILD_NUMBER=11216
+BUILD_NUMBER=11217


AliSoftware:
>Note that the name of the branch is messed up because I'm experimenting with a Bash command combo to generate it.

Are you trying to implement the idea suggested in request paaHJt-1nT-p2#comment-3513 ? 😺

AliSoftware (approved):

mokagio:
@AliSoftware oh, I had missed that specific conversation but yeah, although, I just went off a tangent and tried to do it as a single command so I can pick it up from my history for now.

It's a bit ugly... Well actually, it's beautiful how much stuff you can do in shell, but it's ugly to look at:

```
git checkout -b merge/$(git branch | grep -E "release/\d+\.\d+(\.\d+)*$" \
  | tail -n1 | sed "s/ *//" | sed "s/\//-/")-into-$(echo $(git symbolic-ref \
  -q HEAD || git name-rev --name-only --no-undefined --always HEAD) | \
  sed "s/refs\/heads\///")
```

There's likely better way to do it as a one-liner, but if we go down the path of automating this we'd likely rely on the Ruby code in the toolkit, so 🤷‍♂️

Issue #853, opened 2021-02-10
Merge Simplenote + Simplenote-AppStore Targets


jleandroperez:
### Details:
Having two targets is error prone (files can be added only to one of the two!).

Let's merge them so that we have a single Xcode target for the project, with different Build Configurations, instead.

mokagio:
When I saw #852, I thought about opening an issue along these very same lines.

_Great minds think alike_ 🤣 

I've assigned myself to this, and if you don't mind the timeline, I'd like to tackle it after #707. That way, the new build configuration will be neatly contained in `.xcconfig` file(s) with only the few parameters that differ from the App Center release.

jleandroperez:
@mokagio of course I don't mind!! Would love to team up with you, if there's anything I could help with!!

I was planning on attacking this one during the HackWeek we've got scheduled in march, but literally any timeframe works great!!

Thank you Giooooo!!

PR #852, opened 2021-02-10
Fixing AppStore Target

jleandroperez:
### Fix
In this PR we're adding to the AppStore target few new files that were only added to the development target.

cc @eshurakov 
(Thank you!!)

### Test
- [x] Verify the AppStore Target builds correctly

### Release
These changes do not require release notes.

PR #851, opened 2021-02-10
Skinning Searchbar

jleandroperez:
### Details
In this PR we're fine tunning the Search Bar's UX.

cc @eshurakov (Thanks in advance!!)
Ref. #626

### Test
- [x] Verify the Search Bar gets highlighted when you click on it
- [x] Verify that typing any keyword filters the Notes List, and highlights keywords in the Editor
- [x] Verify that hiding the Tags List causes the Search Field to compress (avoiding the Window's Semaphore)
- [x] Verify the Note Editor / Tags Editor divider is now white (in dark mode).

### Release
`RELEASE-NOTES.txt` was updated in 955822d with:

> SearchBar UI was adjusted #851

PR #854, opened 2021-02-11
Merge release 2.8 beta 3

mokagio:
cc @pachlava hey mate, here's new beta 😳 

It's unclear from the commits history because apparently I didn't open a PR against `develop` for the previous betas, but the only difference between this and beta 3 are the changes in #851 and #852.

I'd never though of if, but another way to check what new changes went into a beta is to compare the tags. E.g. if we [compare the tag for this beta with the previous one (`2.8.0.1` 👈 `2.8.0.2`)](https://github.com/Automattic/simplenote-macos/compare/2.8.0.1...2.8.0.2), the list of commits shows the two PRs that went in beta 3 since beta 2. I only mentioned #853 because #852 is a build infrastructure change only.


jleandroperez (approved):
Thank you Gio!!

:shipit:

mokagio:
Closing in favor of #858 .

PR #857, opened 2021-02-11
Issue/update simplenote search version

charliescheer:
### Fix
This is a quick PR to prepare Simplenote Mac for the most recent changes to SimplenoteSearch.

All that is being changed here is the addition of a default value for SearchQuerySettings.default and updating one search method call to include the use of the settings, and update the NoteList  tests to use the settings as well.

### Test
This update should not modify the functionality of the app.
Make sure to test that searching in the app and tag searches still work

### Review
one developer is required to review these changes, but anyone can perform the review.

### Release


jleandroperez (approved):
Looking sharp!!

:shipit:

PR #863, opened 2021-02-16
Build App Store target on CI

mokagio:
With two different targets, it can sometime happen to forget to add new source files to the target that's not currently selected.

The long term solution to this problem is to consolidate the two targets in favor of a new build configuration (#853) but while we wait for that to happen, let's make sure the App Store target can build on CI.

Based off `release/2.8` because `develop` doesn't build at the moment. #854 will fix that.

### Test

You can see the CI build working [here](https://app.circleci.com/pipelines/github/Automattic/simplenote-macos/1989/workflows/925e6a55-0f0d-44b6-9167-c487867f8e34).

Screen Shot 2021-02-16 at 2 52 03 pm

### Review

### Release

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -32,7 +32,7 @@ jobs:
     steps:
       - setup_environment
       - run:
-          name: Build
+          name: Build & Test
           # We've seen some builds timeout because CircleCI didn't receive
           # output for 10m. Making it wait for 20m instead should address that.
           #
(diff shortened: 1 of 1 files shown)


AliSoftware (approved):
LGTM 👍 :shipit: 

(Surprised we don't use `fastlane` and `gym` to build the product here, but 🤷 that's for another day)

AliSoftware:
PS: You might want to update the [GitHub branch settings](https://github.com/Automattic/simplenote-macos/settings/branch_protection_rules/17145567) (†) to make this new CI check _required_ once you've merged that change.

_(†) link is for `develop` but should probably be applied to other branches like `trunk` and `release/*` too._

image

jkmassel (approved):
Good with me as well, strong 👍 to the "we should use fastlane for this" comment from @AliSoftware

mokagio:
Thanks for the reviews and suggestions folks. 

I updated the settings (totally forgot!) and created #869 so I won't forget about updating to use Fastlane.

mokagio:
Changed base to `develop` now that #854 went in.

PR #870, opened 2021-02-17
Use Fastlane on CI to run the tests and build the app

### LINKED ISSUE #869: Use Fastlane to build and test the app on CI
- Consistent with the rest of our infrastructure
- Unlocks the nice benefits of having the Fastlane maintainers deal with the occasional `xcodebuild` oddities under the hood
- Since we use / will use Fastlane to upload builds to App Center and the App Store, using the same tool for the tests in CI will add an extra check for the build configuration

Prompted by https://github.com/Automattic/simplenote-macos/pull/863#pullrequestreview-591161389.

mokagio:
Fixes #869.

Based on top of #874 because otherwise the App Store target won't build.

### Test
Nothing to test, if CI is green we're good.

### Review

### Release


.circleci/config.yml
```diff
@@ -28,47 +28,22 @@ jobs:
...
-          command: |
-            # Build without code signing to avoid missing cert errors
-            xcodebuild clean test \
-                -workspace 'Simplenote.xcworkspace' \
-                -scheme 'Simplenote' \
-                -configuration 'Debug' \
-                COMPILER_INDEX_STORE_ENABLE=NO \
-                CODE_SIGN_IDENTITY="" \
-                CODE_SIGNING_REQUIRED=NO
+          command: bundle exec fastlane test
```
mokagio, line 36: Love how less code there's in the CircleCI layer. Would/Will make it simpler to migrate to a different provides.

.circleci/config.yml
```diff
@@ -28,47 +28,22 @@ jobs:
   Test:
     executor:
       name: ios/default
-      xcode-version: "12.0"
+      xcode-version: "12.4.0"
```
mokagio, line 31: Did this just to get on a more up to date version.

Gemfile.lock
```diff
@@ -117,10 +118,11 @@ GEM
     faraday-net_http (1.0.1)
     faraday_middleware (1.0.0)
       faraday (~> 1.0)
-    fastimage (2.2.1)
-    fastlane (2.171.0)
+    fastimage (2.2.2)
+    fastlane (2.174.0)
```
mokagio, line 122: Before doing any work with Fastlane, better to get on the latest version 😄

.circleci/config.yml
AliSoftware: nod

fastlane/Fastfile
```diff
@@ -13,6 +13,20 @@ ENV["DOWNLOAD_METADATA"]="./fastlane/download_metadata.swift"
 
 USER_ENV_FILE_PATH = File.join(Dir.home, '.simplenotemacos-env.default')
 
+workspace = 'Simplenote.xcworkspace'
+internal_scheme = 'Simplenote'
+app_store_scheme = 'Simplenote-AppStore'
+build_folder = 'build'
```
AliSoftware: nit: since those are constants, you should scream them in all-uppercase. 👮‍♂️  #manual-rubocoping

```suggestion
WORKSPACE = 'Simplenote.xcworkspace'
INTERNAL_SCHEME = 'Simplenote'
APP_STORE_SCHEME = 'Simplenote-AppStore'
BUILD_FOLDER = 'build'
```

(+ then of course update the references in rest of code)

fastlane/Fastfile
```diff
@@ -13,6 +13,20 @@ ENV["DOWNLOAD_METADATA"]="./fastlane/download_metadata.swift"
...
+
+# On CI, it's useful to skip code signing unless necessary to avoid spending
+# time syncing the certificates and profiles with match. And while not
+# neccessary, it doesn't hurt on a dev machine either.
+skip_code_signing_xcargs = [
+  'COMPILER_INDEX_STORE_ENABLE=NO',
+  "CODE_SIGN_IDENTITY=''",
+  'CODE_SIGNING_REQUIRED=NO',
+]
```
AliSoftware: I made it possible to pass a `Hash` to `xcargs` a while ago (https://github.com/fastlane/fastlane/pull/8051) so that it shell-escapes everything for you and makes it more readable; we should use this here 😇 

```suggestion
skip_code_signing_xcargs = {
  COMPILER_INDEX_STORE_ENABLE: 'NO',
  CODE_SIGN_IDENTITY: '',
  CODE_SIGNING_REQUIRED: 'NO',
}
```

fastlane/Fastfile
```diff
@@ -29,6 +43,30 @@ before_all do
   end
 end
 
+desc 'Runs the unit tests for the project'
+lane :test do | options |
+  run_tests(
+    workspace: workspace,
+    scheme: internal_scheme,
+    clean: true,
+    xcargs: skip_code_signing_xcargs.join(' ')
```
AliSoftware: Join is handled by https://github.com/fastlane/fastlane/pull/8051 already if you pass a `Hash` as per previous comment/suggestion.

```suggestion
    xcargs: skip_code_signing_xcargs
```

fastlane/Fastfile
```diff
@@ -29,6 +43,30 @@ before_all do
...
+lane :build_app_store do | options |
+  build_mac_app(
+    workspace: workspace,
+    scheme: app_store_scheme,
+    configuration: 'Release',
+    clean: true,
+    output_directory: build_folder,
+    # Skip archive so we don't need to worry about code signing
+    skip_archive: true,
+    xcargs: skip_code_signing_xcargs.join(' ')
```
AliSoftware: Ditto.
```suggestion
    xcargs: skip_code_signing_xcargs
```

fastlane/Fastfile
```diff
@@ -176,8 +214,8 @@ end
     cocoapods()
 
     build_app(
-      workspace: "Simplenote.xcworkspace", 
-      scheme: "Simplenote",
+      workspace: workspace,
+      scheme: internal_scheme,
       clean: true,
       output_directory: "./build/")
```
AliSoftware: Combine some [Han Solo](https://github.com/woocommerce/woocommerce-android/pull/3592#pullrequestreview-592833189) (reuse the constant) + me playing human rubocop (closing parentheses alignment), and you get:
```suggestion
      output_directory: BUILD_FOLDER
    )
```

fastlane/Fastfile
```diff
@@ -29,6 +43,30 @@ before_all do
   end
 end
 
+desc 'Runs the unit tests for the project'
+lane :test do | options |
+  run_tests(
+    workspace: workspace,
+    scheme: internal_scheme,
```
AliSoftware: Even if that's the default configuration set up on the project (is it?), I think it's better to be explicit here anyway:
```suggestion
    scheme: internal_scheme,
    configuration: 'Debug',
```

AliSoftware (commented):
Looking good in general, but left a couple of nits that would be nice to be addressed 🙂

AliSoftware:
Side note: is it intentional that this repo is not configured to "prevent PR merge until the PR gets at least one approval"?
(Is it because there are not enough people working on SN so it would slow down the process too much?)

fastlane/Fastfile
mokagio: I DIDN'T KNOW YOU HAD TO SCREAM WHEN DEFINING CONSTANTS... Although, I saw it around enough times that I should have figured it out 😅 .

Would they remain uppercase if they became global values? That's a change I'm considering doing later while extracting some logic in a function.

```ruby
@WORKSPACE = ...

# or 

@workspace = ...
```

AliSoftware: Global **variables** should not be all uppercase. The convention-over-configuration rule in ruby about this is that all-uppercase means the value should be considered **constant** — like if you `.freeze` it (in fact it's also good practice to `.freeze` constants, even though I admit I always forget).

The only reasonable reason to move those to be global **variables** instead is if you really need to make those values be **variable**, ie you plan to change their value during the script. And if you do, that means they won't be constant anymore so shouldn't be all-uppercase anymore.

That being said, global variables are frown upon in ruby, like they are in many other languages — even if not as bad as class variables (I even think that rubocop considers `@@classvar` usage to be a big warning if not an error), but still not ideal. And I can't really see a reason why you'd need those values in your script in particular (workspace name and scheme name etc) to be (global) **variables** rather than **constants** in that case anyway?

_(And also, `@` is for **instance** variables not global variables. I rarely use _global variables_ in ruby because it's generally bad practice anyway, but I _think_ it's supposed to use the `$` prefix? Would need to Google that to be sure though)

mokagio:
@AliSoftware thank you for the review and sorry I have't been able to follow up on your great suggestions yet.

> Side note: is it intentional that this repo is not configured to "prevent PR merge until the PR gets at least one approval"?

🤔 maybe it was a glitch, or someone enabled it in the meantime? Anyways it's enabled now.

image

mokagio:
Everything should be addressed now, thanks @AliSoftware.

AliSoftware (approved):

PR #872, opened 2021-02-17
Decoupling PopoverWindowController: Mk 2

jleandroperez:
### Details
In this PR we're implementing **InterlinkProcessor** (and dropping the Interlink processing logic from the EditorViewController!)

cc @eshurakov 
(Thanks in advance!!)

### Test
1. Open any note
2. Type `[` + keyword (Note that **keyword** must be present in any of your note's titles)

- [x] Verify the Interlink Popover shows up
- [x] Verify that you can move the Keyboard Arrows (Up / Down) to cycle thru results
- [x] Verify that typing anything else causes the UI to refresh the results
- [x] Verify that pressing **Return** inserts the Interlink
- [x] Verify that clicking elsewhere (while the popover is visible) dismisses the Interlink UI

### Release

### DIFF
--- a/Simplenote/InterlinkProcessor.swift
+++ b/Simplenote/InterlinkProcessor.swift
@@ -1,0 +1,137 @@
+import Foundation
+
+
+// MARK: - InterlinkProcessor
+//
+class InterlinkProcessor: NSObject {
+
+    /// Interlink Popover
+    ///
+    private lazy var interlinkWindowController: PopoverWindowController = {
+        let popoverWindowController = PopoverWindowController()
+        popoverWindowController.contentViewController = interlinkViewController
+        return popoverWindowController
+    }()
+
+    /// Interlink ViewController!
+    ///
+    private lazy var interlinkViewController = InterlinkViewController()
+
+    /// Hosting TextView
+    ///
+    private let parentTextView: SPTextView
+
+
+    /// Designated Initialier
+    ///
+    init(parentTextView: SPTextView) {
+        self.parentTextView = parentTextView
+    }
+
+
+    /// Displays the Interlink Lookup Window at the cursor's location when all of the following are **true**:
+    ///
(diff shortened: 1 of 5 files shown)


Simplenote/InterlinkProcessor.swift
```diff
@@ -0,0 +1,137 @@
...
+    ///
+    func refreshInterlinks(for keywordText: String, in replacementRange: Range<String.Index>, excluding excludedID: NSManagedObjectID?) -> Bool {
+        interlinkViewController.onInsertInterlink = { [weak self] text in
+            self?.parentTextView.insertTextAndLinkify(text: text, in: replacementRange)
+            self?.dismissInterlinkWindow()
+        }
+
+        return interlinkViewController.refreshInterlinks(for: keywordText, excluding: excludedID)
+    }
+}
```
eshurakov, line 137: I find this method confusing, because it does quite a few different things :-)
* because it is called `refresh..` and returns boolean. Having 0 results I think is still a success. What do you think about having a different method to check if there are interlinks to display or not? Can call it `isempty` or `hasResults`
* do you think maybe it's worth extracting `onInsertInterlink` listener into separate method? And maybe set it up before or after we show the window, around this line `displayInterlinkWindow(around: keywordRange)`

eshurakov (approved):
Looks good, sending you a small suggestion :-)

:shipit:

Simplenote/InterlinkProcessor.swift
eshurakov: Update: I see you changed this in the next PR 😆 
Please disregard the comment above.

jleandroperez:
Thank you sir!!

pachlava:
[testing] fix/feature verified on MBP 16" Intel | macOS Big Sur 11.2 | Simplenote 2.9 (11223)

PR #883, opened 2021-02-22
Update secrets

mokagio:
### Fix
Same as https://github.com/Automattic/simplenote-ios/pull/1166, but with
more files changed because we hadn't updated this in a while.

### Test
Run `bundle exec fastlane run configure_apply` to verify your local copy of the files touched in this diff change.

### Review

### Release


jleandroperez:
Thank you Gio!!

PR #889, opened 2021-02-25
Dropping Login Inheritance

jleandroperez:
### Fix
In this PR we're dropping inheritance from the Authentication UI, making it easier to modify the signup flow.

cc @eshurakov (Thanks in advance!!)
cc @charliescheer We're taking a few shortcuts, this code is far from beautiful!

### Test
- [x] Verify the Authentication UI looks great
- [x] Verify you can log into your account
- [x] Verify Signup still works
- [ ] Verify WordPress SSO also works well
- [x] Verify that entering invalid credentials cause the UI to display an error
- [x] Verify that while the app is Signing In / Signing Up, a spinner shows up, and the UI is locked

### Release

### DIFF
--- a/Simplenote/LoginWindowController.h
+++ b/Simplenote/LoginWindowController.h
@@ -1,22 +1,10 @@
-//
-//  LoginWindowController.h
-//  Simplenote
-//
-//  Created by Jorge Leandro Perez on 10/26/15.
-//  Copyright © 2015 Simperium. All rights reserved.
-//
-
 #import <Foundation/Foundation.h>
 #import <AppKit/AppKit.h>
 @import Simperium_OSX;
 
 
 
-/**
- *  @class      LoginWindowController
- *  @brief      The purpose of this class is to extend Simperium's Authentication screen, and provide
- *              extra functionality, Simplenote-Y.
- */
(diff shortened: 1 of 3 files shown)


charliescheer:
Worked my way through the testing steps above, and had an issue with the WP SSO.  When I clicked on Sign in with Wordpress everything acted as I would expect, I was taken to Wordpress and given a chance to authenticate -> I Authenticated -> Safari asked if I wanted to open SN -> The login window returned and the app was not logged in.  I also shortly after that received an email telling me that an account for that email already existed.  So it seems that it was trying to create me an account and not login with the existing account.

charliescheer (commented):
Checked out the code and it all looks good to me! :)   

but something seems to be off about the sign in with apple trying to create me an account where one already exists

eshurakov (approved):
Looks good to me. 

I checked sign in with Wordpress and it worked for me. But I had to delete all other Simplenote apps, otherwise the callback link wouldn't open in a dev app.

jleandroperez:
Thank you both!! Merging since the SSO code wasn't modified here (not sure what might have been going on, but that flow falls on the web side!)

PR #895, opened 2021-03-03
Add lane to deploy to App Store and publish to GitHub

mokagio:
Originally, I had a CI part in this PR, too. I removed it to implement the API-driven deployment as opposed to the tag-driven workflow (see https://github.com/Automattic/simplenote-ios/pull/1214) instead.

Note: I'm pushing this to `release/2.10` in the hope that I'll be able to benefit of these improvements when it comes the time to make the build for the App Store.

### Test

- Checkout this PR and create a new branch from it
- Bump the `BUILD_NUMBER` in `config/Version.public.xcconfig` otherwise ASC will reject your build (via email, the upload will still work, unlike what happens with TestFlight 🙄 ) – ~~I run my test with `11227`, so the first reviewer should use `11228`, the second `11229` and so on...~~ The build number from #932 is `11230` so the first person reviewing should use `11231`, if there's a second reviewer, they should use `11232`, and so on...
- Run `bundle exec fastlane run configure_apply` to get the latest version of the secrets
- Add two new keys to your to your environment file `~/.simplenotemacos-env.default`,  `APP_STORE_CONNECT_API_KEY_PATH` and `SIGH_API_KEY_PATH` both with value `".configure-files/app_store_connect_fastlane_api_key.json"` see `user.env-example` for more info.
- `bundle install && bundle exec fastlane build_and_upload_app_store create_github_release:true`

It should succeed, upload your build to App Store Connect, and create a new release in this repo.

<table>
<tr>
<td>

image

</td>
<td>

image

</td>
</tr>
</table>

Once verified:

- Delete your test branch
- Delete the release from GitHub

### Review

### Release


fastlane/Fastfile
```diff
@@ -13,15 +13,15 @@ ENV["DOWNLOAD_METADATA"]="./fastlane/download_metadata.swift"
...
 USER_ENV_FILE_PATH = File.join(Dir.home, '.simplenotemacos-env.default')
 
-WORKSPACE = 'Simplenote.xcworkspace'
-INTERNAL_SCHEME = 'Simplenote'
-APP_STORE_SCHEME = 'Simplenote-AppStore'
-BUILD_FOLDER = 'build'
+@WORKSPACE = 'Simplenote.xcworkspace'
+@INTERNAL_SCHEME = 'Simplenote'
+@APP_STORE_SCHEME = 'Simplenote-AppStore'
+@BUILD_FOLDER = 'build'
```
mokagio: I turned these into global constants to use them in the helper function I added below (see also the comment for it).

fastlane/Fastfile
```diff
@@ -36,6 +36,29 @@ def get_required_env(key)
   ENV[key]
 end
 
+# Helper function to DRY the code to build the app
+def build_simplenote(codesign:, archive_path: nil)
```
mokagio, line 43: I've never done or seen this on a `Fastfile`, but I liked the idea of using the keyword arguments to make it clear to the caller what values to pass. As opposed to using a private lane, where it's not clear (to me) what to pass every time.

What do you think? Too original?

fastlane/Fastfile
```diff
@@ -46,26 +69,17 @@ end
...
-    scheme: APP_STORE_SCHEME,
-    configuration: 'Release',
-    clean: true,
-    output_directory: BUILD_FOLDER,
-    # Skip archive so we don't need to worry about code signing
-    skip_archive: true,
-    xcargs: SKIP_CODE_SIGNING_XCARGS
-  )
+  build_simplenote(codesign: false)
 end
```
mokagio: I made this lane for the CI step that ensures the App Store target builds. There are two different _targets_, actual targets, see #853.

That might not be the case for long, but we'll still need to have a clean option to make the Internal and App Store builds, regardless of whether they live in two targets or the same target but different build configurations.

fastlane/Fastfile
```diff
@@ -238,12 +252,39 @@ end
...
+
+    archive_path = File.join(@BUILD_FOLDER, 'Simplenote-Mac.xcarchive')
+    build_simplenote(codesign: true, archive_path: archive_path)
+
+    archive_zip_path = "#{archive_path}.zip"
+    zip(path: archive_path, output_path: archive_zip_path)
+
+    # TODO: Upload to App Store Connect
+
+    # TODO: Add "create_release" parameter to opt out from here
```
mokagio: These TODOs will obviously be addressed before this gets out of the Draft state.

mokagio (commented):
@Automattic/release-managers I'd like some feedback on the helper functions + keyword parameter idea. Thanks 🙇‍♂️ .

.circleci/config.yml
```diff
@@ -6,14 +6,16 @@ orbs:
...
 commands:
-  setup_environment:
+  copy_demo_credentials:
     steps:
-      - checkout
       - run:
           name: Copy Demo SPCredentials
           command: |
             mkdir -p Simplenote/Credentials
             cp Simplenote/SPCredentials-demo.swift Simplenote/Credentials/SPCredentials.swift
```
mokagio, line 15: Rationale for having this outside `setup_environment` here: 558e932995ba7d4ef23c8e3c9c6d15247ea2b6ca

.circleci/config.yml
```diff
@@ -41,12 +44,31 @@ jobs:
...
+      - setup_environment
+      - run:
+          name: Decrypt Secrets
+          command: bundle exec fastlane run configure_apply
+      - run:
+          name: Install Sentry CLI
+          command: brew install getsentry/tools/sentry-cli
+      - run:
+          name: Build and Upload for App Store
+          command: bundle exec fastlane build_and_upload_app_store create_github_release:true
```
mokagio: What do you think of having to explicitly pass `true` here?

The other projects have this logic, too, only the value is computed at runtime based on the kind of tag and on the `is_ci` return value.

Simplenote.xcodeproj/project.pbxproj
```diff
@@ -2716,7 +2716,7 @@
 				PRODUCT_NAME = Simplenote;
 				PROVISIONING_PROFILE = "";
 				"PROVISIONING_PROFILE[sdk=macosx*]" = "";
-				PROVISIONING_PROFILE_SPECIFIER = "Simplenote Mac Production Profile";
+				PROVISIONING_PROFILE_SPECIFIER = "";
```
mokagio: Removed the hardcode values so we can use whatever `match` fetches for us.

fastlane/Fastfile
```diff
@@ -36,6 +36,34 @@ def get_required_env(key)
...
+  build_mac_app(
+    workspace: @WORKSPACE,
+    scheme: @APP_STORE_SCHEME,
+    configuration: 'Release',
+    clean: true,
+    output_directory: @BUILD_FOLDER,
+    archive_path: archive_path,
+    skip_archive: skip_archive,
+    export_method: 'mac-application',
+    export_team_id: CredentialsManager::AppfileConfig.try_fetch_value(:team_id),
```
mokagio, line 62: Other projects read this value from an encrypted dotfile. I'm not sure why, since the value is tracked in clear as part of the `Appfile`.

I don't have a problem with using the encrypted dotfile. I didn't set it up here because it's extra work... 😅 If we want to do that, though, should we then remove the value from the `Appfile`? I don't think it's a security risk. Having the team identifier is useless without an account (_last famous words?_)

fastlane/Fastfile
AliSoftware: Yeah, tbh I never really understood why fastlane invented the concept of private_lane when using regular functions via `def` already provides the same functionality 🤷 

The only difference of using `def` vs using `private_lane` at runtime is that you will see "Cruising to private lane XYZ" in the log/output when you were running your main lane and it calls out to the private one; then "Cruising back to lane ABC" once it's back. While a call to a function won't be logged into the fastlane output. So, each solution has its pros and cons 🤷‍♂️

.circleci/config.yml
```diff
@@ -41,12 +44,31 @@ jobs:
...
       - setup_environment
+      - copy_demo_credentials
       - run:
           name: Build App Store Target
           command: bundle exec fastlane build_app_store
 
+  Upload to App Store:
+    executor:
+      name: ios/default
+      xcode-version: "12.4.0"
```
AliSoftware: nit: wdyt of using a YAML anchor to DRY the `xcode-version:` value that is currently repeated in each job?

On first occurrence in the YAML:
```yaml
executor: &xcode-image
  name: ios/default
  xcode-version: "12.4.0"
```

On all subsequent jobs needing the same executor:
```yaml
executor: *xcode-image
```

.circleci/config.yml
AliSoftware: I actually find this (passing explicitly like here) better personally. Though in the end it's nbd.

fastlane/Fastfile
AliSoftware: This is not the syntax for global constants in ruby though:
 - all-uppercase (without any prefix char) = to declare a **constant**. Think `let` in Swift. If that constant is declared at the top level, it's a global one, if it's declared inside some scope (inside a function or whatever), it's a local one.
 - lowercase = a **variable**. Like `var` in Swift.
 - prefixed with `@` = **instance variable** (more especially the _backing storage_ of attributes in a ruby class, for which you usually have a corresponding `attr_reader` or `attr_accessor` to provide public accessors). Think `@foo` ≈ the private ivar you put in a `.h` in ObjC, and `attr_reader :foo` ≈ the associated public `@property foo` accessor for it)
 - _(There are other char-prefixes existing in ruby with different meanings, like `$` and `@​@`, but they are not relevant to the conversation here)_

Here you're using a strange mix, and tbh I have no idea how ruby consider those cases. An instance variable `@` but declared outside of the scope of a `class`? So not an instance thing after all… and all-uppercase so a constant, but with `@` = ??

TL;DR: you could have kept the `WORKSPACE = ` without the `@` prefix all along. Since those constants are declared at top-level, there's no problem in referencing/using them anywhere else in the `Fastfile`, be it inside a lane or a function or wherever.

fastlane/Fastfile
AliSoftware: Yeah I don't see the point of having the team ID be encrypted and pulled from secrets. I like your solution here better.

fastlane/Fastfile
```diff
@@ -238,12 +257,71 @@ end
       project_slug: 'simplenote-macos',
       dsym_path: dSYM_PATH,
     )
-   
+  end
+
+  lane :build_and_upload_app_store do | options |
```
AliSoftware, line 281: This would benefit from a `desc` documentation – including documenting what are the valid `options` you can pass to this lane (same for all other lanes tbh, could be worth documenting valid options for all of them)

fastlane/Fastfile
```diff
@@ -238,12 +257,71 @@ end
...
+    zip(path: archive_path, output_path: archive_zip_path)
+
+    create_release(
+      repository: GHHELPER_REPO,
+      # version: ios_get_app_version(),
+      version: "test-#{Time.now.to_i}",
+      release_notes_file_path: 'Simplenote/Resources/release_notes.txt',
+      release_assets: [archive_zip_path],
+      prerelease: options[:prerelease],
+    )
```
AliSoftware: ~This action should be called conditionally. At least that's what is suggested by the fact that you call the `build_and_upload_app_store` lane in `.circleci/config.yml` using `create_github_release:true`… that you don't seem to use here~

nvm, I missed that you had the `next unless options[:create_github_release]` above.

Usually I like guard statements to exit early, but having one here in the middle of the code (as opposed to at having guard clauses at the beginning of the function/lane, which is the usual place where we find/use them) makes me think they are harder to spot (proof: I almost missed it 😛 ) in such contexts (= mid-functions). wdyt of nesting the code inside a big `if` instead here?

AliSoftware (commented):
Left a couple of suggestions.

PS: I checked and `configure-files/google_cloud_keys.json.enc` file is not empty (it's 2KB), despite what the GitHub UI says (not the first time GitHub UI is misleading us on encoded/non-textual files)

It could actually be worth adding `*.enc` files to the `.gitattributes` file (and commit that change) to specify that they should be treated as binary files, not textual ones (could help both display of diffs but also help git merge + conflict auto-resolution strategies)

AliSoftware:
> PS: I checked and `configure-files/google_cloud_keys.json.enc` file is not empty (it's 2KB), despite what the GitHub UI says (not the first time GitHub UI is misleading us on encoded/non-textual files)
> 
> It could actually be worth adding `*.enc` files to the `.gitattributes` file (and commit that change) to specify that they should be treated as binary files, not textual ones (could help both display of diffs but also help git merge + conflict auto-resolution strategies)

Follow-up: https://github.com/Automattic/simplenote-macos/pull/899

fastlane/Fastfile
mokagio: Thanks @AliSoftware for the Ruby lesson. I looked for that information online but couldn't find a good post about it. I might have used a wrong query, though, because the mental model I had was not accurate.

As you may remember, [I originally extracted `WORKSPACE` and those other values using the lowercase syntax](https://github.com/Automattic/simplenote-macos/pull/870#discussion_r578371300), that is, as variables. I did that work as part of the original version of this branch, and when I tried to access them in the helper function, it didn't work. The only way to make it work in my flawed assumption that uppercase was just a convention rather than a language feature, was to make them instance variable, prefixing `@`, which in my mental model was (not sure why?) the equivalent of a global constant.

After applying your correction to make those values proper Ruby constants, I rebased the changes and kept the `@`; it didn't occur to me to check if it worked.

I know better now, thanks 🙇‍♂️

mokagio: I wanted to verify all that before posting it, and now that I wrote that code I can't let go of it. So here it is, my process in pictures.

First step: extract stuff, feel good about yourself, then wonder why Fastlane can't find the value.

image

Second step: make it an instance variable, because... not sure, maybe it worked in the past in other scripts I wrote for some reason? 🤷‍♂️ 

image

Third step: learn proper Ruby and have neat 🌯

image

fastlane/Fastfile
mokagio: :nods: I should have added a TODO about it. Was just leaving it for last.

fastlane/Fastfile
mokagio: I prefer it like this because it's linear, but I do see your point.

Other projects use the `if` approach, too. Between that and your point, I'm happy to [disagree and commit](https://en.wikipedia.org/wiki/Disagree_and_commit) 👍 (That is, I'll update to use the `if`)

.circleci/config.yml
mokagio: Good point. I keep thinking about it and never do it. _Later is never_.

fastlane/env/user.env-example
```diff
@@ -1,3 +1,7 @@
 GHHELPER_ACCESS=
 APPCENTER_API_TOKEN=
-SENTRY_AUTH_TOKEN=
\ No newline at end of file
+SENTRY_AUTH_TOKEN=
+
+# These should point to the same path
+APP_STORE_CONNECT_API_KEY_PATH=
+SIGH_API_KEY_PATH=
```
mokagio: If I'll have time this week or the next, I'll try opening a PR on Fastlane to make `sigh` (used by `match` under the hood) fallback to `APP_STORE_CONNECT_API_KEY_PATH` if `SIGH_API_KEY_PATH` is not given.

Simplenote.xcodeproj/project.pbxproj
mokagio: This turned out to be a mistake! For some reason, it resulted in the sandboxing entitlement not being ported over to the `.pkg` which failed the upload to App Store Connect.

mokagio:
Switched the base branch back to `develop` and bumped to 2.11 milestone because we missed the window for 2.10 (which I just finalized, see #925).

mokagio:
_I moved this to 2.12, but then told myself to be ambitious and aim for this release, so back to 2.11!_

fastlane/env/user.env-example
AliSoftware: Can't we set these values on the `project.env` file, like we did for WCiOS and WPiOS (see `~/.mobile-secrets/iOS/{WP,WC}iOS/project.env`)?

That would avoid having each developer / RM set those manually in their user env since those env vars will always point to the same path for anyone using mobile-secrets anyway…

.circleci/config.yml
```diff
@@ -26,21 +28,23 @@ commands:
 
 jobs:
   Test:
-    executor:
+    # This uses a YAML anchor to make the configuration repeatable in the rest
+    # of the file.
+    executor: &xcode_image
```
AliSoftware, line 33: ❤️

fastlane/Fastfile
```diff
@@ -37,6 +37,38 @@ def get_required_env(key)
...
+    skip_archive: skip_archive,
+    export_method: 'app-store',
+    export_team_id: CredentialsManager::AppfileConfig.try_fetch_value(:team_id),
+    xcargs: xcargs,
+    # When exporting for the App Store, we need a .pkg file because that's the
+    # only format upload_to_app_store supports.
+    #
+    # Note for the futures: this might need to be disabled for the developer-id
+    # export, see
+    # https://github.com/fastlane/fastlane/issues/15963#issuecomment-593649185
```
AliSoftware: nit:
```suggestion
    # Note for the future: this might need to be disabled for the developer-id export
    # see https://github.com/fastlane/fastlane/issues/15963#issuecomment-593649185
```

fastlane/Fastfile
```diff
@@ -37,6 +37,38 @@ def get_required_env(key)
...
+# Helper function to DRY the code to build the app
+def build_simplenote(codesign:, archive_path: nil)
+  if codesign
+    skip_archive = false
+    xcargs = []
+  else
+    # Skip archiving when not codesigning
+    skip_archive = true
+    xcargs = SKIP_CODE_SIGNING_XCARGS
+  end
```
AliSoftware, line 51: Completely up to style preferences, just putting this here in case you like it:
```suggestion
  skip_archive, xcargs = if codesign
    [false, {}]
  else
    [true, SKIP_CODE_SIGNING_XCARGS] # Skip archiving when not codesigning
  end
```
Note: not sure about the guidelines for indentation in such cases; I'll let you check with `rubocop`

fastlane/Fastfile
AliSoftware: I found it a bit confusing that the lane was called `build_app_store` but that you would use `codesign: false`. Without any more context, I would think that this lane would be used by `build_and_upload_app_store` and would require codesign and would think this was a bug.

Maybe there can be a better name for this then, like `ensure_builds_for_appstore`? (not loving that one either tbh). Or make it more clear in the `desc` and via comments that this is intended only to test that the project just builds, but not intended to upload the generated build to app store?

fastlane/Fastfile
```diff
@@ -239,7 +262,80 @@ end
...
+      # This Mac app also needs a Mac Installer Distribution certificate
+      additional_cert_types: ['mac_installer_distribution'],
+      # Will default to false, unless a readonly value is specified, or is
+      # running on CI, in which case it will always be true.
+      readonly: options[:readonly] || is_ci,
+      app_identifier: 'com.automattic.SimplenoteMac'
+    )
+
+    archive_path = File.join(BUILD_FOLDER, 'Simplenote-Mac.xcarchive')
+    build_simplenote(codesign: true, archive_path: archive_path)
```
AliSoftware, line 299: This is where I was kinda expecting for you to call `build_app_store` until I realized the true intent of that other lane 😛

AliSoftware (commented):
Took another look at the code, overall LGTM 👍  but left a couple of suggestions about the env vars / `project.env` + about trying to find better naming or doc.

Didn't try to actually do a build because I need to run, will try to follow the test instructions to trigger an actual build later if no one beats me to it.

fastlane/env/user.env-example
mokagio: Oh this is a great suggestion, thanks! I started with `user.env` to get it to work locally and forgot about the `project.env` option. 🙇‍♂️

mokagio: I started implementing this when I thought... why do we need to set it as an environment variable in the first place? I mean the file itself is encrypted, and `deliver` has an `api_key_path` parameter.

Maybe it would be simpler and more direct to pass the path using that parameter on the `deliver` call site. What do you think?

I might be missing some rationale on why we're using the environment variable. Summoning @jkmassel for that.

Thanks.

fastlane/Fastfile
mokagio: Ah! Good catch, thanks!

Addressed in 065dafae

fastlane/Fastfile
mokagio: What about `test_app_store_build`? This one is confusing, too, because it doesn't actually run the tests. I'll go for this one and update if we can find a better alternative.

By the way, this should be a _temporary**™**_ check, just till we address https://github.com/Automattic/simplenote-macos/issues/853. That's to say, even if we don't settle on a perfect naming it's fine because it'll be gone "soon".

fastlane/env/user.env-example
AliSoftware: I thought about it too and the only rationale I could think of is so that it can work and be overridden by people not using the mobile secrets, like external contributors. But then I can't see why non A12s would want to or be allowed to run those lanes in the first place, so yeah…

fastlane/Fastfile
AliSoftware: Another option is to:
- get rid of `build_simplenote()` function and move its code to this lane directly
- Call this `build_app_store` lane from `build_and_upload_appstore` (instead of `build_simplenote`) then
- And just explicitly use `codesign:false` in the CI config at call site when the intent is only to test the build works but not upload it

Would avoid having to find names that could end up being confusing and instead make it clear at call site that codesign is disabled in CI when the job is about building without uploading

fastlane/env/user.env-example
mokagio: True 🤔 

I don't think there's a good ROI in keeping the release process flexible so that an external contributor can run it.

Let's see what @jkmassel has do say about this...

fastlane/Fastfile
jkmassel: I agree with this – the team ID is not a secret, no need to treat it as one. I'd even go one step further and say "let's get rid of the `Appfile` – IMHO having a ton of files that just have configuration is less good than having one file that has all the lanes + some configuration at the top that's shared between all of them.

Doesn't have to be part of this PR though.

fastlane/Fastfile
```diff
@@ -55,18 +87,9 @@ lane :test do | options |
...
-    workspace: WORKSPACE,
-    scheme: APP_STORE_SCHEME,
-    configuration: 'Release',
-    clean: true,
-    output_directory: BUILD_FOLDER,
-    # Skip archive so we don't need to worry about code signing
-    skip_archive: true,
-    xcargs: SKIP_CODE_SIGNING_XCARGS
-  )
+desc 'Builds the app using the App Store target but without codesignig, just to check the target builds'
```
jkmassel:
```suggestion
desc 'Builds the app using the App Store target but without code signing, just to check the target builds'
```

fastlane/Fastfile
```diff
@@ -55,18 +87,9 @@ lane :test do | options |
...
+lane :test_app_store_build do | options |
+  build_simplenote(codesign: false)
```
jkmassel, line 94: This reads very nicely 😍

fastlane/Fastfile
```diff
@@ -239,7 +262,80 @@ end
...
+  # Example:
+  # bundle exec fastlane build_and_upload_app_store
+  # bundle exec fastlane build_and_upload_app_store create_github_release:true
+  #############################################################################
+  lane :build_and_upload_app_store do | options |
+    # This allows code signing to work on CircleCI. It is skipped if this isn't
+    # running on CI.
+    #
+    # See https://circleci.com/docs/2.0/ios-codesigning/
+    setup_circle_ci
```
jkmassel, line 286: For other projects we have this in the `before` hook. Maybe it doesn't matter though?

fastlane/Fastfile
```diff
@@ -239,7 +262,80 @@ end
...
+
+    sync_code_signing(
+      type: 'appstore',
+      platform: 'macos',
+      # This Mac app also needs a Mac Installer Distribution certificate
+      additional_cert_types: ['mac_installer_distribution'],
+      # Will default to false, unless a readonly value is specified, or is
+      # running on CI, in which case it will always be true.
+      readonly: options[:readonly] || is_ci,
```
jkmassel:
```suggestion
      readonly: true,
```

Nit: IMHO it's preferable to be extra-extra-extra safe here to avoid weird code signing issues. Not a requirement tho

fastlane/env/user.env-example
jkmassel: > I don't think there's a good ROI in keeping the release process flexible so that an external contributor can run it.

I agree with this – using the file path in `.configure-files` is probably best.

jkmassel (approved):
This LGTM.

One thing I'd like to note – whenever we end up calling `build_and_upload_app_store`, we should ensure that we're persisting the build artifacts regardless of success/failure. That way if the App Store Connect succeeds but Sentry symbol upload doesn't, we can just upload the symbols (for example).

fastlane/Fastfile
mokagio: Right! I remember we discussed about this... I just copy pasted from elsewhere and forgot to updated 😳

fastlane/Fastfile
mokagio: Good point. I'll update in a followup PR. 

The only thing I'll say is that is seems wasteful to call it on every lane if when not in CI. But... it's a minor waste and if we think about the future where "everything happens in CI" then it makes sense.

PR #900, opened 2021-03-05
Fix: AuthTextField UI Glitch

jleandroperez:
### Fix
In this PR we're fixing a UI glitch causing the Authentication Text Fields not to properly display the grayed-out background, whenever their responder status changes.

cc @eshurakov (Thank youuu!!)

### Test
1. Fresh install the app
2. Verify that clicking over the Username (or) Password field causes the BG to change
3. Click over the **Password** field
4. Click over the **SignUp** button

- [x] Verify that the **Username** field is highlighted (and the background is grayed out)

### Release

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -13,7 +13,8 @@ abstract_target 'Automattic' do
   # Automattic Shared
   #
   pod 'Automattic-Tracks-iOS', '0.8.0'
-  pod 'Simperium-OSX', '1.6.0'
+  # pod 'Simperium-OSX', '1.6.0'
+  pod 'Simperium-OSX', :git => "https://github.com/Simperium/simperium-ios.git", :commit => "94f7c26"
 
   # Main Target
   #
--- a/Podfile
+++ b/Podfile
@@ -14,7 +14,7 @@ abstract_target 'Automattic' do
   #
   pod 'Automattic-Tracks-iOS', '0.8.0'
   # pod 'Simperium-OSX', '1.6.0'
-  pod 'Simperium-OSX', :git => "https://github.com/Simperium/simperium-ios.git", :commit => "94f7c26"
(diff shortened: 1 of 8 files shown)


eshurakov (approved):
Works great!

:shipit:

pachlava:
@jleandroperez I'm trying to test it this way:
1. Be logged out
2. Fill in the email
3. Focus the password edit
4. Click `Sign Up` ~> The focus should be set to email input, and input background should grey?

I'm testing this in latest 2.10 and also compare to 2.9, and see no difference (in both version behaviour looks correct). Can it be that PR also landed in 2.9?

jleandroperez:
@pachlava it should be in 2.10 only (just checked!!). 

With those steps, the Email input wasn't getting grayed out. Perhaps it's a macOS version thing? (running Big Sur on my end!!).

mokagio:
If it helps, this is what happens on my end on 2.10 on Cata~~vista~~lina (I refuse to upgrade till absolutely necessary 🙃)

https://user-images.githubusercontent.com/1218433/112076068-cd120980-8bcd-11eb-906b-89026a8c5dad.mov

pachlava:
@mokagio thanks for the vid! I have almost the same, except the selected text backlight colour is a bit different (I guess this is not what issue is about anyway).

@jleandroperez have Big Sur too... In any case, I think this is what should happen? (2.10 + macOS 11.2.3):

https://user-images.githubusercontent.com/73365754/112139229-33e9ef80-8bdb-11eb-9075-69ecf0586c63.mov

mokagio:
>  except the selected text backlight colour is a bit different (I guess this is not what issue is about anyway).

Correct, that's just the highlight color settings at the macOS level. In System Preferences > General:

image

I change mine from time to time, just because I can 🤷

jleandroperez:
@pachlava The original bug involved the background remaining White when switching to SignUp (even though the Email Field was selected).

Yes sir! that's the expected behavior. Thanks for the clips @mokagio !!

PR #902, opened 2021-03-08
Merge version 2.9 beta 1 into develop

mokagio:
This is the first step of the release cycle.

This PR is important because, once merged, a bot will pick up the new localized strings and upload them to [GlotPress](https://translate.wordpress.com/) for translation.


Scripts/extract_release_notes.rb
```diff
@@ -0,0 +1,80 @@
...
+#
+# ruby ./this_script >| Simplenote/Resources/release_notes.txt
+#
+# To generate the App Store Connect release message:
+#
+# ruby ./this_script | pbcopy
+#
+# To generate the GitHub and App Center release message:
+#
+# ruby ./this_script -k | pbcopy
```
mokagio, line 14: See comment here https://github.com/Automattic/simplenote-ios/pull/1192#discussion_r589120626

RELEASE-NOTES.txt
```diff
@@ -1,7 +1,10 @@
+2.10
+-----
```
mokagio, line 2: Interestingly, this file was updated to use 3.0 (as per "WordPress versioning") but the next milestone was created as 2.10 🤔 

2.10 is the version I'd expect, to be be consistent with what the other Simplenote apps (albeit inconsistent with the other apps in the division 😅).

loremattei (approved):
LGTM!

PR #912, opened 2021-03-10
Merge finalized release 2.9 into develop

mokagio:
Doing this earlier to re-enable the signup flow. Internal ref. paaHJt-1Sn-p2.

### DIFF
--- a/Simplenote/SignupVerificationViewController.swift
+++ b/Simplenote/SignupVerificationViewController.swift
@@ -99,6 +99,15 @@ extension SignupVerificationViewController {
     func backWasPressed(_ sender: Any) {
         presentAuthenticationInteface()
     }
+
+    @IBAction
+    func contactWasPressed(_ sender: Any) {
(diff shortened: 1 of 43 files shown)


AliSoftware (approved):
LGTM, though not sure why the files `fastlane/metadata/it/release_notes.txt` and `fastlane/metadata/pt-BR/release_notes.txt` got deleted in that PR (I mean, I think our scripts delete those files at some point once the release notes `.po` file is generated from them… but what I mean is I'm not sure why those 2 were still around and the only one remaining, and were not deleted at the same time as the ones for the other locales were)

Approving nonetheless, but just curious if you know what happened here.

mokagio:
> Approving nonetheless, but just curious if you know what happened here.

Good spotting. I think what happened is the following. In the previous version those were the only translated release notes files created. In this version, none of them were translated (that's not an issue because the app is not localized on the App Store yet *). So, when the release finalization run, it deleted all the previous version but didn't update any because none were available.

`*` I'll need to followup on why that's the case. I suspect something fell through the cracks in my task management.

Issue #880, opened 2021-02-18
SearchBar: First Responder Bug


jleandroperez:
### Details
1. Click over the Search Bar twice

The Search Bar should preserve its first responder status. Instead, the second time you click on it, it'll resign the First Responder status.

jleandroperez:
Closed via #921

PR #921, opened 2021-03-26
Bugfix: SearchField + Double Click

jleandroperez:
### Description
In this PR we're fixing a bug that caused the SearchField not to properly accept double click events.

Thanks in advance Evgeny!!
cc @eshurakov 🪓 

### Details
1. We relied on an empty / inactive **NSToolbar** to increase the titlebar's height
2. Dropping such hack causes the Semaphore Buttons to get clipped (at the desired location)

As a workaround, we're:

A. Moving the buttons to the Window's contentView (to avoid clipping by its superview)
B. Updating **SplitViewController** (AKA "contentViewController), so that the contentView != splitView. This ensures the Semaphore buttons will always render above any SplitView artifacts.

### Test
- [ ] Verify the Window semaphore buttons look great
- [ ] Verify that the Window semaphore buttons also lookg great in Fullscreen 
- [ ] Verify double clicking over the Search Bar highlight the Search Text
- [ ] Verify that the SplitView's dividers **always show up below** the Semaphore Buttons

### Release
`RELEASE-NOTES.txt` was updated in 129c5bc with:

> Fixed a bug that caused double click over the SearchBar to fail #880


eshurakov (commented):
Looks good overall!

I checked on Catalina and noticed that mouse-over area for a semaphore is off.

https://user-images.githubusercontent.com/54751/112826058-caf7f100-908c-11eb-91d7-266709ba2a19.mov

jleandroperez:
Nice catch @eshurakov !!!

Should be fixed now, may I trouble you again? **Thank you!!**

eshurakov (approved):
Looks good!
:shipit:

jleandroperez:
Thank you Evgenyyyy!!

PR #927, opened 2021-04-01
Add logic to trigger App Store & GitHub upload via CircleCI APIs

mokagio:
### Fix
Add the Fastlane and CircleCI machinery to trigger a deployment to the App Store via the CircleCI APIs from Fastlane, and updates the `finalize_release` lane to trigger such a deployment.

### Test

1. Checkout this branch
2. Manually bump the build number
3. Push the change
3. Run the `trigger_app_store_ci_build branch_to_build:trigger-app-store-deploy-via-api`

You should see a build starting on CircleCI. ~It should fail because there's already a version with that version on App Store connect.~ The reason I asked to manually bump the build number is that, unlike for TestFlight in iOS, this build won't fail if you upload a binary with a version+build that already exists. It will finish happily, then spam the rest of the team with an email about the duplicated build number 😅 .

E.g. running the lane from cae6ce1 resulted in [this job](https://app.circleci.com/pipelines/github/Automattic/simplenote-macos/2303/workflows/fbc59f6f-01d7-42c6-a492-9bdd69e1bb7b/jobs/2572) starting in CI

image

After it finished, the build could be found on App Store Connect

image

I don't think there's much value into hacking the `code_freeze` lane to verify it calls this new lane instead of creating a tag and pushing it. Looking at the code should be enough.

### Review

### Release

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,10 +59,6 @@ jobs:
     steps:
       - setup_environment
       - run:
-          name: Decrypt Secrets
-          command: |
-            bundle exec fastlane run configure_apply
-      - run:
           name: Build and Upload to App Store and GitHub
           command: |
             bundle exec fastlane build_and_upload_app_store create_github_release:true
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,6 +4,11 @@ orbs:
   # This uses the Orbs located at https://github.com/wordpress-mobile/circleci-orbs
   ios: wordpress-mobile/ios@1.0
 
+parameters:
+  app_store_build:
+    type: boolean
+    default: false
+
 # Reusable sets of steps
 commands:
   copy_demo_credentials:
@@ -49,8 +54,31 @@ jobs:
           name: Verify App Store Target Builds
           command: bundle exec fastlane test_app_store_build
 
+  App Store Upload:
+    executor: *xcode_image
+    steps:
+      - setup_environment
+      - run:
+          name: Decrypt Secrets
+          command: |
+            bundle exec fastlane run configure_apply
+      - run:
+          name: Build and Upload to App Store and GitHub
+          command: |
+            bundle exec fastlane build_and_upload_app_store create_github_release:true
+
 workflows:
   simplenote_macos:
+    when:
+      not: << pipeline.parameters.app_store_build >>
     jobs:
       - Test
       - Verify App Store Target Builds
+
+  app_store:
+    when: << pipeline.parameters.app_store_build >>
+    jobs:
+      - Test
+      - App Store Upload:
+          requires:
+            - Test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,6 +59,9 @@ jobs:
     steps:
       - setup_environment
       - run:
+          name: Install Sentry CLI
+          command: brew install getsentry/tools/sentry-cli
+      - run:
           name: Build and Upload to App Store and GitHub
           command: |
             bundle exec fastlane build_and_upload_app_store create_github_release:true
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,9 +1,9 @@
 GIT
   remote: https://github.com/wordpress-mobile/release-toolkit
-  revision: 48f500bd9212f13d2bf71220c89a8794cdd46e03
-  tag: 0.13.0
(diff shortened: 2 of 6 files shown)


Gemfile.lock
```diff
@@ -1,9 +1,9 @@
 GIT
   remote: https://github.com/wordpress-mobile/release-toolkit
-  revision: 48f500bd9212f13d2bf71220c89a8794cdd46e03
-  tag: 0.13.0
+  revision: 348de22c625922cca9a5060261fa7a0c4005786e
+  tag: 0.16.0
```
mokagio, line 4: 0.16.0 is not the latest version. I'll make sure to update to the latest once the full CI automation work is done; I have other gems updates done in my [WIP branch](https://github.com/Automattic/simplenote-macos/tree/fix-internal-app-automation-wip).

fastlane/Fastfile
```diff
@@ -373,6 +383,25 @@ end
...
   end
 
+  #############################################################################
+  # trigger_app_store_ci_build
+  # ---------------------------------------------------------------------------
+  # This lane triggers a build for App Store distribution on CI
+  # ---------------------------------------------------------------------------
+  # Usage:
+  # bundle exec fastlane trigger_release_build [branch_to_build:<branch_name>]
+  #
```
mokagio: TODO! Remove this empty line

fastlane/Fastfile
```diff
@@ -373,6 +383,25 @@ end
...
+  #############################################################################
+  lane :trigger_app_store_ci_build do | options |
```
mokagio, line 391: I thought about making this a `private_lane`, but I think it's useful to be able to call it manually if we need to.

What do you think?

fastlane/Fastfile
```diff
@@ -278,6 +281,10 @@ end
   # bundle exec fastlane build_and_upload_app_store create_github_release:true
   #############################################################################
   lane :build_and_upload_app_store do | options |
+    configure_apply()
+    sh("rake dependencies:pod:clean")
+    cocoapods()
```
mokagio, line 287: I ported this from the lane that builds the beta version.

Originally, I had CI running `configure_apply`. This is a better approach because it moves more of the imperative set of steps to a CI-agnostic location.

Because CI already runs `pod install`, I wondered whether to keep the two instructions related to it. https://github.com/Automattic/simplenote-macos/blob/97fd3ea29ec300d00fffe1b84b5786aec7977c24/.circleci/config.yml#L17-L19

I value a CI-agnostic setup, but in the case of external dependencies it's actually valuable to explicitly install them in CI, so we can tap into its caching mechanism.

What do you think? If we decide to remove these, I'll do it in a dedicated PR to cover both production and beta lanes.

.circleci/config.yml
```diff
@@ -49,8 +54,30 @@ jobs:
...
+    when:
+      not: << pipeline.parameters.app_store_build >>
     jobs:
       - Test
       - Verify App Store Target Builds
+
+  app_store:
+    when: << pipeline.parameters.app_store_build >>
+    jobs:
+      - Test
```
mokagio, line 80: Running the tests at this step may be redundant in most cases. But, because we now trigger this via Fastlane locally, there is no hard guarantee that no new change made it into the codebase.

I think the extra peace of mind is a price worth paying the time delay.

fastlane/env/user.env-example
```diff
@@ -1,3 +1,7 @@
 GHHELPER_ACCESS=
 APPCENTER_API_TOKEN=
 SENTRY_AUTH_TOKEN=
+
+# Token to trigger builds via the CircleCI APIs
+# See: https://circleci.com/docs/2.0/managing-api-tokens/
+CIRCLE_CI_AUTH_TOKEN=<CircleCI Auth Token>
```
mokagio, line 7: Should this be:

```suggestion
CIRCLE_CI_AUTH_TOKEN=
```

instead?

mokagio:
Those [random timeouts after the tests succeed](https://app.circleci.com/pipelines/github/Automattic/simplenote-macos/2317/workflows/c24298de-903c-409d-8b0c-2eab63d6bdc4/jobs/2604) are killing me!

image

jkmassel (approved):
> Those random timeouts after the tests succeed are killing me!

This happened to me too 😅

But after retrying it worked great

PR #940, opened 2021-04-09
AuthViewController: SignUp by default

jleandroperez:
### Fix
In this PR we're adjusting the Authentication UI, so that the app starts off with the SignUp UI.

cc @eshurakov YO!!

### Test
1. Fresh install Simplenote

- [x] Verify there is no animation at launch
- [x] Verify the Authentication UI is, at launch, in SignUp mode
- [x] Verify that clicking over the **Log In** button switches the UI to Login 

### Release

### DIFF
--- a/Simplenote/AuthViewController+Swift.swift
+++ b/Simplenote/AuthViewController+Swift.swift
@@ -59,12 +59,12 @@ extension AuthViewController {
     @objc(refreshInterfaceWithAnimation:)
     func refreshInterface(animated: Bool) {
         clearAuthenticationError()
-        refreshButtonTitles(signingIn: signingIn)
-        refreshEnabledComponents(signingIn: signingIn)
-        refreshVisibleComponents(signingIn: signingIn, animated: animated)
+        refreshButtonTitles()
+        refreshEnabledComponents()
+        refreshVisibleComponents(animated: animated)
     }
 
-    func refreshButtonTitles(signingIn: Bool) {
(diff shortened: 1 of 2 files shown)


jleandroperez:
cc @eshurakov Ready for review sir, thanks in advance!!

Simplenote/AuthViewController+Swift.swift
```diff
@@ -79,37 +74,55 @@ extension AuthViewController {
         switchActionButton.title   = switchText.uppercased()
     }
 
-    /// Displays / Hides components, based on the ViewController Mode (SignUp / LogIn)
+    /// Makes sure unused components (in the current mode) are effectively disabled
     ///
-    func refreshVisibleComponentsWithAnimation() {
-        let mustHide = !signingIn
+    func refreshEnabledComponents(signingIn: Bool) {
```
eshurakov: `signingIn` is a property on `AuthViewController`, do we really need to pass it around? Will it happen that we will pass a value here that won't match `self.signingIn`? 🤔

eshurakov (approved):
Added a small question / comment.
Other than that, looks good! 
:shipit:

Simplenote/AuthViewController+Swift.swift
jleandroperez: Too much Erlang. You're right, it makes no sense to actually pass it around!

jleandroperez:
Thank you Evgeny!!

PR #960, opened 2021-05-13
Updated signup remote to lowercase emails before sending to simperium

### LINKED ISSUE #951: Signup: Emails should be lowercased before sending to the backend
### Expected

If a user enters an email using some capital letters, it should be converted to lowercase on form submission.

### Observed

The app sends a mixed-case email to the backend for registration.

Further context in pbMoDN-2oI-p2

charliescheer:
### Fix
This PR fixes #951 

This PR is almost identical to the changes made to SNiOS in https://github.com/Automattic/simplenote-ios/pull/1272 

This PR changes the case on email addresses before they are sent to the server during signup. Previously there was an issue where mixed case email addresses where causing accounts to be created immediately borked. Setting the emails to lowercase on signup is part of fixing this problem.

There isn't much to see in this PR as the email case is being set just before being sent to the server. As such to verify the changes I have created unit tests to confirm what is being sent. While I was there I also wrote tests for the rest of SignupRemote

### Test
1. run the unit tests for SNiOS
2. Check the tests on SignupRemoteTests.swift, they should all pass and confirm the correct casing of the email address.

### Review


jleandroperez (approved):
Thank you for doing this Charlie!!

:shipit:

PR #962, opened 2021-05-14
Merge finalized release 2.12 into trunk

mokagio:
We discussed not shipping 2.12 to the users because it doesn't include user facing changes. Because of that, this is not a Draft PR as it's custom for `trunk` release merges.


JavonDavis (approved):
Same feedback as #963 🙂

PR #963, opened 2021-05-14
Merge finalized release 2.12 into develop

mokagio:
On top of release finalization, includes the automated deployment changes from #927 and some `Deliverfile` tweaks.

_Note on the failing tests:_ This is an issue in `develop` (off which merge/release-2.12.into-develop` comes from) which @jleandroperez fixed in https://github.com/Automattic/simplenote-macos/pull/964.


config/Version.public.xcconfig
```diff
@@ -2,8 +2,8 @@ VERSION_SHORT=2.12
 
 // Public long version. Example: 2.0.0.0. The last 0 means the first build for
 // version 2.0.0
-VERSION_LONG=2.12.0.2
+VERSION_LONG=2.12.0.3
 
 // This is the value for the CFBundleVersion it should be incremented on every
 // build that gets distributed
-BUILD_NUMBER=11233
+BUILD_NUMBER=11236
```
mokagio, line 9: This non linear jump is due to the testing of #927.

fastlane/Deliverfile
```diff
@@ -7,18 +9,28 @@
...
+
+# We are using the API key to upload the metadata to ASC via Deliver, but, as
+# the tool reports:
+#
+# > [!] Precheck cannot check In-app purchases with the App Store Connect API
+# > Key (yet). Exclude In-app purchases from precheck, disable the precheck step
+# > in your build step, or use Apple ID login
+#
+# Simplenote iOS doesn't use IAPs, for the moment, so we can safely skip them.
+precheck_include_in_app_purchases false
```
mokagio, line 36: Ported this change from the other apps `Deliverfile` setup, but didn't really get a chance to verify it's useful because of the issue [above](https://github.com/Automattic/simplenote-macos/pull/963#r632775227).

fastlane/Deliverfile
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal:true
+
```
mokagio, line 2: There's a bunch of RuboCop styling fixes here. 

I tried to use `deliver` to go around the issue with App Store Connect rejecting locales because "the app name is already used in this account" but it didn't work out. That's not surprising, but it was worth a shot.

I thought it would be useful to track these changes regardless.

JavonDavis (approved):
I think we should go ahead and get #964 in now that it's approved but besides from that - changes LGTM!

mokagio:
The timeline is a bit confusing. For the record, @JavonDavis and I discussed this offline. Javon's comment regarding #964 landed just _after_ my commit to address it, accbe41. 😄

PR #968, opened 2021-05-19
Tweak tests and CI configuration to work around recent flakyness

mokagio:
Since around March 2021, we've been experiencing flakyness in the CI builds, in particular one of these two scenarios:

- The tests hang on `SignupRemoteTests`, usually before any of its tests run
- The tests succeed, but CircleCI receives no output from `xcodebuild` afterwards

I don't have a solution to either, but this PR introduces a few of workarounds that I hope will help us get faster feedback and pinpoint the culprit.

- When running the tests in CI, skip what seems to be the flaky test
- Reduce the no output timeout in CI, so that if the tests hangs we won't waste 10 minutes waiting for them

---

I run the tests three times and they always passed:

image

### DIFF
--- a/Simplenote.xcodeproj/xcshareddata/xcschemes/Simplenote.xcscheme
+++ b/Simplenote.xcodeproj/xcshareddata/xcschemes/Simplenote.xcscheme
@@ -29,7 +29,8 @@
       shouldUseLaunchSchemeArgsEnv = "YES">
       <Testables>
          <TestableReference
-            skipped = "NO">
+            skipped = "NO"
+            testExecutionOrdering = "random">
             <BuildableReference
                BuildableIdentifier = "primary"
                BlueprintIdentifier = "B5CBB068241976230003C271"
--- a/Simplenote.xcodeproj/xcshareddata/xcschemes/Simplenote.xcscheme
+++ b/Simplenote.xcodeproj/xcshareddata/xcschemes/Simplenote.xcscheme
@@ -60,6 +60,13 @@
             ReferencedContainer = "container:Simplenote.xcodeproj">
          </BuildableReference>
       </BuildableProductRunnable>
+      <EnvironmentVariables>
+         <EnvironmentVariable
+            key = "CIRCLECI"
+            value = "$(CIRCLECI)"
+            isEnabled = "YES">
+         </EnvironmentVariable>
+      </EnvironmentVariables>
    </LaunchAction>
    <ProfileAction
       buildConfiguration = "Release"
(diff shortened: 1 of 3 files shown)


SimplenoteTests/SignupRemoteTests.swift
```diff
@@ -9,7 +9,18 @@ class SignupRemoteTests: XCTestCase {
...
+        // At some point in May 2021, we started experiencing CircleCI handing on this test case.
+        // After some crude experimentation, it seems that this test is the culprit. We're skipping
+        // it for the moment, while waiting for time to investigate better.
+        //
+        // - Failures examples:
+        //   https://app.circleci.com/pipelines/github/Automattic/simplenote-macos?branch=merge%2Frelease-2.12-into-develop
+        // - CI experimentation:
+        //   https://app.circleci.com/pipelines/github/Automattic/simplenote-macos?branch=explore-ci-failures
+        let isCI = (ProcessInfo.processInfo.environment["CIRCLECI"] ?? "").isEmpty == false
+        try XCTSkipIf(isCI, "Skipped because it seems to be leading to timeous in CI")
```
mokagio, line 19: On CircleCI, where `CIRCLECI` is `true` the test is skipped:

image

But locally, it isn't 

via Fastlane | via Xcode
--- | ---
`bundle exec fastlane test` image | image


_Note: I took those screenshot when there was only one test skipped, but [this build](https://app.circleci.com/pipelines/github/Automattic/simplenote-macos/2369/workflows/1745e407-0282-4612-8082-1a2465ce2397/jobs/2754) showed CI hanging on the other status code test which is now skipped, too._

Simplenote.xcodeproj/xcshareddata/xcschemes/Simplenote.xcscheme
```diff
@@ -29,7 +29,8 @@
       shouldUseLaunchSchemeArgsEnv = "YES">
       <Testables>
          <TestableReference
-            skipped = "NO">
+            skipped = "NO"
+            testExecutionOrdering = "random">
```
mokagio, line 33: This should help show whether the hanging we're experiencing in `SignupRemoteTests` is due to the test case itself or is perhaps something that happens after n seconds of the tests launch.

If it's due to the test case, it will keep hanging when running it, regardless of its position in the randomly ordered list of tests. Otherwise, we should expect the tests to hang on random test cases.

eshurakov: Should we try only with this option first, without skipping a test?

jleandroperez (approved):
+1'ing Evgeny's comment (maybe the random test execution flag would help).

Other than that, **thank you** for doing this!!

Another thing we might just consider is... dropping the troublesome tests.
But that being said, if we do so, I'd very much love to know why they're breaking. Possibly related to us subclassing an SDK networking class (and some unexpected behavior in the CI)

:shipit: on my end sir!

Simplenote.xcodeproj/xcshareddata/xcschemes/Simplenote.xcscheme
mokagio: Great idea, thanks!

Done in #970.

mokagio:
Closing in favor of #970.

PR #965, opened 2021-05-18
Merge release 2.13 beta 1 into develop

mokagio:
This is the first step of the release cycle.

This PR is important because, once merged, a bot will pick up the new localized strings and upload them to [GlotPress](https://translate.wordpress.com/) for translation.

PR #974, opened 2021-05-28
Merge finalized release 2.13 into develop

mokagio:


mokagio:
Screen Shot 2021-05-28 at 11 35 07 am

This test failure is not a failure, actually, is just that CircleCI cancelled the build when it started the app store release build on the same branch.

I'll rerun once that finishes to get it to green.

Given CircleCI won't stay with us for long, I don't think this is worth addressing tooling-wise.

fastlane/metadata/it/release_notes.txt
```diff
@@ -1 +0,0 @@
-- Corretto un bug che impediva lo spostamento della finestra dell'app
```
mokagio, line 1: Notice there's other three _deleted_ release notes locales.

That would be a problem in other apps because we wouldn't have localizations on ASC. Here, though, we have an issue that prevents store localization on the ASC end, so it doesn't matter whether or not we have up to date release notes 😞

mokagio (commented):
Notice that there's no updated localizations. That's okay though, because no new string went into 2.13 (see https://github.com/Automattic/simplenote-macos/pull/965).

fastlane/metadata/it/release_notes.txt
AliSoftware: 😒 😢 
Thanks for documenting this in your release scenario update and in p2XJRt-2Ua-p2 in the meantime 👍

AliSoftware (approved):
Unfortunate thing about the release notes 😒 
Rest LGTM. :shipit:

PR #987, opened 2021-06-29
Scrollable Tags Editor

### LINKED ISSUE #985: Tags Editor: Embed in TextView Scroll
### Details:
Replicate iOS Tags Editor behavior:

- [x] Drop top divider between Note Editor / Tags Editor
- [x] Scroll the Tags Editor along with the note contents
- [x] Ensure there's padding between Content / Tags
- [x] The Tags Editor must stay anchored at the bottom, whenever the Note is not large enough
- [x] Verify: Should we kill the horizontal scroller?
- [x] Verify: Metrics
- [x] Verify: Should we automatically focus the Tags Editor?

jleandroperez:
### Details:
In this PR we're replicating iOS's Tags Editor behavior:

1. From now on, the Tags Editor will live below the Note Editor Area
2. Whenever the document is short, the Editor will be anchored at the bottom edge of the screen
3. Whenever the document is long enough, the Editor will scroll along with the Note's contents,
 
Closes #985

@eshurakov Sir! May I bug you with this one as well?
**Thanks in advance!!**

### Test: New Note
1. Add a new Note

- [x] Verify the Top Padding of the first line of text (Editor) matches the Notes List's first cell
- [x] Verify the Tags Editor is anchored at the bottom
- [x] Verify that you can click over the Tags Editor
- [x] Verify that if you add enough content, the Tags Editor scrolls along horizontally
- [x] Verify that you can remove or edit Tags

### Test: Existing Short Note
1. Click over any super short note

- [x] Verify the Tags Editor is anchored at the bottom of the Editor

### Test: Existing Long Note
1. Click over any super long note

- [x] Verify the Tags Editor is not immediately visible
- [x] Verify that if you scroll down, the Editor is revealed

### Test: Markdown Preview
1. Enable Markdown Preview in a note
2. While in Edition mode, click over the Tags Editor
3. Press the Markdown Preview button

- [x] Verify you can no longer edit the Tags (while in MD preview!)

### Test: UX
- [x] Verify the Tags Editor placeholder reads as `Tag...` (below the content editor!)
- [x] Verify that in a note with tons of Tags, the Tags Editor no longer clips its contents. [Ref. this comment!](https://github.com/Automattic/simplenote-macos/pull/987#issuecomment-872295078)
- [x] Verify that in a note with lots of text, pressing **CMD + Shift + Y** causes the Editor to scroll all the way down (and thus, making the Tags Editor visible)

### Release
`RELEASE-NOTES.txt` was updated in 7c31128 with:

> Tags Editor is now scrollable, making more room for the editor when appropriate

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,7 @@
 2.14
 -----
 - New Status Bar: Lets you quickly know what's your Tags / Notes selection #979
+- Tags Editor is now scrollable, making more room for the editor when appropriate #987
 
 2.13
 -----
--- a/Simplenote/HorizontalScrollView.swift
+++ b/Simplenote/HorizontalScrollView.swift
@@ -40,4 +40,11 @@ class HorizontalScrollView: NSScrollView {
         contentView.frame = bounds
         horizontalScroller?.isHidden = true
     }
+
+    /// Scrolls to the left-most edge
+    ///
+    func resetScrollPosition() {
+        let target = NSPoint(x: contentView.contentInsets.left * -1, y: .zero)
+        contentView.setBoundsOrigin(target)
+    }
 }
--- a/Simplenote/Main.storyboard
+++ b/Simplenote/Main.storyboard
@@ -646,15 +646,6 @@
                             <customView translatesAutoresizingMaskIntoConstraints="NO" id="Zgz-qj-sTm" userLabel="Tags View">
                                 <rect key="frame" x="0.0" y="0.0" width="528" height="48"/>
                                 <subviews>
-                                    <customView translatesAutoresizingMaskIntoConstraints="NO" id="K6m-2B-IQ0" userLabel="Bottom Divider View" customClass="BackgroundView" customModule="Simplenote" customModuleProvider="target">
-                                        <rect key="frame" x="0.0" y="47" width="528" height="1"/>
-                                        <constraints>
-                                            <constraint firstAttribute="height" constant="1" id="6Tb-0t-bpO"/>
-                                        </constraints>
-                                        <userDefinedRuntimeAttributes>
-                                            <userDefinedRuntimeAttribute type="boolean" keyPath="drawsTopBorder" value="YES"/>
-                                        </userDefinedRuntimeAttributes>
-                                    </customView>
                                     <scrollView placeholderIntrinsicWidth="462" placeholderIntrinsicHeight="25" borderType="none" horizontalLineScroll="10" horizontalPageScroll="10" verticalLineScroll="10" verticalPageScroll="10" hasVerticalScroller="NO" usesPredominantAxisScrolling="NO" verticalScrollElasticity="none" translatesAutoresizingMaskIntoConstraints="NO" id="FMH-MU-LHX" customClass="HorizontalScrollView" customModule="Simplenote" customModuleProvider="target">
                                         <rect key="frame" x="23" y="13" width="482" height="25"/>
                                         <clipView key="contentView" drawsBackground="NO" copiesOnScroll="NO" id="dTq-DR-Qae">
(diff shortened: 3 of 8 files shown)


eshurakov (approved):
Looks good!

Things I noticed while testing:
1. Unexpected new tag created
* Add `tag` tag to the note
* Double click on it to edit
* Type 'a', new tag `taga` is immediately created without any confirmation. 
 
2. Should we scroll to the tags editor when using a shortcut `cmd + shift + Y`? /cc @SylvesterWilmott

SylvesterWilmott:
Looks good and feels great to use.

As pointed out by @eshurakov when using the focus tags shortcut it would be good to scroll it into view and scroll back to the position of the cursor by pressing it again.

Are we able to eliminate the paddings on either side of the tags bar that causes then to be chopped off? See example below.

Screenshot 2021-07-01 at 15 26 13

Can we also keep the tag placeholder when there are tags present in the same way that we do on iOS? We can also modify that label to "Tag..." to match iOS.

pachlava:
One thing I've noticed so far is that tags are no longer shown/editable in Preview mode, putting it closer to the iOS experience. They were shown when there still was a separator (as happens now in Electron). Is this by design?

Fun addition: if the focus was in tags area before entering preview mode, it's still possible to add/remove tags while in preview mode, even though it's invisible:


https://user-images.githubusercontent.com/73365754/124160028-7ce7f480-daa4-11eb-96d7-213da7813d9c.mov

jleandroperez:
Thank you @eshurakov !! Figured out the trick to trigger that bug 10 out of 10. Filed #989 to track it down (happens in Appstore as well!).

jleandroperez:
### Update:

✅ Editing content is no longer possible when in Markdown Preview
✅ Tags Editor placeholder was updated
✅ Tags Editor no longer clips the bullets. We're using an inset instead
✅ We'll now always make sure the Tags editor is visible when `CMD + Shift + Y` happens

@SylvesterWilmott I'm afraid rendering the `Tag...` placeholder when there's already content is non standard, and will definitely be a time hog. I'll focus on the Tinting adjustments, and will get back to that if time allows (promise!!).

Gentleman, ready for another look. Test scenarios updated accordingly.
Thanks in advance!!

eshurakov (approved):
🎉

SylvesterWilmott:
Looks fantastic! I noticed one more tiny thing. In dark mode, the breadcrumb uses white for the currently selected text. Instead can we follow the text colour used else where which should be gray5.

SylvesterWilmott:
Looks fantastic! Thanks @jleandroperez 

>One thing I've noticed so far is that tags are no longer shown/editable in Preview mode, putting it closer to the iOS experience. They were shown when there still was a separator (as happens now in Electron). Is this by design?

This change was quite impromptu and in response to the inclusion of the breadcrumb bar. Eliminating the fixed tags area was something we discussed before but was never defined with a design. I wouldn't say that omitting the tags editor from the preview is "by design" but more of a product of the visual changes which neatly brings it in line with iOS and Android.

Since there is no event that allows me to differentiate between whether a tag was added in edit or preview mode, it's impossible to tell with usage data, if this change will impact how people tag their notes. However, as far as I know, we don't receive any requests to be able to add tags while in preview mode on mobile and there are no mentions of it on our ideas P2. We can keep an eye on what kind of feedback we get with this change.

jleandroperez:
Thanks everyone!!

pachlava:
I have noticed this thing (scroll in the tags bar):

https://user-images.githubusercontent.com/73365754/124444719-9086aa00-dd87-11eb-8528-29914a84997f.mov

And wanted to know if this is known / under construction somewhere else, and does this need a separate issue? Thanks.
cc @jleandroperez

jleandroperez:
Hi there @pachlava !! That would be on purpose, we're not clipping the Tags anymore, so you can "over-scroll" and recreate that effect.

cc @SylvesterWilmott for double check!! (TY both!!)

jleandroperez:
### Quick update:
Discussed with @SylvesterWilmott!! we'll disable horizontal scrolling, unless it's actually required. Filed in #994

Thank you @pachlava and @SylvesterWilmott !!

PR #992, opened 2021-07-01
Bugfix: Focus Mode Launch + Status Bar

jleandroperez:
### Fix
In this PR we're fixing a bug that caused the Status Bar not to display the selected filter.

### Details:
This glitch is caused by `viewWillAppear` not being invoked when State Restoration kicks in, and the app launches directly into Focus Mode.

Workaround ^ N, ideas on how to better address this one would be more than welcome!

**Thank yoouuu**

cc @eshurakov 

### Test
1. Launch Simplenote
2. Enter Focus Mode **(Option + Command + F)**
3. Press **Command Q** to quit the app
4. Relaunch Simplenote

- [x] Verify the Status Bar reads as `All Notes > [TITLE]`

### Release

### DIFF
--- a/Simplenote/TagListViewController+Swift.swift
+++ b/Simplenote/TagListViewController+Swift.swift
@@ -118,8 +118,24 @@ extension TagListViewController {
     }
 
     @objc
+    func startListeningToLaunchNotifications() {
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(finishedLaunching),
+                                               name: NSApplication.didFinishLaunchingNotification,
+                                               object: nil)
+    }
+
+    @objc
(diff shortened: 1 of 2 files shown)


eshurakov (approved):
Works as advertised!

---

> This glitch is caused by viewWillAppear not being invoked when State Restoration kicks in, and the app launches directly into Focus Mode.
> Workaround ^ N, ideas on how to better address this one would be more than welcome!

I think it's fine :-) 

One idea for future refactoring is to keep app state separate from view controllers. Interested parties can then observe the current state and all the changes.

jleandroperez:
Thank you sir!!!

PR #993, opened 2021-07-02
Toolbar Action: Insert Checklist

jleandroperez:
### Details
In this PR we're implementing a new Toolbar action, which allows us to insert Checklists.

@eshurakov sir may I bug you with yet another PR? (Thanks in advance!!)

cc @SylvesterWilmott @elibud 

### Test
- [x] Verify that the **Insert Checklist** button is enabled only when the Editor is the First Responder
- [x] Verify that the **Insert Checklist** button is hidden when you're visualizing the Trash filter
- [x] Verify that the **Insert Checklist** button is hidden when you select Multiple Notes
- [x] Verify that clicking over the new Checklist Toolbar Button causes the Editor to effectively insert a Checklist

### Release
`RELEASE-NOTES.txt` was updated in 5177592 with:

> You can now insert Checklists right from the Editor's Toolbar #993


eshurakov (approved):
Nice!

Q: should the button be disabled when we switch to markdown preview?

jleandroperez:
Done! thank you sir!

pachlava:
I did a bit of testing, a bit late, after this has already been merged, and noticed one tiny inconsistency: when the focus is inside note tags, the `Checklists` button is disabled, but it's still enabled in the format menu (shortcut also works). Please let me know if I should create an issue for that.

Also (just asking): wdyt about making this button have the `Insert Checklist` tooltip?

SylvesterWilmott:
>when the focus is inside note tags, the Checklists button is disabled, but it's still enabled in the format menu (shortcut also works)

Hmm, in the build @jleandroperez sent me directly, this wasn't the case (see below). But the shortcut and format menu item should indeed be disabled if the editor is in focus.

Screenshot 2021-07-05 at 13 04 56

>wdyt about making this button have the Insert Checklist tooltip?

I think this is a good suggestion. @jleandroperez can we update the tooltip string?

jleandroperez:
Sorry I'm late!! Filed in #995. TY @pachlava and @SylvesterWilmott !!

PR #1001, opened 2021-07-25
Refactor Remotes to use a Remote super class

charliescheer:
### Fix
This PR is laying the ground work for in app account deletion. 

What I have done in this PR is refactor the existing remote classes, AccountVerificationRemote and SignupRemote, to use a Remote super class.

Ultimately this has not changed any of how the remotes work, it has just removed much of the repeat code into it's own class in the same way that has been done for iOS, and preparing to change the AccountVerificationRemote into the AccountRemote which handles account verification and account deletion

### Test
1. Run the unit tests and make sure all test cases pass
2. Attempt an account signup
3. attempt an email verification.

### Review


mokagio:
@charliescheer the [error from CI](https://app.circleci.com/pipelines/github/Automattic/simplenote-macos/2446/workflows/b3e6bffa-0d7a-4ba0-8ddb-6cd63192bc26/jobs/2937?invite=true#step-111-1728) is the following:

```
❌  /Users/distiller/project/Simplenote/AccountVerificationController.swift:74:50: cannot find type 'Remote' in scope

    func verify(completion: @escaping (_ result: Remote.Result) -> Void) {
                                                 ^~~~~~
```

There are actually more errors than that one, but that's the first. From what I can see, they're all related.

## How to find errors in the Fastlane logs

Fastlane's output is a bit confusing, or rather, is very verbose so one might have to scroll up a bit on the CI logs before finding the actual failure reason.

In this instance, at the bottom of the log, Fastlane gives us some info on its latest changes:

Screen Shot 2021-07-26 at 1 04 02 pm

That's not the actual error, as the green color used sort of suggests.

Scrolling up a bit, we can find the Fastlane error output, which explains Fastlane "ran into a build/archive error":

image

The error is a build error: `error: cannot find 'performDataTask' in scope`.

Scrolling up a bit more, we can find the full `xcodebuild` output with all the errors:

image

## Why did this happen?

For historical reasons, Simplenote macOS has two different targets for internal build and App Store builds. This PR adds new files to the internal build target (the default one) but not to the App Store one.

Having multiple targets is inconvenient because it opens the door to this kind of errors. You're not the first one to not notice a new file belongs to only one of the targets and you won't be the last either. We've all been there 👍 

In fact, the whole point we have that CI check that failed here is to help us notice when we forget to add files to the App Store target sooner rather than later.

A while back, we discussed merging the two targets and achieving the different build behaviors via build configurations, preprocessor macros, and `.xcconfig`. See #853. Unfortunately, we haven't had the time to do it yet 🙁

## How to fix it

In case you never bumped into this, the way to fix this issue is to add the new files to the App Store target, too:

image

charliescheer:
Thanks! @mokagio I think this was the first file I added to SNMac, good to know about the app store target. Adding Remote.swift to the target fixed the issue

Simplenote/AccountVerificationRemote.swift
```diff
@@ -2,34 +2,16 @@ import Foundation
...
-
-    init(urlSession: URLSession = URLSession.shared) {
-        self.urlSession = urlSession
-    }
-
+class AccountVerificationRemote: Remote {
     /// Send verification request for specified email address
     ///
-    func verify(email: String, completion: @escaping (_ success: Bool) -> Void) {
+    func verify(email: String, completion: @escaping (_ result: Result) -> Void) {
```
jleandroperez: Can we switch to [Swift's Result Type](https://developer.apple.com/documentation/swift/result)?

jleandroperez (commented):
Nipticky request!! It'd be super cool if we could wire the standard Result type 🔥

Simplenote/AccountVerificationRemote.swift
charliescheer: Totally.  I have gone ahead and setup to use the Swift Result instead of the custom one built into remote.

I haven't used that before, so check it out and let me know if there are any changes that need to be made. @jleandroperez

Simplenote/AccountVerificationController.swift
```diff
@@ -71,7 +71,7 @@ class AccountVerificationController: NSObject {
 
     /// Send verification request
     ///
-    func verify(completion: @escaping (_ result: Result<Int, RemoteError>) -> Void) {
+    func verify(completion: @escaping (_ result: Result<Data?, RemoteError>) -> Void) {
```
mokagio, line 74: _Just checking, sorry if this is obvious, I don't have context around the code._

Seeing an `Optional` wrapped in a `Result` seems a bit odd to me. Is the case `.success(.none)`, that is a "successful" operation with no `Data` returned a legit state of the system? If it isn't, then we might want to use `Data` instead and return `.failure(<some error>)` if `Data` is `nil`.

PR #1003, opened 2021-07-26
Removed uppercase from h5/h6 in markdown preview

### LINKED ISSUE #1002: Markdown Preview H5/H6 converted to uppercase
This issue mirrors the iOS issue https://github.com/Automattic/simplenote-ios/issues/1304

Currently the CSS for H5 and  H6 in markdown preview is set to automatically capitalize all characters.  Per the discussion in that issue, we decided to remove that change.

### Expected
H5 and H6 would have normal casing

### Observed
H5 and H6 are uppercased

### Reproduced
1. in a note with markdown enabled add some text that is prefixed with 5 or 6 hashtags eg : `##### Heading 5##### `
2. Check the markdown preview and it will show capitalized letters

charliescheer:
### Fix
This PR fixes #1002  

Currently H5 and H6 headings in markdown preview are forced to be uppercase.  We decided to remove that uppercase.

Also this PR reorganizes the css so that the colors are in their own file which will simplify updates to the css in the future

### Test
1. In a note with markdown preview, add some text that uses H5 and H6 
2. display the preview and make sure that the text is not uppercased
3. also check and make sure both dark mode and light mode are loading the correct CSS in preview

### Review

### DIFF
--- a/Simplenote/SPMarkdownParser.m
+++ b/Simplenote/SPMarkdownParser.m
@@ -78,7 +78,7 @@
 
 + (NSString *)cssPath:(BOOL)isDarkMode
 {
-    return isDarkMode ? @"markdown-dark.css" : @"markdown-default.css";
+    return isDarkMode ? @"markdown-dark.css" : @"markdown-light.css";
 }
 
 + (NSString *)htmlFooter
--- a/Simplenote/SPMarkdownParser.m
+++ b/Simplenote/SPMarkdownParser.m
@@ -63,14 +63,20 @@
     headerStart = [headerStart stringByAppendingString:[NSString stringWithFormat:colorCSS, textHexColor]];
 
     NSString *headerEnd = @"</style></head><body><div class=\"note-detail-markdown\"><div id=\"static_content\">";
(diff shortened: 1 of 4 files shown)


Simplenote/SPMarkdownParser.m
```diff
@@ -63,16 +63,22 @@ + (NSString *)htmlHeader
...
+    NSString *colors = [NSString stringWithContentsOfURL:[[NSBundle mainBundle] URLForResource: colorCssPath withExtension:nil]
+                                                 encoding:NSUTF8StringEncoding error:nil];
+
+    css = [css stringByAppendingString:colors];
     
     return [[headerStart stringByAppendingString:css] stringByAppendingString:headerEnd];
 }
 
-+ (NSString *)cssPathForDarkMode:(BOOL)isDarkMode
++ (NSString *)cssPath:(BOOL)isDarkMode
```
jleandroperez: Mind if we keep the old API name? (It's more descriptive).

When invoking `[self cssPath:YES]` it's not clear what the boolean stands for.

jleandroperez (approved):
One nipticky comment. Other than that, looks great!!

:shipit:

PR #1006, opened 2021-07-29
Track `Package.resolved` file

mokagio:
This file is useful to ensure all developers fetching the project from Git use the _resolved_ version of the Swift packages. This is the same concept as `Gemfile.lock`, `Podfile.lock`, `package-lock.json`, etc.

See discussion at https://github.com/woocommerce/woocommerce-ios/pull/4392#discussion_r648001032

### Test
I think the fact that CI builds the app and the test pass is proof enough that the change is harmless. I'm pretty sure anyone who built the app locally would have this file there already, and it's something that Xcode generates, not a custom file I added.

Still, maybe it's worth checking out the PR locally and see if it builds?

### Review

### Release


charliescheer (approved):
Checked out, and builds fine.  Thanks for making these changes @mokagio

PR #1007, opened 2021-07-29
Setup RuboCop

mokagio:
This PR sets up RuboCop to sanitize the Ruby files we have in this project.

On top of adding it to the `Gemfile` to ensure every user adopts the same version, it also run `Gemfile` and `Podfile` through its autocorrect command. I wanted to run `Fastfile` through it, too, but decided to do it in a followup PR, just in case @oguzkocer is or will make new changes to it. Having RuboCop linting is a minor devx improvement and doesn't need to happen _right now_.

### Test
Nothing to test. If CI can build and run the tests successfully, we'll have validation that all the machine made changes in this PR are okay.

### Review

### Release


Gemfile
```diff
@@ -1,10 +1,12 @@
 source 'https://rubygems.org'
-gem 'rake'
+
 gem 'cocoapods', '>= 1.10'
-gem 'xcpretty-travis-formatter'
-gem 'octokit', "~> 4.0"
 gem 'dotenv'
 gem 'fastlane', '~> 2'
+gem 'octokit', '~> 4.0'
+gem 'rake'
```
jkmassel: Could we put a version constraint on this while we're changing it?

And we probably don't need `xcpretty-travis-formatter` since we're not using Travis CI anymore, but at the very least we could version-constrain that too?

jkmassel (approved):
Approved pending version constraints

Gemfile
mokagio: Good point. I only looked at the shape of the diff, not it's actual content 😳

mokagio: > And we probably don't need `xcpretty-travis-formatter` since we're not using Travis CI anymore, but at the very least we could version-constrain that too?

I agree, we shouldn't need it. I just discovered, though, that it's used in the Rake tasks. I think removing it is out of scope for this PR. I opened https://github.com/Automattic/simplenote-macos/issues/1008 to track that.

https://github.com/Automattic/simplenote-macos/blob/f2e2c5bdc7e81e6e1bb6cdbbca82c2bd6198d87a/Rakefile#L289

PR #1011, opened 2021-08-02
Move Fastlane secrets outside of repository

mokagio:
Similar to https://github.com/wordpress-mobile/WordPress-iOS/pull/16780.

Moves the Fastlane-related secrets outside of the repository into `~/.configure/simplenote-macos/secrets/`

To test locally, checkout this branch, run `bundle exec run configure_apply` and then inspect the content of `~/.configure/simplenote-macos/secrets/`:

```
$ ls ~/.configure/simplenote-macos/secrets
app_store_connect_fastlane_api_key.json
google_cloud_keys.json
```

I also opened #999 to test these changes. It exercises each secret by:

1. Calling `match` for each build configuration, to test the `google_cloud_keys.json` location
2. Making a build and uploading it to TestFlight, to test the ASC API key JSON location. _On the iOS pipelines, this fails because App Store Connect rejects builds that reuse build numbers. But for macOS, this doesn't happen, meaning the builds [succeeds](https://app.circleci.com/pipelines/github/Automattic/simplenote-macos/2478/workflows/391e44f2-551b-4a35-b95e-729414afe8fb/jobs/3008) but you'll get an email saying there's an error in the binary._

### Review

### Release


.configure
```diff
@@ -10,12 +10,12 @@
...
     {
       "file": "shared/google_cloud_keys.json",
-      "destination": ".configure-files/google_cloud_keys.json",
+      "destination": "~/.configure/simplenote-macos/secrets/google_cloud_keys.json",
       "encrypt": true
     },
     {
       "file": "iOS/app_store_connect_fastlane_api_key.json",
-      "destination": ".configure-files/app_store_connect_fastlane_api_key.json",
+      "destination": "~/.configure/simplenote-macos/secrets/app_store_connect_fastlane_api_key.json",
```
mokagio, line 18: At some point in the future, we'll add some kind of `files_to_remove` annotation to clean up the devs' local repo. See also https://github.com/woocommerce/woocommerce-ios/pull/4576/files#r670733139.

fastlane/Pluginfile
```diff
@@ -2,6 +2,6 @@
 #
 # Ensure this file is checked in to source control!
 
-gem 'fastlane-plugin-wpmreleasetoolkit', git: 'https://github.com/wordpress-mobile/release-toolkit', tag: '0.18.1'
+gem 'fastlane-plugin-wpmreleasetoolkit', '~> 1.3'
```
mokagio, line 5: Now with SemVer 😄

fastlane/Matchfile
```diff
@@ -1,4 +1,7 @@
-# Store certs/profiles encrypted in Google Cloud
-storage_mode("google_cloud")
-google_cloud_bucket_name("a8c-fastlane-match")
-google_cloud_keys_file(".configure-files/google_cloud_keys.json")
```
mokagio, line 4: Sorry for the noise in the diff. While I was on this file, I converted Ruby `""` into `''`, which is the recommended syntax for strings that don't use interpolation.

AliSoftware:
> Making a build and uploading it to TestFlight, to test the ASC API key JSON location. On the iOS pipelines, this fails because App Store Connect rejects builds that reuse build numbers. But for macOS, this doesn't happen, meaning the builds succeeds but you'll get an email saying there's an error in the binary.

Ohhh that's what this email was about! Was about to post a message in our Slack to warn about the Apple rejection, but now I see that it came from here and that there's nothing to worry about 👍

.configure
AliSoftware: FYI, this idea is also tracked in https://github.com/Automattic/configure/issues/19

AliSoftware (approved):
Code looks good, and given you mentioned that you tested it in #999 already, I think it's ok to :shipit:

.configure
mokagio: Excellent, thanks!

PR #1012, opened 2021-08-03
Move swift secrets outside of the repository

mokagio:
Moves the `SPCredentials.swift` secret outside of the repository. Similar to what we did in https://github.com/wordpress-mobile/WordPress-iOS/pull/16806.

### To Test

1. Checkout this branch
2. Build the app, you should get a warning

Screen Shot 2021-08-04 at 11 28 01 pm

3. Run `bundle exec fastlane run configure_apply`
4. Verify that `SPCredentials.swift` is now in `~/.configure/simplenote-macos/secrets`
4. Build the app again, the warning should be gone

### Review

### Release

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,13 +11,6 @@ parameters:
 
 # Reusable sets of steps
 commands:
-  copy_demo_credentials:
-    steps:
-      - run:
-          name: Copy Demo SPCredentials
-          command: |
-            mkdir -p Simplenote/Credentials
-            cp Simplenote/SPCredentials-demo.swift Simplenote/Credentials/SPCredentials.swift
   setup_environment:
     steps:
       - checkout
@@ -40,7 +33,6 @@ jobs:
       xcode-version: "12.4.0"
     steps:
       - setup_environment
-      - copy_demo_credentials
       - run:
           name: Build & Test
           command: bundle exec fastlane test
@@ -58,7 +50,6 @@ jobs:
     executor: *xcode_image
     steps:
       - setup_environment
-      - copy_demo_credentials
       - run:
           name: Verify App Store Target Builds
           command: bundle exec fastlane test_app_store_build
--- a/Scripts/Build-Phases/copy-secret.sh
+++ b/Scripts/Build-Phases/copy-secret.sh
@@ -1,0 +1,38 @@
+#!/bin/bash -o pipefail
+
+SOURCE_PATH=$1
+TARGET_PATH=$2
+
+USAGE="error: Usage $0 source_path target_path"
+
+if [[ -z ${SOURCE_PATH:+x} || -z ${TARGET_PATH:+x} ]]; then
+  echo $USAGE
+  exit 1
+fi
+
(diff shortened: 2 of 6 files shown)


.circleci/config.yml
```diff
@@ -11,13 +11,6 @@ parameters:
 
 # Reusable sets of steps
 commands:
-  copy_demo_credentials:
-    steps:
-      - run:
-          name: Copy Demo SPCredentials
-          command: |
-            mkdir -p Simplenote/Credentials
-            cp Simplenote/SPCredentials-demo.swift Simplenote/Credentials/SPCredentials.swift
```
mokagio, line 20: This is unnecessary now, because the build phase script now takes care of falling back to the demo credentials.

Scripts/Build-Phases/copy-secret.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash -euo pipefail
```
mokagio, line 1: This is an adapted copy of the [WordPress iOS version](https://github.com/wordpress-mobile/WordPress-iOS/pull/16806).

fastlane/Fastfile
```diff
@@ -95,6 +95,7 @@ end
 
 desc 'Builds the app using the App Store target but without code signing, just to check the target builds'
 lane :test_app_store_build do | options |
+  configure_apply
```
mokagio, line 98: Above I wrote that coping the demo secrets in CI:

> is unnecessary now, because the build phase script now takes care of falling back to the demo credentials.

On the other hand, because this step builds the App Store target in Release mode and the script  fails if the secrets can't be found on a Release build.

jkmassel (approved):
This worked well, though I only got the error on the initial build – after that I had to clear my build directory to see it again. Not sure how I feel about this – on the one hand it's good to not spam the developer with the warning over and over, but on the other it might get missed? 

WDYT?

mokagio:
@jkmassel 

> This worked well, though I only got the error on the initial build

I assume "error" is the warning. If that's the case, then there's something wrong, as I agree with your statement:

> on the one hand it's good to not spam the developer with the warning over and over, but on the other it might get missed?

mokagio:
Actually, I just realized it might be because the build phase run script is "cached": Xcode will only run it on a clean build or if one of the input files has changed.

🤔 I'm going to merge this, then think of a solution. I think we may need to split this in two, adding a dedicated script that outputs the warning and make that not cached.

AliSoftware:
@mokagio Yes, this is because of the input files. Xcode doesn't re-run the Script Build Phase if the input files have not changed nor the output files already exists. Just like in a `Makefile` for example. This is all part of the logic of how incremental builds work to avoid rebuilding/running everything every time.

I wonder if just unchecking the "Based on dependency analysis" checkbox on the Script Build Phase could help disable that behavior?

_(One thing that can help here is by having this configured as an aggregate target that other targets depend on, instead of a Script Build Phase repeated on each app target. When this is an aggregate target you only have to edit the script phase settings at once place and ensure it's consistent)_

PR #1019, opened 2021-08-17
Adds account verification response to login flow

charliescheer:
### Fix
This PR adds a new error case while logging in where a user's email address has not been verified they will receive an error alerting them to the fact that they can not login till their email is verified and then supplies them with a button to resend the verification email.

### Test
1. setup a tool like Charles proxy to get in the way of the login request response.
2. When the response returns, change the response value to 403
3. You should see an alert with two options, verify and okay.

Note this will require more testing when the backend changes are made to return the 403 response to an unverified email, currently that has not been added in, so it can not be fully tested.

### Review


RELEASE-NOTES.txt
```diff
@@ -1,6 +1,7 @@
 2.15
 -----
 -   [Internal] added login alert for compromised password #1017
+- [Internal] added error response case on login for unverified emails #1019
```
jleandroperez, line 3:
```suggestion
-   [Internal] added error response case on login for unverified emails #1019
```

jleandroperez: (Nipticky!! so that it has the exact same spacing as the line before)

Simplenote/AuthViewController.m
```diff
@@ -353,6 +355,17 @@ -(void)presentPasswordCompromisedAlert
     }];
 }
 
+-(void) presentUnverifiedEmailAlert
```
jleandroperez:
```suggestion
- (void)presentUnverifiedEmailAlert
```

Simplenote/AuthViewController+Swift.swift
```diff
@@ -201,6 +201,22 @@ extension AuthViewController {
...
+        alert.informativeText = Localization.unverifiedInformativeText
+        alert.addButton(withTitle: Localization.unverifiedActionText)
+        alert.addButton(withTitle: Localization.unverifiedCancelText)
+
+        alert.beginSheetModal(for: window, completionHandler: completion)
+    }
+
+    @objc
+    func sendVerificationMessage(for email: String) {
+        AccountVerificationRemote().verify(email: email) { _ in }
```
jleandroperez: We should probably replicate what @eshurakov mentioned [here](https://github.com/Automattic/simplenote-ios/pull/1398#discussion_r690185403). Pasting for convenience!!

```
I think we should show an activity indicator (can use the one we have for account deletion) and handle errors similar to what we're doing in AccountVerificationViewController
```

jleandroperez (commented):
Few notes!! most important one is the comment regarding Spinner / Error Handling.

TY!!

Simplenote/AuthViewController+Swift.swift
charliescheer: Good idea.  I have added the check to confirm that the 403 error is "requires verification" and added a ProgressIndicator

Simplenote/AuthViewController+Swift.swift
```diff
@@ -234,4 +264,12 @@ private enum Localization {
     static let compromisedPasswordMessage = NSLocalizedString("This password has appeared in a data breach, which puts your account at high risk of compromise. It is recommended that you change your password immediately.", comment: "Compromised password alert message")
     static let changePasswordAction = NSLocalizedString("Change Password", comment: "Change password action")
     static let dismissChangePasswordAction = NSLocalizedString("Not Now", comment: "Dismiss change password alert action")
+    static let unverifiedInformativeText = NSLocalizedString("You must verify your email before being able to login.", comment: "Erro for un verified email")
+    static let unverifiedMessageText = NSLocalizedString("Account Verification Required", comment: "Email verification required alert title")
+    static let cancelText = NSLocalizedString("Okay", comment: "Email unverified alert dismiss")
```
eshurakov: Similar comment as on iOS, we were always using "OK"

eshurakov (approved):
Looks good to me!

:shipit:

Simplenote/NSProgressIndicator+Simplenote.swift
```diff
@@ -0,0 +1,19 @@
+import Foundation
+
+extension NSProgressIndicator {
+    static func addProgressIndicator(to view: NSView) -> NSProgressIndicator {
+        let progressIndicator = NSProgressIndicator.init(frame: NSRect(x: view.frame.minX, y: view.frame.midY, width: 20, height: 20))
```
jleandroperez:
```suggestion
        let progressIndicator = NSProgressIndicator(frame: NSRect(x: view.frame.minX, y: view.frame.midY, width: 20, height: 20))
```

jleandroperez: Since you're using autolayout... can we try just running `NSProgressIndicator()` instead?

jleandroperez: Plus let's also add `progressIndicator.controlSize = .small`, looks way cooler!

Simplenote/AuthViewController+Swift.swift
```diff
@@ -202,6 +202,36 @@ extension AuthViewController {
...
+        alert.addButton(withTitle: Localization.unverifiedActionText)
+        alert.addButton(withTitle: Localization.cancelText)
+
+        alert.beginSheetModal(for: window, completionHandler: completion)
+    }
+
+    @objc
+    func sendVerificationMessage(for email: String, inWindow window: NSWindow) {
+        let progressIndicator = NSProgressIndicator.addProgressIndicator(to: view)
+
```
jleandroperez, line 221: Let's add `setInterfaceEnabled(false)` (and true on completion)

jleandroperez (approved):
Looks great!! few comments (re: Spinner size, locking the UI).

Other than that, :shipit: when ready!!

PR #1026, opened 2021-08-31
Updated copy for account deletion alert message to include user email

### LINKED ISSUE #1023: Mention the account email in the first account deletion alert
### What
macOS native does not show the currently logged in account. Because of this, when `Delete Account` is clicked, it's not clear what is the account email from the first alert:

Screenshot 2021-08-31 at 11 53 42

It _is_ shown in the next alert, the one saying that the email was sent to a specified email address, but I think it's good to have it appear in both places.

p5T066-2Co-p2#comment-9776

### Why
Users will know which account are they going to delete faster.

charliescheer:
### Fix
This is a quick update that fixes #1023 

Currently the alert message that appears when a user tries to delete their account does not include any details that identify their account.  This update adds their email address to the alert message for clarity

### Test
1. go to the menu Simplenote -> Delete Account
2. an alert will appear that should include the email address for the account being deleted

### Review

### Release
> `RELEASE-NOTES.txt` was updated in 9aa5fb with:
> 
> > - Updated the account deletion flow for clarity #1026

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,6 @@
 2.16
 -----
-
+- Updated the account deletion flow for clarity #1026
 
 2.15
 -----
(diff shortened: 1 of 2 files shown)


eshurakov (approved):
Looks good! :shipit:

PR #1031, opened 2021-09-07
Hotfix light mode window issue

### LINKED ISSUE #1029: Simplenote v2.15 doesn't open
Please, be as descriptive as possible.  Issues lacking detail, or for any other reason than to report a bug, may be closed without action.

### Expected
The app should open.

### Observed
The app starts (because I see that the menu bar changes to "Simplenote"), but the main window doesn't open. 

#### Screenshots

Screen Shot 2021-09-06 at 19 05 57
^ The above screenshot is what it looks like when the Simplenote app is opened. 

***

Screen Shot 2021-09-06 at 19 06 12
^ I am able to click on the Simplenote menu > "About Simplenote" to get this window to appear.

***

Screen Shot 2021-09-06 at 19 06 31
^ When I click on the Note menu > New the above blank window appears.

***

I previously had Simplenote installed on my system from the App Store. I deleted the app and tried reinstalling from the App Store. This didn't fix the issue. (I even tried again (deleting and reinstalling) using the app "App Cleaner" to no avail.)

Make|Model|macOS Version|App Version
-|-|-|-
Apple|MacBook Pro 16-inch 2019|macOS v11.5.2|v2.15

charliescheer:
### Fix
The most recent version of Simplenote Mac has an issue where if the theme is set to light mode the window will not load when the app is opened.  This PR fixes that issue

### Test
1.  Set simplenote to load in the light theme (if set to use the system default set your device to light, otherwise go to View -> Theme -> Light)
2. run the app
3. you should see the window open as normal

### Review
> Only one developer iss required to review these changes, but anyone can perform the review.

### Release
> `RELEASE-NOTES.txt` was updated in cd1a67 with:
> 
> > - Fixed a bug that caused the app window not to load in light mode #1031

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -2,6 +2,11 @@
 -----
 
 
+2.15.1
+-----
+- Fixed a bug that caused the app window not to load in light mode #1031
+
+
 2.15
 -----
 - [Internal] added error response case on login for unverified emails #1019
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,12 +1,13 @@
 2.16
 -----
-- Updated the account deletion flow for clarity #1026
+
 
 2.15
 -----
 - [Internal] added error response case on login for unverified emails #1019
 - In app account deletion: Now you can delete your account from the app #1013
 - [Internal] added login alert for compromised password #1017
+- Updated the account deletion flow for clarity #1026
 
 2.14
 -----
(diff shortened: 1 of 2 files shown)


mokagio (approved):
Tested on `develop` and 2.15 App Store build to reproduce the issue, then verified that it doesn't occur on this branch.

I also run the app in dark mode, just in case.

I'm going to take over merging and releasing this. Thanks for the quick turnaround @charliescheer 👏

charliescheer:
@mokagio Thank you for all of the help!

eshurakov:
@charliescheer, @mokagio, thank you for taking care of this! 👍

Issue #1037, opened 2021-09-20
MacOS Preferences Window


charliescheer:
This is a project issue that covers adding a Preferences menu to Simplenote MacOS.  Internal ref: p2XJRt-3hF-p2

Design:
<img src="https://cdn-std.droplr.net/files/acc_593859/sAUCDt" />

<img src="https://cdn-std.droplr.net/files/acc_593859/MYzs7v" />

Tasks:

- [ ] UI
- [ ] Remove existing menu controls for preferences

charliescheer:
This is finished

PR #1042, opened 2021-09-24
Preferences View Controller and UI

### LINKED ISSUE #1038: MacOS Preferences: PreferencesController
This is a project list for the PreferencesController part of the [MacOS Preferences Project](https://github.com/Automattic/simplenote-macos/issues/1037). 

To Dos:

- [ ] Account settings (email, logout, delete account)
- [ ] View Settings (Note Sort Order, Line Length, Condensed Note List, Alphabetical Tags)
- [ ] Theme
- [ ] Text Size
- [ ] Privacy

### LINKED ISSUE #1039: MacOS Preferences: UI
This is a project task list for the UI part of the [MacOS Preferences Project ](https://github.com/Automattic/simplenote-macos/issues/1038)

Tasks:

- [x] Main layout
- [x] connect to settings button, menu settings, keyboard shortcut
- [x] connect to preferences view controller so changes in view reflect in app
- [ ] change how UI elements are created so their style can be more easily changed
- [ ] remove old menu items that are replaced by the preferences view

charliescheer:
### Fix
This is the first PR in the process of adding the Preferences window to Simplenote Mac (over all issue #1037, fixes #1038 )

In this PR I have added all of the UI for the preferences view and linked all of the controls into the window.

Note: one thing that is not done here is the button color for the Delete Account  button.  Still working on getting that going.

### Test
1. Test opening the preferences window by going to Simplenote > Preferences, hitting Command + , and clicking on the Settings button in the tag list
2. Make sure that you can only open one preferences view at a time (shouldn't be possible to have multiple windows of preferences appear)
3. Check and see if your account email appears in the top
4. Test log out button
5. log in with a testing account, test delete account button
6. change the sort order of your notes and make sure the note list updates
7. change the line length setting from full to narrow and back
8. select the condensed note list and deselect it
9. select the sort tags alphabetically and deselect it
10. try the different theme settings
11. use the font slider to change the font size
12. test to make sure these setting changes persist when you close the preferences window and when you restart the app
13. click on the "About Analytics and Privacy" link to be taken to the a8c privacy page

### Release

> `RELEASE-NOTES.txt` was updated in a4948377 with:
> 
> > Added preferences menu to centralize app options and customizations #1042


jleandroperez (commented):
@charliescheer sending you few UX notes, getting to the actual code in about a second!! Thank you!!

### Preferences Window
I've noticed the `Maximize` button is enabled. Clicking on it toggles, obviously!!, the Preferences UI into a fullscreen window.



### Theme Switch
1. Open the Preferences UI
2. Switch over to a different theme

Notice the Preferences UI doesn't reflect the newly selected Theme.

### Preferences Sync
1. Open the Preferences UI
2. While the Preferences UI is onscreen, update any of the settings, but from the App's Menu

Notice that the Preferences UI falls out of sync with the actual underlying setting.

### Menu
@SylvesterWilmott can we drop `Log Out` and `Delete Account` from the `Simplenote` macOS App Menu? (feels like it'd be okay to have those two _only_ in the Preferences UI).

charliescheer:
@jleandroperez Thanks for the feedback.

>I've noticed the Maximize button is enabled. Clicking on it toggles, obviously!!, the Preferences UI into a fullscreen window.

I've removed the minimize and maximize buttons

>Notice the Preferences UI doesn't reflect the newly selected Theme.

Oh nice catch! I fixed that.


We will be removing most of the menu links in another PR, so falling out of sync shouldn't be an issue.  We are keeping the font size changes, so will need to do something about making sure those are up to date, but the rest shouldn't be an issue

Simplenote/SimplenoteConstants.swift
```diff
@@ -31,4 +31,11 @@ class SimplenoteConstants: NSObject {
     static let simplenoteVerificationURL    = currentEngineBaseURL.appendingPathComponent("/account/verify-email/")
     static let simplenoteRequestSignupURL   = currentEngineBaseURL.appendingPathComponent("/account/request-signup")
     static let accountDeletionURL           = currentEngineBaseURL.appendingPathComponent("/account/request-delete/")
+
+    /// Font Constants
+    ///
+    static let NoteFontSizeMinimum = 10
+    static let NoteFontSizeNormal = 15
+    static let NoteFontSizeMaximum = 30
+    static let NoteFontSizeStep = 5
```
jleandroperez: Since all of these are prefixed as `NoteFontSize`, it would probably be a good idea to move it to its own enum / struct.

How about... `FontSettings`?  (Oh! another note, constants should not be capitalized in Swift) 😄

External/Notepad/Theme.swift
```diff
@@ -17,7 +17,7 @@ class Theme: NSObject {
 
     /// Default Font Size
     ///
-    private static let defaultFontSize = CGFloat(14)
+    private static let defaultFontSize = CGFloat(SimplenoteConstants.NoteFontSizeNormal)
```
jleandroperez: It'd probably be a good idea to drop this constant, since it's a constant of a constant.

Plus: noted a bit below, since there are a bunch of constants named `NoteFontSize`, we could probably group them into their own namespace. Perhaps `FontSettings.defaultFontSize` ? 

Or... `FontConstants`, but `FontSettings` somehow feels cooler?

External/Notepad/Theme.swift
```diff
@@ -96,11 +96,11 @@ private extension Theme {
 private extension Theme {
 
     private static var fontSize: CGFloat {
-        var fontSize = CGFloat(UserDefaults.standard.integer(forKey: "kFontSizePreferencesKey"))
+        var fontSize = CGFloat(Options.shared.fontSize)
```
jleandroperez: Q: Can we have the `fontSize` be a `CGFloat` directly, rather than converting here?

External/Notepad/Theme.swift
```diff
@@ -96,11 +96,11 @@ private extension Theme {
 private extension Theme {
 
     private static var fontSize: CGFloat {
```
jleandroperez, line 94: Perhaps this entire method could be taken away (and replaced by, directly, using fontSize, granted that we patch it up to return the default Size if nothing is set)

Simplenote/Int+Extensions.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
+
+extension Int {
+    func roundToNearestFive() -> Int {
+        return 5 * Int((Double(self) / 5.0).rounded())
+    }
```
jleandroperez: Since this API is *not* generic (it's very, very specific to just one use case), perhaps it could be better off as a function in `FontSettings` ? (or elsewhere?)

jleandroperez: Second idea would be to parametrize. `roundToNearest(multiple: Double)` (or something like that, but idea is to drop the constant)

Simplenote/NSImageName+Simplenote.swift
```diff
@@ -38,6 +38,10 @@ extension NSImage.Name {
     ///
     static let ellipsisOutline = NSImage.Name("button_ellipsis_outline")
 
+    /// Tags List: Settings
+    ///
+    static let settings = NSImage.Name("button_settings")
```
jleandroperez: Really Nipticky: constants in this file are sorted by name 😂 

Told you. Seriously nipticky!!

Simplenote/NoteEditorViewController+Swift.swift
```diff
@@ -926,6 +926,18 @@ extension NoteEditorViewController {
             noteEditor.scrollToSelectedLocation()
         }
     }
+
+    @objc
+    func getRoundedFontSize() -> Int {
```
jleandroperez: Can we pleaaaase drop the`get`? (since it's not swift-y). Some of the old code, though, has that, buuut we should 🔥

Plus: perhaps `FontSettings` would be a more appropriate location?

Simplenote/NoteEditorViewController+Swift.swift
```diff
@@ -926,6 +926,18 @@ extension NoteEditorViewController {
...
+        var currentFontSize = Options.shared.fontSize
+        let fontSizeIsNotValid = currentFontSize % SimplenoteConstants.NoteFontSizeStep != .zero
```
jleandroperez: Mind verifying if we **do** need this? it's possible that calling the rounding method, if it's already rounded, won't do anything.

Simplenote/PreferencesViewController.swift
```diff
@@ -0,0 +1,314 @@
...
+    @IBOutlet private var sortTagsAlphabeticallyCheckbox: NSButton!
+    @IBOutlet private var themePopUp: NSPopUpButton!
+    @IBOutlet private var textSizeSlider: NSSlider!
+    @IBOutlet private var shareAnalyticsCheckbox: NSButton!
+
+    // Mark: Background Views
+    @IBOutlet var accountSectionBackground: BackgroundView!
+    @IBOutlet var layoutSectionBackground: BackgroundView!
+    @IBOutlet var themeSectionBackground: BackgroundView!
+    @IBOutlet var textSectionBackground: BackgroundView!
```
jleandroperez: `private`

Simplenote/PreferencesViewController.swift
```diff
@@ -0,0 +1,314 @@
+import Cocoa
+
+class PreferencesViewController: NSViewController {
+    let simperium = SimplenoteAppDelegate.shared().simperium
```
jleandroperez: This one shouldn't be a constant (perhaps a dynamic property?).

We're not doing it, but in theory, the `Simperium` instance could be swapped, and we could be left there with an old reference.

Plus: `private`

Simplenote/PreferencesViewController.swift
```diff
@@ -0,0 +1,314 @@
...
+        condensedNoteListCheckbox.state = Options.shared.notesListCondensed ? .on : .off
+        sortTagsAlphabeticallyCheckbox.state = Options.shared.alphabeticallySortTags ? .on : .off
+
+        updateSelectedTheme()
+
+        textSizeSlider.intValue = Int32(Options.shared.fontSize)
+
+        shareAnalyticsCheckbox.state = Options.shared.analyticsEnabled ? .on: .off
+
+    }
```
jleandroperez, line 75: I've noticed some components have an `update` API, some don't. Would it be possible to make it all standard?

(Plus: extra newline!). TY!!

Simplenote/PreferencesViewController.swift
```diff
@@ -0,0 +1,314 @@
...
+
+    private func setupSortModeFields() {
+        let menuItems: [NSMenuItem] = SortMode.allCases.map { mode in
+            let item = NSMenuItem()
+            item.title = mode.description
+            item.identifier = mode.noteListInterfaceID
+            return item
+        }
+        
+        menuItems.forEach({ sortOrderPopUp.menu?.addItem($0) })
```
jleandroperez: Both `SortMode.allCases` are being iterated twice (by `map`, and then in the `forEach`).

Plus, perhaps we should unwrap `sortOrderPopUp` before doing anything (and assert on that), otherwise this API will fail to do anything, but we wouldn't notice.

Simplenote/PreferencesViewController.swift
```diff
@@ -0,0 +1,314 @@
...
+
+    private func setupThemeFields() {
+        let menuItems: [NSMenuItem] = ThemeOption.allCases.map { theme in
+            let item = NSMenuItem()
+            item.title = theme.description
+            item.tag = theme.rawValue
+            return item
+        }
+
+        menuItems.forEach({ themePopUp.menu?.addItem($0) })
```
jleandroperez: Same note as above

Simplenote/SimplenoteAppDelegate+Swift.swift
```diff
@@ -160,6 +160,24 @@ extension SimplenoteAppDelegate {
 //
 extension SimplenoteAppDelegate {
 
+    @IBAction
+    func preferencesWasPressed(_ sender: Any) {
+        SPTracker.trackPreferencesWasOpened()
+
+        if let preferencesWindow = preferencesWindowController?.window {
+            if preferencesWindow.isVisible {
```
jleandroperez: You can collapse these two `if` statements into a single one:

```
if preferencesWindowController?.window?.isVisible == true
```

Simplenote/SimplenoteAppDelegate+Swift.swift
```diff
@@ -602,4 +620,6 @@ extension SimplenoteAppDelegate {
 //
 private struct Constants {
     static let noteEditorMetadataCacheFilename = ".editor-metadata-cache"
+    static let preferenceStoryboardIdentifier = "Preferences"
```
jleandroperez: No need for this constant. Check `NSStoryboard+Simplenote`

Simplenote/SimplenoteAppDelegate+Swift.swift
```diff
@@ -602,4 +620,6 @@ extension SimplenoteAppDelegate {
...
+    static let preferencesWindowControllerIdentifier = "PreferencesWindowController"
```
jleandroperez: No need for this constant either. Check on the `instantiateWindowController` NSStoryboard extension 😄

Simplenote/SimplenoteAppDelegate.h
```diff
@@ -49,6 +49,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (strong, nonatomic) NoteEditorMetadataCache                       *noteEditorMetadataCache;
 
+@property (nullable, strong, nonatomic) NSWindowController                            *preferencesWindowController;
```
jleandroperez: Nipticky: different spacing with regards of the other properties (yes, I know, we should probably not do this!!)

charliescheer:
I've added another commit that fixes the font size getting out of sync in the preferences view controller

Simplenote/TagListViewController+Swift.swift
```diff
@@ -218,6 +220,12 @@ extension TagListViewController: NSTableViewDataSource, SPTableViewDelegate {
             return
         }
 
+        if let selectedRow = selectedRow,
+           selectedRow == .settings {
+            SimplenoteAppDelegate.shared().preferencesWasPressed(self)
+            return
+        }
```
jleandroperez, line 227: There's a UX problem with having `Settings` in the Tags list:

1. Click on the `Settings` row
2. Preferences UI shows up
3. Close the preferences window

- The Tags list remains highlighting `Settings`, but there are no more settings onscreen.
- While `Settings` is highlighted, it's not clear what `Notes` are being filtered out.

Since it creates a few nipticky UX problems, @SylvesterWilmott do you think we could drop the `Settings` row from the Tags List?

jleandroperez (commented):
Thank you Charlieeee!!! apologies in advance about the annoying notes!!

Simplenote/SimplenoteConstants.swift
charliescheer: >another note, constants should not be capitalized in Swift)

A shoot that is a copy/paste error on my part.  Will move this to an enum somewhere

PR #1071, opened 2022-11-10
Update release tooling

spencertransier:
### Summary

This includes a few changes: 
- Updated Fastlane to version `2.210.1`
- Updated the Release Toolkit to `5.6.0`
- Updated the Buildkite pipeline to use the Xcode 14.0.1 image
- Moved the `Pluginfile` contents over to the `Gemfile`


.buildkite/commands/build-and-test.sh
```diff
@@ -2,10 +2,6 @@
 
 set -eu
 
-# FIXIT-13.1: Temporary fix until we're on the Xcode 13.1 VM
-echo "--- :rubygems: Fixing Ruby Setup"
-gem install bundler
-
```
AliSoftware, line 8: Woohoo, one less hack to keep around! 🎉

.ruby-version
```diff
@@ -1 +1 @@
-2.6.4
+2.7.4
```
AliSoftware, line 1: 👌

Gemfile
```diff
@@ -6,6 +6,9 @@ gem 'octokit', '~> 4.0'
 gem 'rake', '~> 12.3'
 gem 'rubocop', '~> 1.18'
 gem 'xcpretty-travis-formatter', '~> 1.0'
+gem 'fastlane-plugin-wpmreleasetoolkit', '~> 5.6'
+gem 'fastlane-plugin-sentry', '~> 1.14'
+gem 'fastlane-plugin-appcenter', '~> 1.11'
 
 plugins_path = File.join(File.dirname(__FILE__), 'fastlane', 'Pluginfile')
 eval_gemfile(plugins_path) if File.exist?(plugins_path)
```
AliSoftware: Since you deleted the `fastlane/Pluginfile`, you can remove those two lines too as the file don't exist anymore.

(Which would ensure that dependabot would not be tempted to complain about it not supporting such setup)

AliSoftware (approved):
Approving to unblock but I'd recommend removing the `eval_gemfile` lines now that they're not needed—and to also make Dependabot happy—before merging the PR 🙂

Gemfile
spencertransier: @AliSoftware I tried removing those lines, but Fastlane complains about it when trying to run `bundle exec fastlane update_plugins` : 

> Updating plugin dependencies...
> [08:03:39]: It looks like fastlane plugins are not yet set up for this project.
> [08:03:39]: fastlane will modify your existing Gemfile at path '/Users/spencer/GitHub/simplenote-macos/Gemfile'
> [08:03:39]: This change is necessary for fastlane plugins to work
> [08:03:39]: Should fastlane modify the Gemfile at path '/Users/spencer/GitHub/simplenote-macos/Gemfile' for you? (y/n)

My questions: 
1. Should I not be using `update_plugins`? 
2. If it is okay to use `update_plugins`, is there a way to not see that message each time? Pressing `y` just adds those two lines back to the `Gemfile`. Pressing `n` exits out with an error.

AliSoftware: > Should I not be using update_plugins?

TL;DR: Indeed. If we migrate away from `fastlane/Pluginfile`, then no you shouldn't use `update_plugins`.

It's (1) either we use `Pluginfile`, but then we're not supposed to ever edit it manually, and instead we use fastlane's `update_plugins` / `add_plugin` commands to edit the files for us… (2) or we manage the plugins ourselves, put them directly in the `Gemfile`, and then we don't use `add_plugins`/`update_plugins` (but just `bundle update …` like for any other gem)

---

As a matter of fact, if you keep those two lines above in the `Gemfile`, but don't have a `fastlane/Pluginfile` anymore, then using `update_plugins` will likely be a no-op, as it will parse the `fastlane/Pluginfile` (or maybe create it if it doesn't exist, like when you run `add_plugins`?), and find nothing there, so probably won't update any plugin, as it will be considering that you don't have any.

After all the whole point for fastlane to have the plugins listed and maintained in a separate and dedicated `fastlane/Pluginfile` file is only so that fastlane commands like `add_plugins` / `update_plugins` do their modifications in a dedicated and isolated file that they assume is following a given structure and is not manually edited. Because fastlane doesn't want to make any changes to the users' `Gemfile` directly, to avoid potentially messing with custom `Gemfile` syntaxes / structures. So running `update_plugins` definitively won't touch the `Gemfile`, and thus won't update the plugins' versions in the first place if those are declared in the `Gemfile` and not in the `Pluginfile` — See paaHJt-3BY-p2#comment-6307

---

In fact so far every time we've needed to update e.g. the release toolkit plugin in the past, we've just updated the `~> x.y` version in the `Pluginfile` **manually** (e.g. in VSCode) if needed, then ran `bundle update fastlane-plugins-wpmreleasetoolkit`. I'm not sure many people have in our team have used/relied on the official `add_plugin` / `update_plugins` commands, as opposed to editing the `gem 'fastlane-plugin-…', '…'` line manually themselves like we do for other gems in the `Gemfile`.

spencertransier: Good to know, thank you for explaining @AliSoftware! I will follow those steps from now on when updating plugins.

PR #1084, opened 2023-02-06
Deprecate `Deliverfile` in favor of `Fastfile`

iangmaia:
### Fix
Similarly to what was done for WPiOS on https://github.com/wordpress-mobile/WordPress-iOS/pull/16805, this PR removes the `Deliverfile` moving everything to a new lane in the `Fastfile`. For that I also moved out the `privacy_url` and the `copyright` to separate files in the `metadata` folder.

Additionally, the need for an `app_version` set in the `Deliverfile` has been deprecated and its usage is gonna soon be [removed altogether from release-toolkit](https://github.com/wordpress-mobile/release-toolkit/pull/447), so this PR already cleans this up.

### Test
Testing the new lane is similar to https://github.com/wordpress-mobile/WordPress-iOS/pull/16805 (and is particularly difficult for me to do completely, so I'd also appreciate some help):
> 1. Checkout this branch
> 2. Run `bundle exec fastlane update_metadata_on_app_store_connect`
> 3. Verify the `Preview.html` contains valid data. In particular, that it's not attempting to upload screenshots
> 4. **Cancel the upload** 🙏 `Does the Preview on path './fastlane/Preview.html' look okay for you? (y/n) n`

Also good to test the use of the parameter `with_screenshots:true` to see that it will attempt to load the screenshots.

### Review
I have noticed that this project uses an older version of the [release-toolkit](https://github.com/wordpress-mobile/release-toolkit), so perhaps this upgrade needs to happen before this PR lands.

### Release


AliSoftware (changes requested):
Similar to https://github.com/woocommerce/woocommerce-ios/pull/8831#pullrequestreview-1293255247 and https://github.com/Automattic/simplenote-ios/pull/1515#pullrequestreview-1293499317, we need to add `skip_deliver: true` to the call site which will try to bump the version from the `Deliverfile` now that that file is gone

_(this is something we also forgot to do [the first time we did such a change in WPiOS](https://github.com/Automattic/simplenote-ios/pull/1515#:~:text=for%20WPiOS%20on-,wordpress%2Dmobile/WordPress%2DiOS%2316805,-%2C%20this%20PR%20removes)… which required [a follow-up commit to fix it later](https://github.com/wordpress-mobile/WordPress-iOS/commit/6b88866b2bde6125fef1846ad2bb825000c9912d)…)_

AliSoftware:
> Since one lane was removed and another one was added, I imagine the release process needs to be amended as well. Please review carefully the parameters used in the new lane `update_metadata_on_app_store_connect`, making sure it covers the current use cases.
> 
> I have also noticed that this project uses an older version of the [release-toolkit](https://github.com/wordpress-mobile/release-toolkit), so perhaps this upgrade needs to happen before this PR lands.

Good points, and same answers as the ones I gave in https://github.com/Automattic/simplenote-ios/pull/1515#issuecomment-1426053435:
 - [Current scenario for Simplenote Mac](https://github.com/Automattic/mobile-toolbox/blob/trunk/desktop/simplenotemacos-release-scenario.sh) (private repo) doesn't mention the `upload_screenshots` lane you deleted, so there won't be a need to amend the release process to account for the change in this PR. (If anything, we could consider adding a _missing_ step to the Simplenote release process to mention calling this new lane _if necessary_…)
 - Simplenote-macos is in maintenance mode right now, so no worries about it being on older release-toolkit (though that was nice of you to notice). We already know that we'll need more than just the work you've been doing with these PRs before we're ready to migrate to a newer release-toolkit (even just to the existing `6.x`), but at least your PR makes us closer to that goal now 🙂

AliSoftware (approved):

Issue #1083, opened 2023-02-01
Change minimum window width


Brin-o:
### What
I would like the minimum window width to be changed to a smaller amount to allow for easier vertical window tiling. 

### Why
Look at the screenshot attached. Trying to tile the apps into a 3/4 and 1/4 grid is currently impossible because simple notes does not want to take up as little as 1/4 of the width. I dont really see a reason why this is not possible when the sidebar is collapsed.

<img width="1680" alt="image" src="https://user-images.githubusercontent.com/49071768/215993284-7e74bd5e-c318-4aa7-bad0-f9520e6a69b2.png">

charliescheer:
updated to a new minimum in #1086  closing

PR #1086, opened 2023-03-17
Updated logic to let the user Change minimum window

bubudrc:
### Fix
This PR let the users to change the min width of the main window, when is on editor mode. This will fix the issue: https://github.com/Automattic/simplenote-macos/issues/1083

Because the main window width is set to a min of 720 (on Main.storyboard), every time that the user are in the editor mode, it's impossible to reduce the size.

In this PR, I'm listening every time that the user collapse/expand the Split View to set the correct min size.
I'm set the minWidth to 300, because was the min size that let the user to see all the buttons on the tab bar, included the markdown.

Screen Shot 2023-03-17 at 12 30 00

Rollback:
> 1. Tap the collapse 
> 2. Tap on collapse to add the others panels
> 3. Resize the window

### DIFF
--- a/Simplenote/MainWindowController.swift
+++ b/Simplenote/MainWindowController.swift
@@ -14,6 +14,10 @@ class MainWindowController: NSWindowController {
     ///
     override var contentViewController: NSViewController? {
         didSet {
+            if let splitVC = contentViewController as? SplitViewController {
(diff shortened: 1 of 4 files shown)


jleandroperez (approved):
:shipit: 

Thank you @bubudrc !!

Issue #1064, opened 2022-05-25
Automatically delete a note when it's content is empty


michaelkolesidis:
### What
When I delete the whole text content of a note, the note stays in the database as an empty note and in order to be removed, one has to manually trash it. I propose that whenever a user deletes all the text content of a note and then focuses on another element (click on another note, perfoms another action elsewhere), the note is deleted automatically. It is a common feature of note taking apps, implemented in Apple Notes for Mac for example.

### Why
It will save a lot of time, and provide an easy way to delete unwanted notes.

charliescheer:
Thank you for the suggestion.  We discussed this change internally and decided that we are not going to make notes automatically delete if they aren't new and their content is removed.  Primarily we decided not to do this because it makes accessing the version history of a note more difficult if the note is automatically moved to the trash empty as all notes deleted this way will look the same, it makes recovery very difficult.

spoerri:
> we are not going to make notes automatically delete if they aren't new

[Implying that they might automatically delete if they _are_ new :)](https://github.com/Automattic/Simplenote-United/issues/116)

PR #1087, opened 2023-03-17
Updated logic to support duplicate note

### LINKED ISSUE #1066: Adding the ability to duplicate a note
Currently when we right click on any note we get only 3 options, namely:
- Pin to Top
- Copy Internal Link 
- Move to Trash

It would be really helpful if there could also be a **duplicate note option** as well

image

bubudrc:
### Fix
This PR will add the feature that let users to duplicate an existing note.  Will fix: [issue-1066](https://github.com/Automattic/simplenote-macos/issues/1066)
The idea was to create a new note, and set the content and tags of the existing one.

For that, I must to add new menu items and updated the layout and conditions to not duplicate code

> 1. Select any note of the list
> 2. In the Editor view, select the "more options" menu item
> 3. Select "Duplicate Note" option

> 1. Select any note of the list
> 2. In main menu select Note
> 3. Select "Duplicate Note" option

duplicate_note

### Notes
Because this is a new complete feature, and I'm not part of the company, there are some steps that need your interaction (if you want to merge this). 
The new "Duplicate Note" needs localization strings.
The new Duplicate Note, needs update the analytics with a new event, to track this and know from where was called (Notes list, Main menu, Editor Menu)

### DIFF
--- a/Simplenote/NoteEditorViewController+Swift.swift
+++ b/Simplenote/NoteEditorViewController+Swift.swift
@@ -374,7 +374,7 @@ extension NoteEditorViewController: NSMenuItemValidation {
             return validateEditorCopyInterlinkMenuItem(menuItem)
 
         case .editorDuplicateNoteMenuItem:
-            return validateSystemDuplicateNoteMenuItem(menuItem)
+            return validateEditorDuplicateNoteMenuItem(menuItem)
 
         case .editorPinMenuItem:
             return validateEditorPinMenuItem(menuItem)
@@ -423,10 +423,9 @@ extension NoteEditorViewController: NSMenuItemValidation {
         return isDisplayingNote
     }
 
-    func validateListDuplicateNoteMenuItem(_ item: NSMenuItem) -> Bool {
-        // TODO: Add localization keys+values for this new item
+    func validateEditorDuplicateNoteMenuItem(_ item: NSMenuItem) -> Bool {
         item.title = NSLocalizedString("Duplicate Note", comment: "Duplicate Note List Action")
-        return isDisplayingNote
+        return isDisplayingNote && selectedNotes.count == 1
     }
 
     func validateEditorPinMenuItem(_ item: NSMenuItem) -> Bool {
@@ -469,7 +468,7 @@ extension NoteEditorViewController: NSMenuItemValidation {
     }
 
     func validateSystemDuplicateNoteMenuItem(_ item: NSMenuItem) -> Bool {
-        return isDisplayingNote && !viewingTrash
+        return isDisplayingNote && !viewingTrash && selectedNotes.count == 1
     }
 
     func validateSystemNewNoteMenuItem(_ item: NSMenuItem) -> Bool {
--- a/Simplenote/NoteEditorViewController+Swift.swift
+++ b/Simplenote/NoteEditorViewController+Swift.swift
@@ -373,6 +373,9 @@ extension NoteEditorViewController: NSMenuItemValidation {
         case .editorCopyInterlinkMenuItem:
             return validateEditorCopyInterlinkMenuItem(menuItem)
 
+        case .editorDuplicateNoteMenuItem:
+            return validateSystemDuplicateNoteMenuItem(menuItem)
+
         case .editorPinMenuItem:
             return validateEditorPinMenuItem(menuItem)
 
@@ -394,6 +397,9 @@ extension NoteEditorViewController: NSMenuItemValidation {
         case .editorCollaborateMenuItem:
             return validateEditorCollaborateMenuItem(menuItem)
 
+        case .systemDuplicateNoteMenuItem:
+            return validateSystemDuplicateNoteMenuItem(menuItem)
+
         case .systemNewNoteMenuItem:
             return validateSystemNewNoteMenuItem(menuItem)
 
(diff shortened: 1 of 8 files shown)


Simplenote/NoteEditorViewController+Swift.swift
```diff
@@ -417,6 +423,12 @@ extension NoteEditorViewController: NSMenuItemValidation {
         return isDisplayingNote
     }
 
+    func validateListDuplicateNoteMenuItem(_ item: NSMenuItem) -> Bool {
+        // TODO: Add localization keys+values for this new item
```
jleandroperez: Would it be possible to remove this TODO? We've got tooling that takes care of that 😄 

Thank you!

Simplenote/NoteEditorViewController.m
```diff
@@ -599,4 +581,50 @@ - (void)insertChecklistAction:(id)sender
     }
 }
 
+#pragma mark - New Note
+
+- (void)duplicateNote:(nullable Note *)sender
```
jleandroperez: Question:  `duplicateNote` seems to be always invoked with `nil` as a parameter.

WDYT if we remove the parameter altogether?

Simplenote/NoteListViewController.swift
```diff
@@ -731,6 +734,12 @@ extension NoteListViewController: NSMenuItemValidation {
         return isSelectionNotEmpty
     }
 
+    func validateListDuplicateNoteMenuItem(_ item: NSMenuItem) -> Bool {
+        // TODO: Add localization keys+values for this new item
```
jleandroperez:
```suggestion
```

Simplenote/NoteEditorViewController+Swift.swift
jleandroperez:
```suggestion
```

Simplenote/NoteEditorViewController.m
```diff
@@ -599,4 +581,50 @@ - (void)insertChecklistAction:(id)sender
...
+{
+    if (sender != nil) {
+        [self createNoteWasPressed:sender];
+    } else {
+        [self createNoteWasPressed:self.note];
+    }
+}
+
+- (void) createNoteWasPressed:(nullable Note *)sender {
```
jleandroperez: Since this API isn't wired as an IBAction (and `sender` is used usually as the parameter name for the component that's doing the invocation), and UI components won't ever invoke this directly, WDYT if this was renamed as...

- `createNoteFromNote:(nullable Note *)oldNote`
- `createNoteWithNote:(nullable Note *)oldNote`
- `createNewNoteWithContent:(nullable Note *)content`
- `buildNewNoteWithContent:(nullable Note *)content`
- [Another cooler alternative here]

Thank you!

jleandroperez (changes requested):
Hi there @bubudrc! Sending few nipticky notes. Other than that, looks great!!

**Thanks in advance!!**

P.s.: I'll get to your second PR hopefully this week, apologies about the delay.

Simplenote/NoteEditorViewController.m
bubudrc: duplicateNote is called with the selected note on NoteViewController.swift:856
```
@IBAction
    func duplicateNote(_ sender: Any) {
        guard let note = selectedNotes.first else {
            return
        }

        noteEditorViewController.duplicateNote(note)
        // TODO: Set correct analytic event
    }
```
We could remove the parameter and always use the `self.note` value of NoteEditorViewController. It's that ok?

Simplenote/NoteEditorViewController.m
bubudrc: I like your alternatives. For sure I'm will change it :)

bubudrc:
Hi @jleandroperez . Thanks for taking the time to review the PR and sent me the notes.
I'm making the suggested changes, but I'm need a confirmation in one of that changes. With that answer, I'm will be able to update the PR.

Simplenote/NoteEditorViewController.m
jleandroperez: - That sounds super, we could just drop the parameter 😄 
- Perhaps `validateListDuplicateNoteMenuItem` should also check for `selectedNotes.count == 1`. Allowing for Duplicate, when there's more than one entry selected (and only duplicating the first one) would probably not be what I'd expect. WDYT?
- Do you think we could rename the `@IBAction func duplicateNote` to `duplicateNoteWasPressed` ? (so that it looks exactly the same as the other IBAction handlers).

Also, thank you sir!

Simplenote/NoteEditorViewController.m
jleandroperez: Thank you!

jleandroperez:
Apologies about the slow review @bubudrc, and again, thank you for the PR!!

Simplenote/NoteEditorViewController.m
bubudrc: Done.
I'm changed the methods names and also added the `selectedNotes.count == 1` condition, not only on `validateListDuplicateNoteMenuItem` but also on `validateSystemDuplicateNoteMenuItem` to keep a consistence on the logic.

BTW: I renamed `validateListDuplicateNoteMenuItem` to `validateEditDuplicateNoteMenuItem` to match the existing method names

bubudrc:
One more thing. Now, we can duplicate empty notes, but that has none sense. I'm know that [there are an open issue](https://github.com/Automattic/simplenote-macos/issues/1064) asking to remove the empty notes, so fix that issue could be a good opportunity to improve the behavior and avoid duplicate empty notes :)

jleandroperez:
I do agree, perhaps allowing to duplicate an existing note feels odd (but yeah, I also agree, what I would expect is to simply not be able to have empty notes, otherwise the behavior would feel inconsistent!).

Thank you for helping us improve Simplenote @bubudrc !!

:shipit:

jleandroperez (approved):

PR #1090, opened 2023-04-06
Adds new tracks events

jleandroperez:
### Fix
In this PR we're wiring new Tracks Events, related to Note Duplication.

### Test
- [ ] Verify you can Duplicate a Note from the Editor
- [ ] Verify you can Duplicate a Note from the Notes List
- [ ] Verify you can use the **CMD + D** shortcut to duplicate a Note

### Release


eshurakov (approved):
Code looks good! 

I couldn't test the app because I have some issues installing dependencies on my laptop. But I trust that it works :-)

PR #1091, opened 2023-04-18
Merge 2.18 code freeze

mokagio:
This includes a few changes that were necessary to build the app in macOS versions without Python 2.7, and with the Xcode 14.3 toolchain.

## Next steps

I'll need to open a PR to track upgrading the old Python localization script to Python 3 (via built-in `2to3`) and show that the diff in the `strings` it generates is the same as this PR.

The `strings` here were generated on an older Mac on which `python` was still available, but I think it would be foolish to expect an old Mac to keep being available. 

Worth pointing out that we could upgrade the tooling instead, but give the shift in focus away from Simplenote macOS, I think the 10 minutes it'll take me to verify the diff and make a PR will be sufficient.


External/Sparkle/bin/codesign_embedded_executable
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
mokagio, line 1: It's not entirely appropriate to upgrade this file since it should be a copy of Sparkle, but I think we can close an eye given this is the only change that was required to use Python 3

Gemfile
```diff
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'cocoapods', '>= 1.10'
+gem 'cocoapods', '~> 1.12'
```
mokagio, line 3: I had to upgrade CocoaPods because I run into a CDN issue with that older version.

```
[!] CDN: trunk URL couldn't be downloaded: https://cdn.cocoapods.org/deprecated_podspecs.txt Response: URL using bad/illegal format or missing URL
```

Simplenote.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -1,34 +1,32 @@
...
+      "identity" : "simplenotesearch-swift",
+      "kind" : "remoteSourceControl",
+      "location" : "git@github.com:Automattic/SimplenoteSearch-Swift.git",
+      "state" : {
+        "revision" : "499d2809d169fcbeb9ff75568d9f1f937f290ffc",
+        "version" : "1.3.0"
+      }
+    }
+  ],
+  "version" : 2
```
mokagio, line 31: I don't remember at which point, but Xcode changed the `.resolved` version in the last couple of years or so. This is just what happened as I opened the app in Xcode 14.3. You can see the packages haven't changed version.

Podfile
```diff
@@ -23,3 +25,34 @@ abstract_target 'Automattic' do
...
+
+    shell_script_input_lines = File.readlines(shell_script_path)
+    shell_script_output_lines = shell_script_input_lines.map { |line| line.sub("source=\"$(readlink \"${source}\")\"", "source=\"$(readlink -f \"${source}\")\"") }
+    File.open(shell_script_path, 'w') do |f|
+      shell_script_output_lines.each do |line|
+        f.write line
+      end
+    end
+  end
+end
```
crazytonyli, line 41: I just noticed CocoaPods [released 1.12.1](https://github.com/CocoaPods/CocoaPods/releases/tag/1.12.1) 4 hours ago, have you tried the latest version which hopefully fixed these integration issues in Xcode 14.3?

mokagio: Neat! I'll try it out, it does mention the fix we need:

Screenshot 2023-04-19 at 3 55 56 pm

crazytonyli (approved):

PR #1118, opened 2024-01-10
Use ruby 3.2.2

crazytonyli:
What says in the title. It should be okay to merge this PR as long as CI is green.


.buildkite/pipeline.yml
```diff
@@ -5,7 +5,7 @@ common_params:
     - automattic/a8c-ci-toolkit#2.15.0
   # Common environment values to use with the `env` key.
   - &common_env
-    IMAGE_ID: xcode-14.2
+    IMAGE_ID: xcode-14.3.1
```
AliSoftware, line 8: I'm guessing this PR is only the first stage of a larger task to migrate Simplenote macOS to Xcode 15, but you're doing it step by step, first focusing on the Ruby update, to only migrate to Xcode 15 in a subsequent PR?

AliSoftware (approved):

.buildkite/pipeline.yml
crazytonyli: I was just doing the Ruby upgrade. I just saw an update from @spencertransier that [the iOS app was moved to Apple Silicon agents](https://github.com/Automattic/simplenote-ios/pull/1540). I'm not sure if he has plan to migrate the macOS app too.

Issue #1122, opened 2024-01-19
Rewrite drag mechanism for Tag list


charliescheer:
The method we are using to drag and drop tags to re order them in the tableview has been depreciated.  

We should update how tag reordering works to use modern APIs

Issue #1121, opened 2024-01-19
Status bar is expanding to cover the whole window


charliescheer:
When building with Xcode 15.2, if you activate the status bar you can not see most of the SN window.  The status bar is visible, but it seems to expand over all of the other views.  This is not a great experience...

### Reproduced
***(Required)*** If you cannot reproduce this bug consistently, please elaborate.  List the steps to reproduce the behavior.  For example:
1. Launch SNMac from Xcode 15.2
2. Go to view > show status bar

Notice that you can't see most of the UI

<img src="https://cdn-std.droplr.net/files/acc_593859/scPYU2" />

PR #1123, opened 2024-01-19
Update minimum os version

### LINKED ISSUE #1115: Fix existing warnings in SNMac
There are currently a few warnings in the SNMac project.  It would be good to update the app to squash these warning.

Note that one of the things that will need to be done here is updating Automattic Tracks to use the most recent version of the tracks package, which will require moving tracks to SPM instead of cocoa pods.  Tracking that as a separate issue in #1114

### LINKED ISSUE #1116: Update minimum supported OS version to MacOS 12
Simplenote Mac has been static for a little while. It would be good to update Simplenote's minimum supported version to something a bit more modern.

This is gonna require a couple of tasks

- [x] Update the minimum version to MacOS 12
- [x]  Fix any depreciation warnings caused by the minimum version change
- [x] Remove any version specific code for older versions of Mac OS
- [ ]  Update any public documentation that states the minimum version

charliescheer:
### Fix
We are updating the minimum supported OS version for SNMac to MacOS 12.

In this PR I have updated the minimum versions and then fixed the resulting warnings/errors that appeared.  I have also removed any version checking.

Note that a few changes still need to be dealt with but they are larger changes than this PR.  Issues not taken care of:
[Rewrite tag list view drag drop](https://github.com/Automattic/simplenote-macos/issues/1122)
[Status bar renders over whole app](https://github.com/Automattic/simplenote-macos/issues/1121)

### Test
There isn't anything specific that was changed here.  A general smoke test should be fine.

### Review


jleandroperez (approved):
✅ Looks super!!

One minor note, the Deauthenticated State still has a couple issues with UI elements (not sure if that was addressed elsewhere, though!!).

:shipit:

charliescheer:
I'll create an issue for the login view

PR #1124, opened 2024-01-19
Updated tracks to fetch from SPM instead of cocoapods

### LINKED ISSUE #1114: Update version of Automattic Tracks
Simplenote Mac is currently using version tracks version 0.8. This works fine but it is really really old and the sentry version included in that version of tracks is causing an error in the project. We should update tracks to the most recent version, but this is gonna cause issues cause we are importing tracks from cocoa pods and the latest version available on cocoa pods is 2, which is also very old. So we should move tracks to SPM so we can get the most recent version of tracks.

This is gonna require a few tasks:

 Add Automattic tracks iOS to SPM
 Remove Automattic tracks iOS from cocoa pods
 This will break stuff, so fix the usage of tracks inside the project to call to the SPM package (fix the errors)

charliescheer:
### Fix
The version of Automattic tracks is very outdate (0.8.0)

Tracks also no longer supports downloading through cocoa pods, which SNMac is using.  

So we should update that.  Which is what I have done here.  

### Test
1. pull the branch.  Run bundle exec pod install on your snmac repo.  Clean the project
2. Confirm the app builds
3. Attempt to send an event to tracks.  Confirm it arrives.

### Review

### DIFF
--- a/Podfile
+++ b/Podfile
@@ -13,7 +13,6 @@ workspace 'Simplenote.xcworkspace'
 abstract_target 'Automattic' do
   # Automattic Shared
   #
-  pod 'Automattic-Tracks-iOS', '0.8.0'
   pod 'Simperium-OSX', '1.9.0'
 
   # Main Target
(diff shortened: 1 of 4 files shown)


jleandroperez (approved):
Looks great!!!

:shipit:

PR #1125, opened 2024-01-23
Rewrite TagListViewController drag in swift

### LINKED ISSUE #1122: Rewrite drag mechanism for Tag list

charliescheer:
### Fix
Currently TagListViewController is using a deprecated method to be able to drag/drop reorder the tags list.  After updating the minimum OS version to 12, warnings are now appearing that we needed to retire our use of the method  `(BOOL)tableView:(NSTableView *)tableView writeRowsWithIndexes:(NSIndexSet *)rowIndexes toPasteboard:(NSPasteboard *)pboard`. In this PR I have mode that change

While I was in there working on it, I figured why not move all of the drag/drop methods for the tag list to swift.  So did that too.

Fixes #1122 

### Test
1. Build Simplenote Mac and log into an account with multiple tags (or create multiple tags)
2. Attempt to rearrange the tags in the tag bar by clicking and dragging.  Confirm you can rearrange the tags
3. Attempt to drag something onto the tags list from another view or app.  Confirm that you can not
4. Open the settings for simplenote and activate "Sort tags alphabetically" Confirm that you can not drag the tags when this is active.

### Review


jleandroperez (approved):
✅ Verified that Drag + Drop works as expected
✅ Verified that you cannot drag anything else!!

:shipit:

PR #1126, opened 2024-01-23
Status bar clipping

### LINKED ISSUE #1121: Status bar is expanding to cover the whole window

charliescheer:
### Fix
Currently there is an issue when you run SNMac built from Xcode 15.2 where the status bar covers the entire Simplenote window.  This PR Fixes that.

Fixes #1121 

### Test
1. Build Simplenote mac
2. In the menu bar go to View > Show status bar.  Confirm the status bar appears and does not cover the main window.

### Review
> Only one developer is to review these changes, but anyone can perform the review.


jleandroperez (approved):
SUPER interesting... I wonder why it was growing beyond bounds though!?!?

Nice catch!!

:shipit:

PR #1129, opened 2024-01-23
Fixed auth view controller rendering login fields incorrectly on launch

### LINKED ISSUE #1127: Login screen UI layout off
Something is going on with the login view where the email and password fields are squished together

Doesn't look great....   We should fix that up

charliescheer:
### Fix
When the login view is originally rendered we start in a "sign in" state which displays just the email field and the login button.  If you tap the login button then we display the password field, forgot password, all the things you need.  Well.. right now when the view first renders all of the login buttons and fields are still visible, but they have no frame size so they are all on top of all of the other views.  

To make the animations work we still need those size changes, but we should also set the alpha values of those fields to 0 so they don't appear.  This PR fixes that.

Fixes #1127 

### Test
1. Log out of simplenote Mac.  
2. Confirm that when the sign up view appears you do not see a password field/forgot password link/WordPress sso login.
3. Tap the login button and make sure the views animate in
4. tap sign up and confirm that the sign up view returns correctly

### Review
> Only one developer is to review these changes, but anyone can perform the review.


jleandroperez (approved):
:shipit:

PR #1130, opened 2024-02-07
Upgrade CI to Apple Silicon

spencertransier:
This PR updates the CI image from `14.3.1` to `15.2-xl` (an image that was created with 120 GB of space instead of 90 GB) so that it can run on Apple Silicon CI. The Buildkite-CI repo PR was already merged in https://github.com/Automattic/buildkite-ci/pull/358, so we need to merge this PR in order to SNMac to be able to use CI. 

I also updated the CI Toolkit from `2.15.0` to `3.0.1` while I was at it (no relevant breaking changes). 

Merging to `release/2.20` so that release can move forward.

I created a test Buildkite run here that passed successfully: https://buildkite.com/automattic/simplenote-macos/builds/105

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -2,7 +2,7 @@
 common_params:
   # Common plugin settings to use with the `plugins` key.
   - &common_plugins
-    - automattic/a8c-ci-toolkit#2.15.0
+    - automattic/a8c-ci-toolkit#3.0.1
   # Common environment values to use with the `env` key.
   - &common_env
     IMAGE_ID: xcode-14.3.1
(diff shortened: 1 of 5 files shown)


.buildkite/pipeline.yml
AliSoftware, line 1: This could be the perfect occasion to adopt the approach discussed in paaHJt-5Rr-p2, so that ci-toolkit version and IMAGE_ID are shared across pipelines, and only need to be updated in a single place going forward?

(This can also be a separate PR if this needs to be unblocked to do a SN build ASAP in the meantime)

AliSoftware (approved):
Approved with suggestion, either for now or for a future PR.

.buildkite/pipeline.yml
spencertransier: I definitely want to add that, but like you said, I need to get this unblocked to make a new build.

PR #1140, opened 2024-04-11
Add spotlight support for Simplenote

### LINKED ISSUE #1000: Implement Spotlight Search Support
### Details
- [x] Integrate with Spotlight Search API(s)

charliescheer:
### Fix
It was requested in #1000 to index all of the notes in Simplenote so they can be searched for in spotlight.  This is a great idea, so I have done just that!

With this PR you will be able to type in the text or title of any note into spotlight and the not will appear in the search results.  Then you will be able to launch Simplenote directly to that note from spotlight.

fixes #1000 

### Test
Prep:
We want to confirm indexing and removing indexing before confirming the one time indexing of all of the existing notes.  So at first we want to not run that indexing.... so:

1. Go to SimplenoteAppDelegate L.126 and comment out the line `[self indexSpotlightItemsIfNeeded];` so we don't index all your notes right away

Test adding and removing notes from the spotlight index:
1. Open up Simplenote on another device that is logged into the same account and look for a note with some text.  Open up spotlight and type that text in to the search field.  Confirm that the note does not appear
2. Open Simplenote on this device, open up that same note, add some text to it and then open a different note to force a save
3. Return to Spotlight and search for some of the text in that note.  Confirm that the note appears in spotlight as a search result
4. select that search result.  Confirm that Simplenote opens and goes to the note in question
5. Now put that note in the trash.  Go back to spotlight and search the same text, confirm that you do not see the note appear.

One time index:
6. now uncomment out the code we disabled before in the SimplenoteAppDelegate
7. relaunch the app.  Give the app some time to do the indexing, this will happen in the background so you won't notice or get any indication that it is doing so.
8. Now search for text from any of your notes.  Confirm it appears in spotlight
9. Put a break point in SimplenoteAppDelegate L602.  Run the app again and confirm you do not hit this breakpoint.  We only want to run that indexing once.

### Review

### Release
> `RELEASE-NOTES.txt` was updated in ca5d1a with:
> 
> > Added spotlight search support for notes

### DIFF
--- a/Simplenote/CSSearchable+Helpers.swift
+++ b/Simplenote/CSSearchable+Helpers.swift
@@ -31,8 +31,29 @@ extension CSSearchableItem {
 }
 
 extension CSSearchableIndex {
+    // MARK: - Index Notes
+    @objc
+    func indexSpotlightItems(in context: NSManagedObjectContext) {
+        guard Options.shared.indexNotesForSpotlight else {
+            return
+        }
+
+        context.perform {
+            if let deleted = try? context.fetchObjects(for: "Note", withPredicate: NSPredicate(format: "deleted == YES")) as? [Note] {
+                CSSearchableIndex.default().deleteSearchableNotes(deleted)
+            }
+
+            if let notes = try? context.fetchObjects(for: "Note", withPredicate: NSPredicate(format: "deleted == NO")) as? [Note] {
+                CSSearchableIndex.default().indexSearchableNotes(notes)
+            }
+        }
+    }
 
     @objc func indexSearchableNote(_ note: Note) {
+        guard Options.shared.indexNotesForSpotlight else {
+            return
+        }
+
         let item = CSSearchableItem(note: note)
         indexSearchableItems([item]) { error in
             if let error = error {
@@ -42,6 +63,10 @@ extension CSSearchableIndex {
     }
 
     @objc func indexSearchableNotes(_ notes: [Note]) {
+        guard Options.shared.indexNotesForSpotlight else {
+            return
+        }
+
         let items = notes.map {
             return CSSearchableItem(note: $0)
         }
--- a/Simplenote/CSSearchable+Helpers.swift
+++ b/Simplenote/CSSearchable+Helpers.swift
@@ -40,11 +40,11 @@ extension CSSearchableIndex {
 
         context.perform {
             if let deleted = try? context.fetchObjects(for: "Note", withPredicate: NSPredicate(format: "deleted == YES")) as? [Note] {
-                CSSearchableIndex.default().deleteSearchableNotes(deleted)
+                self.deleteSearchableNotes(deleted)
             }
 
             if let notes = try? context.fetchObjects(for: "Note", withPredicate: NSPredicate(format: "deleted == NO")) as? [Note] {
-                CSSearchableIndex.default().indexSearchableNotes(notes)
+                self.indexSearchableNotes(notes)
             }
         }
     }
--- a/Simplenote/CSSearchable+Helpers.swift
+++ b/Simplenote/CSSearchable+Helpers.swift
@@ -83,9 +83,7 @@ extension CSSearchableIndex {
     }
 
     @objc func deleteSearchableNotes(_ notes: [Note]) {
-        let ids = notes.map {
-            return $0.simperiumKey!
-        }
+        let ids = notes.compactMap({ $0.simperiumKey })
 
         deleteSearchableItems(withIdentifiers: ids) { error in
             if let error = error {
--- a/Simplenote/CSSearchable+Helpers.swift
+++ b/Simplenote/CSSearchable+Helpers.swift
@@ -1,0 +1,72 @@
+//
+//  CSSearchableItem+Helpers.swift
+//  Simplenote
+//
+//  Created by Michal Kosmowski on 25/11/2016.
+//  Copyright © 2016 Automattic. All rights reserved.
+//
+
+import Foundation
+import CoreSpotlight
+import UniformTypeIdentifiers
+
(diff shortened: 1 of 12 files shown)


Simplenote/SimplenoteAppDelegate+Swift.swift
```diff
@@ -237,6 +238,24 @@ extension SimplenoteAppDelegate {
 
         SPTracker.trackSettingsStatusBarDisplayMode(hidden: Options.shared.statusBarHidden)
     }
+
+    @objc
+    func handleUserActivity(_ userActivity: NSUserActivity) -> Bool {
+        if userActivity.activityType == "com.apple.corespotlightitem" {
```
charliescheer: This is a very simple implementation of handle User activity for now, but one of the other things I would like to do in the near future is bring SNMac into parity with SNiOS as far as shortcuts are concerned.  When I do that this will need to be expanded, but for now we aren't handling any other user activity types

roundhill:
I think this feature is great but I'm wondering what you think of making it opt-in. I would imagine some users may not like the idea of their notes showing up in spotlight in case they have personal information in them. WDYT?

Simplenote/SimplenoteAppDelegate+Swift.swift
jleandroperez: I think you can use the `CSSearchableItemActionType` constant instead of `com.apple.corespotlightitem`:

```
(lldb) po CSSearchableItemActionType
"com.apple.corespotlightitem"
```

Simplenote/SimplenoteAppDelegate.m
```diff
@@ -557,4 +572,43 @@ - (NSUndoManager *)windowWillReturnUndoManager:(NSWindow *)window
...
+
+#pragma mark ================================================================================
+#pragma mark Spotlight
+#pragma mark ================================================================================
+
+- (void)indexSpotlightItemsIfNeeded
+{
+    // This process should be executed *just once*, and only if the user is already logged in (AKA "Upgrade")
+    NSString *kSpotlightDidRunKey = @"SpotlightDidRunKey";
+    NSUserDefaults *defaults = [NSUserDefaults standardUserDefaults];
```
jleandroperez: These two could potentially live in `Options` ?

jleandroperez (commented):
Code looks super sharp!! 

+1 to what Dan said, perhaps we should have an opt-in setting, for privacy reasons?

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Simplenote/CSSearchable+Helpers.swift
```diff
@@ -0,0 +1,97 @@
...
+            return
+        }
+
+        context.perform {
+            if let deleted = try? context.fetchObjects(for: "Note", withPredicate: NSPredicate(format: "deleted == YES")) as? [Note] {
+                CSSearchableIndex.default().deleteSearchableNotes(deleted)
+            }
+
+            if let notes = try? context.fetchObjects(for: "Note", withPredicate: NSPredicate(format: "deleted == NO")) as? [Note] {
+                CSSearchableIndex.default().indexSearchableNotes(notes)
```
jleandroperez: We can probably replace `CSSearchableIndex.default()` with `self` (same in L43)

Simplenote/CSSearchable+Helpers.swift
```diff
@@ -0,0 +1,97 @@
...
+        }
+    }
+
+    @objc func deleteSearchableNote(_ note: Note) {
+        deleteSearchableNotes([note])
+    }
+
+    @objc func deleteSearchableNotes(_ notes: [Note]) {
+        let ids = notes.map {
+            return $0.simperiumKey!
```
jleandroperez: How abooout if we replace the `!` and turn the `map` into a `compactMap` ? (just for safety?)

Simplenote/SimplenoteAppDelegate+Swift.swift
```diff
@@ -237,6 +238,24 @@ extension SimplenoteAppDelegate {
...
+        if userActivity.activityType == CSSearchableItemActionType {
+            presentNote(for: userActivity)
+            return true
+        }
+
+        return false
+    }
+
+    func presentNote(for userActivity: NSUserActivity) {
+        guard let uniqueIdentifier = userActivity.userInfo?[CSSearchableItemActivityIdentifier] as? String else {
```
jleandroperez: Nit: Hooow about if we just move the `simperiumKey` unwrapping to `handleUserActivity`? (that way we can just call `displayNote`?)

Simplenote/SimplenoteAppDelegate.m
```diff
@@ -557,4 +572,37 @@ - (NSUndoManager *)windowWillReturnUndoManager:(NSWindow *)window
...
+
+    [defaults setBool:true forKey:kSpotlightDidRunKey];
+    [defaults synchronize];
+
+    if (self.simperium.user.authenticated == false) {
+        return;
+    }
+
+    [self indexSpotlightItems];
+}
```
jleandroperez: I'm not 100% sure if we still need this API:

1. In the upgrade flow, I'm seeing the `Spotlight` Option disabled, by default
2. And all of this ends up in `CSSearchableIndex` bailing out

Perhaps we can drop these two API(s)?

jleandroperez (approved):
Sending a couple questions (I'm not sure the `indexSpotlightItemsIfNeeded` API still makes sense?).

Other than that, **REAL nice work!!!**

:shipit:

PR #1142, opened 2024-04-16
Fix warnings in project

charliescheer:
### Fix
I noticed a few warnings had crept into SN Mac (most of which were created by me recently.... oops) this PR updates the project to remove all warnings.

### Test
Not much to test.  
- confirm app builds
- confirm you can drag and drop tags
- confirm on login if you switch from signup to login all of the views appear correctly.

### Review

### Release


jleandroperez (approved):
✅ Verified the app builds
✅ Verified that Tags Drag+Drop still works
✅ Verified you can log out / log in

Thank you Charlie!! :shipit:

PR #1150, opened 2024-05-07
Fastlane `match`: Migrate Code Signing from Google Cloud to S3

spencertransier:
### Summary
(I was just going to update the profiles that had expired, but decided to update these lanes while I was at it)

This PR: 
- Migrates the code signing certificates and provisioning storage from Google Cloud to S3
- Updates Fastlane from `2.219.0` to `2.220.0`
- Updates the needed values in the example user `env` file
- Removes unneeded files from `.configure` - We don't need the Google or Apple API auth files because this PR switches to using S3 and the `app_store_connect_api_key` Fastlane action. By using the `app_store_connect_api_key` action, it'll make rotating the ASC key much easier because it's just pulling from the shared file in Mobile Secrets. 

### Test
1. In your local `~/.simplenotemac-env.default` env file, add values for `MATCH_S3_ACCESS_KEY`, `MATCH_S3_SECRET_ACCESS_KEY`, and `MATCH_PASSWORD` from Mobile Secrets
2. Run `bundle exec fastlane app_store_code_signing`
3. Verify that it completes successfully
4. Run `bundle exec fastlane app_store_code_signing readonly:false`
5. Verify that it completes successfully (everything is up to date, so no changes will be made)

### Release


fastlane/Fastfile
```diff
@@ -495,3 +501,29 @@ lane :download_strings_files_from_glotpress do
...
+  # Fail early if secrets not available via `get_required_env`.
+  # Otherwise, Fastlane will prompt to type them.
+  access_key = get_required_env('MATCH_S3_ACCESS_KEY')
+  secret_access_key = get_required_env('MATCH_S3_SECRET_ACCESS_KEY')
+
+  match(
+    storage_mode: 's3',
+    s3_bucket: 'a8c-fastlane-match',
+    s3_access_key: access_key,
+    s3_secret_access_key: secret_access_key,
```
mokagio, line 514: I got this while testing

image

Configuring the region as well fixed it.

image

---

```suggestion
    s3_region: 'us-east-2',
```

fastlane/Fastfile
```diff
@@ -1,26 +1,41 @@
...
 
+Dotenv.load(USER_ENV_FILE_PATH)
+
+APPLE_TEAM_ID = 'PZYM8XX95Q'
+# Generate the App Store Connect API key to download code signing certificates and profiles with `app_store_code_signing`
+APPLE_API_KEY = app_store_connect_api_key(
+  key_id: get_required_env('APP_STORE_CONNECT_API_PZYM8XX95Q_KEY_ID'),
+  issuer_id: get_required_env('APP_STORE_CONNECT_API_PZYM8XX95Q_ISSUER_ID'),
+  key_content: get_required_env('APP_STORE_CONNECT_API_PZYM8XX95Q_KEY_CONTENT')
+)
```
AliSoftware: Putting these lines here in the global scope not only means that every invocation of fastlane will call the `app_store_connect_api_key` method even when we don't need it for the lane we're running (and even when running `fastlane lanes` to list available lanes), but also that any run of `fastlane` will fail if you don't have one of the `get_required_env` vars set in your environment, while I can totally imagine devs not having those env vars set local and wanting to run a lane which doesn't need, yet won't be able to because of this global call.

---

I'd suggest instead to:
 - Either create a helper method to do those lines and call it only in the lanes that need it
 - Or even better, map the environment variables used by `app_store_connect_api_key` action's `ConfigItem` with the env vars used here (`APP_STORE_CONNECT_API_KEY_KEY_ID="${APP_STORE_CONNECT_API_PZYM8XX95Q_KEY_ID}"` etc)—either in the Fastfile via `ENV['…']="…"`, or arguably better, directly in `~/.mobile-secrets/…/simplenote-macos/environment`—so that you can then just call the `app_store_connect_api_key` without any parameter in the lanes you need the ASC API key (thus no need to "pollute" the Fastfile to create a dedicated helper method when the action w/o params already does what you need)

fastlane/Fastfile
```diff
@@ -1,26 +1,41 @@
 # frozen_string_literal: true
 
-default_platform(:ios)
+default_platform(:mac)
```
AliSoftware, line 3: Is this change intended?

fastlane/Fastfile
```diff
@@ -495,3 +501,29 @@ lane :download_strings_files_from_glotpress do
...
+# Optionally, it can create any new necessary certificates or profiles.
+#
+# @option [Boolean] readonly (default: true) Whether to only fetch existing certificates and profiles, without generating new ones.
+lane :app_store_code_signing do |options|
+  # Fail early if secrets not available via `get_required_env`.
+  # Otherwise, Fastlane will prompt to type them.
+  access_key = get_required_env('MATCH_S3_ACCESS_KEY')
+  secret_access_key = get_required_env('MATCH_S3_SECRET_ACCESS_KEY')
+
+  match(
```
mokagio: Nitpick. I prefer to use the action's full name, `sync_code_signing`, because it's clearer to understand what it does when reading the lane. Same for using `run_tests` instead of `scan`, `upload_to_app_store` instead of `deliver`, etc.

What do you think?

---

```
➜ bf action match
[✔] 🚀
Loading documentation for match:

+--------------------------------------------------------------+
|                            match                             |
+--------------------------------------------------------------+
| Alias for the `sync_code_signing` action                     |
```

mokagio (approved):
Tested locally by running the new code signing lane. 

I also added a test lane to build the app for App Store. It succeeded.

fastlane/Fastfile
```diff
@@ -37,11 +52,20 @@ SKIP_CODE_SIGNING_XCARGS = {
...
+  # Ensure we use the latest version of the Release Toolkit
+  check_for_toolkit_updates unless is_ci || ENV['FASTLANE_SKIP_TOOLKIT_UPDATE_CHECK']
+
+  # Check that the env files exist
+  unless is_ci || File.file?(USER_ENV_FILE_PATH)
+    error_message = <<~ERROR
+      #{USER_ENV_FILE_PATH} not found. Please copy env/user.env-example to #{USER_ENV_FILE_PATH} and fill in the values") using the following command:
+      \tcp #{PROJECT_ROOT_FOLDER}/fastlane/env/user.env-example #{USER_ENV_FILE_PATH}
+    ERROR
+    UI.user_error!(error_message)
```
AliSoftware: I am not sure `before_all` runs before any code that is defined in the global scope (at the top of the `Fastfile` outside any block)… and thus will run before the `Dotenv.load(…)` from line 26 which tries to load said file? 🤔

fastlane/Fastfile
```diff
@@ -416,15 +422,15 @@ lane :build_and_upload_app_store do |options|
     zip(path: archive_path, output_path: archive_zip_path)
 
     create_release(
-      repository: GHHELPER_REPO,
+      repository: GITHUB_REPO,
```
AliSoftware, line 419: 👍🙇🏻

fastlane/env/user.env-example
```diff
@@ -2,6 +2,12 @@ GHHELPER_ACCESS=
...
-CIRCLE_CI_AUTH_TOKEN=<CircleCI Auth Token>
+# Fastlane Match
+MATCH_S3_ACCESS_KEY=<S3 Access Key>
+MATCH_S3_SECRET_ACCESS_KEY=<S3 Secret Access Key>
+MATCH_PASSWORD=<Fastlane Match Password>
+
+# App Store Connect API
+APP_STORE_CONNECT_API_PZYM8XX95Q_KEY_ID=<ASC Key ID>
+APP_STORE_CONNECT_API_PZYM8XX95Q_ISSUER_ID=<ASC Key Issuer ID
+APP_STORE_CONNECT_API_PZYM8XX95Q_KEY_CONTENT=<ASC Key Content>
```
AliSoftware: If you go with my suggestion above to use the env vars expected by the `app_store_connect_api_key` action's `ConfigItem` instead of custom ones, this will allow to run `app_store_connect_api_key` without any parameter and thus make the call site in the `Fastfile` way less verbose 🙂

fastlane/Fastfile
AliSoftware: In addition, calling `app_store_connect_api_key` action only within the lane that needs it means that you also don't need to store the result of the action in a variable to use it afterwards, because `app_store_connect_api_key` sets the key as a shared value in the lane context, which can then be automatically picked up by subsequent actions that need an API Key 😉 

See https://docs.fastlane.tools/app-store-connect-api/#use-the-shared-value-in-lane-context

fastlane/Fastfile
spencertransier: Fixed in https://github.com/Automattic/simplenote-macos/commit/0c40b68a3d58ba47f531d2ce2439b073731d1a95

fastlane/Fastfile
spencertransier: @AliSoftware Yep, I figured we'd want the default to be Mac instead of iOS. Is that incorrect?

AliSoftware: Oh right I didn't notice this was a repo only for macOS (as opposed to eg DayOne for which both platforms live in same repo) 😅

Initially I thought that maybe that change was made while you were debugging stuff and forgot to roll back or something, not understanding why we'd prefer one platform over the other, but now that I've realized this repo is _only_ for macOS that makes way more sense 😅

fastlane/env/user.env-example
spencertransier: Updated in https://github.com/Automattic/simplenote-macos/commit/87b404ef0c480d9a4c846dc85882d5cfc55aaf21

fastlane/Fastfile
spencertransier: Great idea! I implemented this in https://github.com/Automattic/simplenote-macos/commit/1e06f78a2ebe3afe668911f87872af42e244894c

fastlane/Fastfile
spencertransier: Fixed in https://github.com/Automattic/simplenote-macos/commit/4359b9007bd86288d5077ca0aac4762f631320db

fastlane/Fastfile
spencertransier: I've gone back and forth on whether to use the "brand" names like `match`, `deliver`, etc. versus the generic names like `sync_code_signing`, etc. Sometimes I forget that the generic name is a Fastlane action and not a helper method in the Fastfile 😅 . But I do agree that the readability of the generic names makes reading through a Fastfile much easier. 

Updated in https://github.com/Automattic/simplenote-macos/commit/462994f3a4026b32ffadf846512fdd73fa7b2fb9

PR #1153, opened 2024-05-18
Add intents extension code signing 

spencertransier:
### Fix
Completes pdnsEh-1Fh-p2. This PR adds a bundle identifier for the Intents extension. I've added the identifier on the Apple developer portal and ran the command to generate the new provisioning profile. I also added a development provisioning profile to be downloaded through MC. l

### Test
1. Run `bundle exec fastlane app_store_code_signing`
2. Verify that it completes successfully
3. Run `bundle exec fastlane app_store_code_signing readonly:false`
4. Verify that it completes successfully (nothing should actually change since I've already generated the intents extension)

### Release

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -14,6 +14,7 @@ INTERNAL_SCHEME = 'Simplenote'
 APP_STORE_SCHEME = 'Simplenote-AppStore'
 BUILD_FOLDER = 'build'
 APP_STORE_BUNDLE_IDENTIFIER = 'com.automattic.SimplenoteMac'
+APP_STORE_BUNDLE_IDENTIFIER_INTENTS = "#{APP_STORE_BUNDLE_IDENTIFIER}.Intents".freeze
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -14,6 +14,7 @@ INTERNAL_SCHEME = 'Simplenote'
 APP_STORE_SCHEME = 'Simplenote-AppStore'
 BUILD_FOLDER = 'build'
 APP_STORE_BUNDLE_IDENTIFIER = 'com.automattic.SimplenoteMac'
+APP_STORE_BUNDLE_IDENTIFIER_INTENTS = "#{APP_STORE_BUNDLE_IDENTIFIER}.Intents".freeze
```
spencertransier, line 17: Because there are only two bundle identifiers for SNMac, I decided to just keep it nice and simple instead of our more complicated set ups for the other apps with more identifiers.

crazytonyli (approved):

PR #1157, opened 2024-05-22
Refactor Core Data

charliescheer:
### Fix
On the way to being able to fetch core data entities for the shortcuts, I noticed that the core data stack for SNMac was all setup in the app delegate.  This is fine because we aren't sharing the database outside of the core app currently, but soon I will be fetching from the database to do shortcuts, at which point we will have duplication problems.

So here I am extracting the core data setup into it's own class so we can easily setup a core data stack when running shortcuts/intents.

### Test
1. Build Simplenote
2. confirm that you can interact with core data objects in the app, so notes/tags/etc

### Review


charliescheer:
I identified an issue with the signing certificates/intents extension and removed a duplicate target we no longer needed in another PR.  These changes are more easily rebuilt on top of that branch then where they are currently.  Gonna close this PR in favor of ones I will replace it with shortly

PR #1158, opened 2024-05-23
Drop beta target

charliescheer:
### Fix
In the SNMac project we have maintained two targets, Simplenote and Simplenote-AppStore, and every file we added needed to be in both.  This has been fine until we started adding extensions that needed to be embedded in a parent app.  That was going to cause a lot of duplication.  Also, the only reason for the second extension was to make pushing beta versions of the app possible.  We have made changes to how we do beta apps that make this no longer needed.

To simplify the app I have dropped the extra target in this PR.  There was also an intents extension that was broken (how I discovered this) so that was dropped as well.  All in all in this PR I have:

1. Removed the broken intents extension
2. removed the Simplenote target and build scheme
3. Renamed the Simplenote-AppStore target and build scheme to Simplenote
4. Updated the app icon so the beta icon will appear for debug builds
5. Updated the testing target so that it will run on the renamed AppStore target
6. Updated the fastfile to point to the correct target

Note that there were some signing issues preventing the renamed Appstore target and the testing target from running.  So currently I have set their signing to a personal team and Mac developer.  These will need to get fixed

### Test
- [ ] Confirm that you can build the app with the new scheme
- [ ] Run the tests and confirm they are all green
- [ ] Confirm when you run the app you get the beta icon

### Review

### Release


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

jleandroperez (approved):
✅ Verified the app can be built + runs correctly
✅ Verified that unit tests are green

One minor note @charliescheer, we *may wanna* use a different Bundle Identifier for debug builds, so that we avoid conflicts if you've got installed the AppStore build.

cc @spencertransier would defintely know what's best!!

image


Super okay to address that in a follow up, though.

:shipit:

PR #1159, opened 2024-05-24
Adding Intents extension and required entitlements

charliescheer:
### Fix
So this is another pre work PR to continue preparing for intents/shortcuts on Mac.  In this PR I have done a number of things that should free up the work on that project to continue.  Here I have:
1. I have added an intents extension
2. Updated the bundle names for debug/release on intents and the main app to what they should be
3. I have added an app group entitlement to the intents extension and the main app

With this we should be able to move the database to the app group folder so the intents extension can access it and we should be ready to update the development certificates for the last time.  

I promise right now, Friday at 3pm, this is the last time I will bug the infrastructure team.... this week....

### Test
Nothing to test, make sure the app builds.  Confirm things look correct.  There will need to be some certificate work done to make this all good, but that will happen elsewhere.

### Review
***(Required)*** Add instructions for reviewers.  For example:
> Only one developer is required to review these changes, but anyone can perform the review.

### Release
***(Required)*** Add a concise statement to `RELEASE-NOTES.txt` if the changes should be included in release notes. Include details about updating the notes in this section. For example:
> These changes do not require release notes.

PR #1160, opened 2024-05-24
Shortcuts: Add open new note intent

charliescheer:
### Fix
So the very first shortcut to add is one that simply opens Simplenote and creates a new note.  This is a basic proof of concept that the intents and shortcut handling works.

So in this PR you can create a shortcut "Open New Note" that will launch Simplenote and then create a new note.

Relies on #1159 
### Test
1. Build the app.  
2. Go to the shortcuts app, click on the plus button to add a new shortcut, search for Simplenote, select it, choose Open New Note
3. Run the shortcut.

- [ ] confirm that Simplenote opens and that a new note is created

### Review

### Release


IntentsExtension/Info.plist
```diff
@@ -8,8 +8,12 @@
 		<dict>
 			<key>IntentsRestrictedWhileLocked</key>
 			<array/>
-			<key>IntentsSupported</key>
+			<key>IntentsRestrictedWhileProtectedDataUnavailable</key>
```
jleandroperez, line 11: TIL!

Simplenote/ShortcutHandler.swift
```diff
@@ -0,0 +1,37 @@
...
+            return false
+        }
+
+        switch type {
+        case .newNoteShortcut:
+            SimplenoteAppDelegate.shared().noteEditorViewController.createNote(from: nil)
+        }
+
+        return true
+    }
```
jleandroperez: Suggestion!!

I've noticed the `SimplenoteAppDelegate` builds `ShortcutHandler`, and then `ShortcutHandler` access the `SimplenoteAppDelegate` singleton to check for the authentication state, **and then** to create the note.

Both entities are tightly coupled. A cleaner approach could involve:

1. Dropping the `shared` property (the singleton!)
2. Building a protocol that gets you what you need from the AppDelegate (`isAuthenticated`?)
3. Building the Handler, and passing the reference of the "AuthenticationStatusProvider"
4. Making the AppDelegate conform to this protocol

**Alternatively** (and I'd probably go for this one): all of the `Handler` code could just live in  `SimplenoteAppDelegate+Shortcuts`

jleandroperez (approved):
Added a note on the `ShortcutHandler` component.

Other than that, feel free to :shipit: when ready!

Simplenote/ShortcutHandler.swift
charliescheer: Good call.  I was copying how it is setup in iOS, but there is a lot more going on with the shortcut handler in that app.  I have moved all of it into the app delegate since it is so simple

charliescheer:
This is GTG, just waiting on the PR that this relies on

Simplenote/SimplenoteAppDelegate+Swift.swift
```diff
@@ -246,13 +246,35 @@ extension SimplenoteAppDelegate {
...
+
+    private func handleShortcutActivity(_ userActivity: NSUserActivity) -> Bool {
+        guard let type = ActivityType(rawValue: userActivity.activityType),
+              simperium.user?.authenticated() == true else {
+            return false
+        }
+
+        switch type {
+        case .newNoteShortcut:
+            SimplenoteAppDelegate.shared().noteEditorViewController.createNote(from: nil)
```
jleandroperez: You can drop `SimplenoteAppDelegate.shared()` (because that should be `self`)

jleandroperez (approved):
:shipit:

PR #1161, opened 2024-05-24
Refactor core data

charliescheer:
### Fix
Currently Simplenote Mac sets up its core data properties and management directly in the app delegate.  While this is "fine" it's not great practice, and because we are setting up an intents extension, soon we will want to reuse all that setup code.  So here I have extracted the setup of the model, context, and persistent store so that it is all managed by a single class and can easily be reused for the intents in a future PR.

relies on #1160 
### Test
There isn't much to test here.  In theory this code operates the exact same as the previous app delegate code, but it is in swift and in its own class.

Confirm that you can still load and access your notes in simplenote
Check the code and make sure the objc -> conversion looks right

### Review

### Release
> 
> > Added markdown support


Simplenote/StorageSettings.swift
```diff
@@ -0,0 +1,36 @@
...
+
+class StorageSettings {
+    let fileManager: FileManager
+
+    init(fileManger: FileManager = .default) {
+        self.fileManager = fileManger
+    }
+
+    var applicationFilesDirectory: URL? {
+        let libraryURL = fileManager.urls(for: .libraryDirectory, in: .userDomainMask).last
```
jleandroperez, line 19: I think this property should actually be `userLibraryDirectory` ?

Simplenote/CoreDataManager.swift
```diff
@@ -0,0 +1,87 @@
...
+            NSInferMappingModelAutomaticallyOption: true
+        ]
+        let coordinator = NSPersistentStoreCoordinator(managedObjectModel: model)
+
+        try coordinator.addPersistentStore(ofType: NSXMLStoreType, configurationName: nil, at: storageSettings.storageURL, options: options)
+
+        return coordinator
+    }
+
+    static private func validateResourceValueForDirectory(at url: URL) throws {
```
jleandroperez, line 86: This should probably be private?

Simplenote/CoreDataManager.swift
```diff
@@ -0,0 +1,87 @@
...
+        let properties = try url.resourceValues(forKeys: [URLResourceKey.isDirectoryKey])
+
+        if properties.isDirectory != true {
+            throw CoreDataManagerError.foundNotDirectoryAtFilesDirectoryURL
+        }
+    }
+
+    static private func handleDirectoryError(_ error: NSError, directoryURL: URL) throws {
```
jleandroperez, line 94: This one too!

Simplenote/CoreDataManager.swift
```diff
@@ -0,0 +1,87 @@
...
+        guard let applicationFilesDirectory = storageSettings.applicationFilesDirectory else {
+            throw CoreDataManagerError.noApplicationFilesDirectoryURL
+        }
+
+        // Validate the directory for the store DB
+        do {
+            try Self.validateResourceValueForDirectory(at: applicationFilesDirectory)
+        } catch {
+            try handleDirectoryError((error as NSError), directoryURL: applicationFilesDirectory)
+        }
```
jleandroperez: IMHO L50-L59 should probably live within a single method?

You'd just encapsulate all of the URL logic in a single place, and the "preparePSC" method would probably end up being super compact

```
func persistentStoreURL() throws -> URL {

}
``

**Review by jleandroperez — APPROVED** [review]
Few cosmetic notes, but no blockers!!

Feel free to :shipit: when ready!!

File: Simplenote/CoreDataManager.swift (hunk 2 above)
Comment (charliescheer) [reply Simplenote/CoreDataManager.swift:86]: ... it is....

File: Simplenote/CoreDataManager.swift (hunk 3 above)
Comment (charliescheer) [reply Simplenote/CoreDataManager.swift:94]: so is that one...

**Comment by charliescheer** [comment]
This is ready.  Waiting on some changes from the pr it relies on to merge

File: Simplenote/CoreDataManager.swift [hunk 5]
```diff
@@ -0,0 +1,101 @@
...
+
+        let psc = try Self.preparePSC(with: storageSettings, model: mom)
+
+        self.managedObjectModel = mom
+        self.managedObjectContext = context
+        self.persistentStoreCoordinator = psc
+    }
+
+    static private func preparePSC(with storageSettings: StorageSettings, model: NSManagedObjectModel) throws -> NSPersistentStoreCoordinator {
+        let storageURL = try persistentStoreURL(with: storageSettings)
```
jleandroperez, line 54: 🤟

jleandroperez (approved):
:shipit:

PR #1162, opened 2024-05-29
Migrate database to group container

charliescheer:
### Fix
So to continue with the shortcuts project we need to migrate the core data database from the sandbox container to a group container.  This will make it possible to access the core data database from extensions.

In this PR I have added a migrator to attempt to move the database into the new location on launch.  If this fails it should clean its self up and continue using the old database in its old location.  If it succeeds it should rename the old database and preserve it in place as a backup.

Relies on #1161 

### Test
1. Launch the Mac app and confirm that you are still able to access your notes and tags
- [ ] Confirm in the logs you see `Beginning database migration from:` and `Database migration successful!!`
- [ ] In finder go to `~/Library/Containers/com.automattic.SimplenoteMac.debug/Data/Library/Simplenote/ and confirm the old database is there but the file name is Simplenote.storedata.old
- [ ] in finder go to `~/Library/Group Containers/PZYM8XX95Q.com.automattic.SimplenoteMac.Debug/Data/` and confirm you see Simplenote.storedata

### Review
> Only one developer  is required to review these changes, but anyone can perform the review.

### Release


Simplenote/Bundle+Simplenote.swift
```diff
@@ -11,4 +11,50 @@ extension Bundle {
...
+    static let teamIDPrefix = "TeamIDPrefix"
+    static let appSuffix = ".app"
+    static let defaultTeamID = "PZYM8XX95Q"
+
+    static let defaultBundleID: String  = {
+        var id = "com.automattic.SimplenoteMac"
+        #if DEBUG
+        id.append(".Debug")
+        #endif
+        return id
```
jleandroperez: Nipticky:

For readability, would it be possible to just return the two different strings that are expected?

Simplenote/CoreDataManager.swift
```diff
@@ -35,24 +35,22 @@ class CoreDataManager: NSObject {
     private(set) var managedObjectContext: NSManagedObjectContext
     private(set) var persistentStoreCoordinator: NSPersistentStoreCoordinator
 
-    init(storageSettings: StorageSettings = StorageSettings()) throws {
+    init(at storageURL: URL, storageSettings: StorageSettings = StorageSettings()) throws {
```
jleandroperez: It should probably be up to the `StorageSettings` to return the correct `storageURL`

Simplenote/SharedStorageMigrator.swift
```diff
@@ -0,0 +1,131 @@
...
+                    try migrateCoreDataFiles()
+                    try attemptCreationOfCoreDataStack()
+                    NSLog("Database migration successful!!")
+                    backupLegacyDatabase()
+                    return .success
+                } catch {
+                    NSLog("Could not migrate database to app group " + error.localizedDescription)
+                    removeFailedMigrationFilesIfNeeded()
+                    return .failed
+                }
```
jleandroperez: Nipticky: Indentation!

Simplenote/SharedStorageMigrator.swift
```diff
@@ -0,0 +1,131 @@
...
+            NSLog("Could not delete files from failed migration " + error.localizedDescription)
+        }
+    }
+}
+
+enum MigrationResult {
+    case success
+    case notNeeded
+    case failed
+}
```
jleandroperez: Let's build unit tests for this one! 🔥

Simplenote/SharedStorageMigrator.swift
```diff
@@ -0,0 +1,131 @@
...
+
+                do {
+                    try migrateCoreDataFiles()
+                    try attemptCreationOfCoreDataStack()
+                    NSLog("Database migration successful!!")
+                    backupLegacyDatabase()
+                    return .success
+                } catch {
+                    NSLog("Could not migrate database to app group " + error.localizedDescription)
+                    removeFailedMigrationFilesIfNeeded()
```
jleandroperez: There's a flow that (could happen):

1. For reason `X` `attemptCreationOfCoreDataStack` throws
2. `backupLegacyDatabase` never runs
3. `removeFailedMigrationFilesIfNeeded` also fails

I'm not 100% sure if the `migration required` flag should just be `legacyStorageExists`

Simplenote/SimplenoteAppDelegate+Swift.swift
```diff
@@ -8,8 +8,52 @@ import CoreSpotlight
...
+
+        switch migrationResult {
+        case .notNeeded, .success:
+            try validateStorageDirectory(at: settings.sharedUserLibraryDirectory)
+            coreDataManager = try CoreDataManager(at: settings.sharedStorageURL)
+        case .failed:
+            try validateStorageDirectory(at: settings.legacyUserLibraryDirectory)
+            coreDataManager = try CoreDataManager(at: settings.legacyStorageURL)
+        }
+    }
```
jleandroperez: We've got a bit of duplication here, and we're not really doing anything different if the migration was `notNeeded / successful` or failed (all that changes is the URL!)

A possibility could be to:

1. Have `SharedStorageMigrator` return an instance of `StorageSettings` with the right URL
2. Update `StorageSettings` so that the URL to use is a calculated property in there

Simplenote/SimplenoteAppDelegate+Swift.swift
```diff
@@ -8,8 +8,52 @@ import CoreSpotlight
...
+
+    private func handleDirectoryError(_ error: NSError, directoryURL: URL) throws {
+        if error.code == NSFileReadNoSuchFileError {
+            try FileManager.default.createDirectory(at: directoryURL, withIntermediateDirectories: true, attributes: nil)
+        } else {
+            throw error
+        }
     }
```
jleandroperez, line 30: These 3 last API(s) can probably be simplified as follows **(untested)**:

```
extension FileManager {

    func ensureDirectoryExists(at url: URL) throws {
        if directoryExistsAtURL(url) {
            return
        }

        try createDirectory(at: url, withIntermediateDirectories: true)
    }

/// This can prob. be modernized, borrowed from elsewhere!
    func directoryExistsAtURL(_ url: URL) -> Bool {
        var isDir: ObjCBool = false
        let exists = fileExists(atPath: url.path, isDirectory: &isDir)
        return exists && isDir.boolValue
    }
}
```

jleandroperez (commented):
Looks great @charliescheer !!! sending you a few notes, thank youuu!!

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Simplenote/Bundle+Simplenote.swift
```diff
@@ -11,4 +11,50 @@ extension Bundle {
...
     }
+
+    var rootBundleIdentifier: String {
+        guard isAppExtensionBundle,
+              bundleURL.path.components(separatedBy: "/").contains(where: { $0.hasSuffix(Constants.appSuffix) }) else {
+            return bundleIdentifier ?? Constants.defaultBundleID
+        }
+
+        var url = bundleURL
+        while !url.lastPathComponent.hasSuffix(Constants.appSuffix) {
```
roundhill: I'm wary of this loop, although it's probably pretty much a for-sure thing that we'll always have `.app` in the path. is there a way to trim off the sub dirs after `.app` without the loop?

Simplenote/Bundle+Simplenote.swift
```diff
@@ -11,4 +11,50 @@ extension Bundle {
...
+            url.deleteLastPathComponent()
+        }
+
+        return (Bundle(url: url)?.object(forInfoDictionaryKey: kCFBundleIdentifierKey as String) as? String) ?? Constants.defaultBundleID
+    }
```
jleandroperez, line 31: IMHO perhaps it's just best to return the constant, the url path may have worked for a long time but... :| !!

Simplenote/CoreDataValidator.swift
```diff
@@ -0,0 +1,39 @@
+//
+//  CoreDataValidator.swift
+//  Simplenote
+//
+//  Created by Charlie Scheer on 5/30/24.
+//  Copyright © 2024 Simperium. All rights reserved.
```
roundhill: Hmm probably `Automattic` here :)

Simplenote/CoreDataValidator.swift
```diff
@@ -0,0 +1,39 @@
...
+class CoreDataValidator {
+    func validateStorage(withModelURL modelURL: URL, storageURL: URL) throws {
+        guard let mom = NSManagedObjectModel(contentsOf: modelURL) else {
+            fatalError("Could not load Managed Object Model at path: \(storageURL.path)")
+        }
+
+        try loadPersistentStorage(with: mom, at: storageURL)
+    }
+
+    func loadPersistentStorage(with mom: NSManagedObjectModel, at storageURL: URL) throws {
```
jleandroperez: This API doesn't load, but instead, verifies that it can be loaded. How about something like this?

```
    func validateStorage(model: NSManagedObjectModel, storageURL: URL) throws 
```

Simplenote/SharedStorageMigrator.swift
```diff
@@ -0,0 +1,103 @@
...
+            storageSettings.setStorageLocation(to: .legacy)
+            removeFailedMigrationFilesIfNeeded()
+        }
+    }
+
+    private func migrateCoreDataFiles() throws {
+        if !fileManager.directoryExistsAtURL(storageSettings.sharedUserLibraryDirectory) {
+            try fileManager.createDirectory(at: storageSettings.sharedUserLibraryDirectory, withIntermediateDirectories: true)
+        }
+        try fileManager.copyItem(at: legacyStorageURL, to: sharedStorageURL)
```
roundhill, line 65: Surprisingly simple!

roundhill (commented):
My migration went perfectly 😘 

Had a few minor thoughts that aren't blocking

Simplenote/StorageSettings.swift
```diff
@@ -8,29 +8,72 @@
...
+    var sharedStorageURL: URL {
+        sharedUserLibraryDirectory.appendingPathComponent("\(Constants.modelName).\(Constants.storeExtension)")
+    }
+
+    var legacyStorageURL: URL {
+        legacyUserLibraryDirectory.appendingPathComponent("\(Constants.modelName).\(Constants.storeExtension)")
+    }
+
+    func setStorageLocation(to newLocation: StorageLocation) {
+        storageLocation = newLocation
```
jleandroperez: I was thinking more in the lines of: having the logic, in here, to determine which one should be used (rather than having a parameter), in a dynamic property.

**Just** in case?

jleandroperez (approved):
✅ Verified that running a fresh install gets the storage directly into the shared container
✅ Verified that running this branch (on top of a trunk install) does migrate + backup the old DB
✅ Unti tests look great!! Thank you for adding those!!

Real nice work @charliescheer !!!

:shipit:

Simplenote/Bundle+Simplenote.swift
jleandroperez: +1 yeah this API makes me a bit nervous. Works well, but perhaps picking the string from the project settings would be best!!

Simplenote/CoreDataValidator.swift
charliescheer: Several years ago we had been removing those headers from Simplenote iOS and Mac files... I should get back into that habit.  I dropped it from the new files I have created on this PR

Simplenote/Bundle+Simplenote.swift
charliescheer: I hear you.... but in this project I have updated the bundle ID a few times.  I don't want to create a situation where you might have to remember to update the bundle ID manually.

Simplenote/StorageSettings.swift
charliescheer: done

Simplenote/Bundle+Simplenote.swift
charliescheer: I switched this to a for in loop instead.  If that fails to generate a useable url then we will simply return the default which is manually set

roundhill: Much better, thank you!

PR #1165, opened 2024-05-29
Shortcut: Find Note With Content Intent

charliescheer:
### Fix
Okay, so I can't yet do the intents that update notes, so I am doing the ones that I can work on without access to the shared keychain...

So here we are adding the Find Note with Content intent.  This allows you to enter some text that is contained in your notes and it will then return that note.  On it's own this shortcut intent doesn't do much, but you can connect it to other intents (like open note, or  when I have build it later: append note) this makes it really easy to find specific notes and then updated them using shortcuts.

### Test
1. build and launch Simplenote. Make sure you are logged in
2. Open the shortcuts app, create a new shortcut, find simplenote, select Find Note
3. Right click on the "content" button and choose ask every time
4. In the sidebar of the shortcuts app you should also see "Open Note" select that.  This will automatically connect the output of find note to open note
5. Run the shortcut, when prompted enter some text that you know is in multiple notes.
6. you will be prompted to choose which note you mean, select one
7. confirm the note is opened in simplenote

### Review

### Release
> 
> > Added markdown support


jleandroperez (approved):
:shipit:

PR #1167, opened 2024-05-31
Updated the intents extension version number

charliescheer:
### Fix
The version number on the new intents extension wasn't set correctly.  This PR updates that so that the version number is the same as the main target.

### Test
1. Go to the project navigator in Xcode, select the IntentsExtension -> Build Settings and filter for `Version`
2. Confirm that the marketing version is 2.20 and the project version is 11258

### Review

### Release
> 
> > Added markdown support


jleandroperez (approved):
:shipit:

PR #1168, opened 2024-05-31
Shortcuts: Find note with tag

charliescheer:
### Fix
So one thing users might want to do is to access their notes in shortcuts by seeing what notes are part of different tags.  This PR allows for that.

The shortcut intent will allow users to pick a tag, then simplenote will return all the notes that tag is associated with, and the user can choose between them.

On it's own this shortcut is limited, but connect it to Open Note or Copy Note Content and it becomes quite useful

### Test
1. Run the branch and then open the shortcuts app. Create a new shortcut and look for Simplenote in the apps
2. Find Note with Tag.  Select a tag that you know has notes connected to it
3. Now select Open note from the sidebar.
4. run the shortcut.
5.  Confirm that the note you choose is opened in simplenote

### Review


jleandroperez (commented):
@charliescheer I can't get the `Open Note` to pick up the `Find note with Tag`'s result.

Am I missing something in the setup?

TY!!

image

charliescheer:
It looks like your two shortcut pieces aren't connected.  So you aren't receiving the note from the first intent.... that's odd.  I will give it some testing and see what I can identify may be happening

charliescheer:
Ah yup the intent was not set to return the note that it found, that's why it didn't work.  @jleandroperez can you please try this again?

jleandroperez (approved):
✅ Verified I can now Open the selected Note, after Finding by Tag

:shipit:

PR #1176, opened 2024-06-13
Shortcuts: Recover data if append fails

charliescheer:
### Fix
With the new append shortcuts there is the possibility that your shortcut fails and if that happens the content you sent to the shortcut will be lost.  There isn't a lot we can do to prevent that unfortunately, but we can attempt to save the content locally and then restore it on the next launch.  So in this PR I have added a content recovery mechanism incase a shortcut fails.  That way the user doesn't have to restore the written content if a shortcut fails.

### Test
1. Build and run simplenote mac
2. Open up the shortcuts app and create an append to note or a create note with content shortcut
3. Disable the internet connections
4. run the shortcut
5. In finder go to ~/Library/Group Containers/PZYM8XX95Q.com.automattic.SimplenoteMac.Development/Recovery and confirm you have a file there
6. launch simplenote.

 - [ ] Confirm that a new note is created with the content from your shortcut
 - [ ] Confirm the file in the recovery folder was deleted.

### Review


Simplenote/ContentRecoveryManager.swift
```diff
@@ -0,0 +1,80 @@
...
+    public init(fileManager: FileManager = .default) {
+        self.fileManager = fileManager
+    }
+
+    private func createRecoveryDirIfNeeded() {
+        guard !fileManager.directoryExistsAtURL(fileManager.recoveryDirectoryURL) else {
+            return
+        }
+
+        try? fileManager.createDirectory(at: fileManager.recoveryDirectoryURL, withIntermediateDirectories: true)
```
jleandroperez: (Same note as in iOS), this could be a FileManager extension, and you'd avoid the need to call it in multiple spots. Sample!

```
extension FileManager {

    var entryUploadCacheURL: URL {
        let output = syncCachesURL.appendingPathComponent(FileManagerFolders.uploadCache, isDirectory: true)
        if directoryExistsAtURL(output) == false {
            try? createDirectory(at: output, withIntermediateDirectories: true)
        }

        return output
    }
```

Simplenote/ContentRecoveryManager.swift
```diff
@@ -0,0 +1,80 @@
...
+        guard let data = fileManager.contents(atPath: url.path),
+              let recoveredContent = String(data: data, encoding: .utf8) else {
+            return nil
+        }
+
+        var content = Constants.recoveredContentHeader
+        content += "\n\n"
+        content += recoveredContent
+        return content
+    }
```
jleandroperez: `context` isn't used here

Simplenote/ContentRecoveryManager.swift
```diff
@@ -0,0 +1,80 @@
...
+
+    private func url(for identifier: String) -> URL {
+        let formattedID = identifier.replacingOccurrences(of: "/", with: "-")
+        let fileName = "\(Constants.recoveredContent)-\(formattedID)"
+        return fileManager.recoveryDirectoryURL.appendingPathComponent(fileName, conformingTo: UTType.json)
+    }
+
+    // MARK: Restore
+    //
+    public func prepareRecoveredNoteContentIfNeeded(in context: NSManagedObjectContext) async -> [String] {
```
jleandroperez: `context` can be removed (since it's not being used)

Simplenote/SimplenoteAppDelegate+Swift.swift
```diff
@@ -148,6 +148,28 @@ extension SimplenoteAppDelegate {
         // TODO: Temporary workaround. Let's get rid of this? please? 🔥🔥🔥
         mainWindowController.window as! Window
     }
+
+    @objc
+    func attemptContentRecoveryIfNeeded() {
+        let context = NSManagedObjectContext(concurrencyType: .privateQueueConcurrencyType)
+        context.persistentStoreCoordinator = coreDataManager.persistentStoreCoordinator
```
jleandroperez: `context` isn't used

Simplenote/SimplenoteAppDelegate+Swift.swift
```diff
@@ -148,6 +148,28 @@ extension SimplenoteAppDelegate {
...
+
+        Task {
+            let restoredContent = await ContentRecoveryManager().prepareRecoveredNoteContentIfNeeded(in: context)
+            restoredContent.forEach({ insertNote(with: $0) })
```
jleandroperez: Threading problem: `insertNote` uses the Simperium API, which isn't threadsafe.

Simplenote/SimplenoteAppDelegate+Swift.swift
```diff
@@ -148,6 +148,28 @@ extension SimplenoteAppDelegate {
...
+    private func insertNote(with content: String) {
+        guard let note = simperium.notesBucket.insertNewObject() as? Note else {
+            return
+        }
+        note.modificationDate = Date()
+        note.creationDate = Date()
+        note.content = content
+        note.markdown = UserDefaults.standard.bool(forKey: "kMarkdownPreferencesKey")
+        simperium.save()
+    }
```
jleandroperez, line 155: This should probably be done in the ContentRecoveryManager?

Simplenote/SimplenoteAppDelegate+Swift.swift
```diff
@@ -148,6 +148,28 @@ extension SimplenoteAppDelegate {
...
+
+    private func insertNote(with content: String) {
+        guard let note = simperium.notesBucket.insertNewObject() as? Note else {
+            return
+        }
+        note.modificationDate = Date()
+        note.creationDate = Date()
+        note.content = content
+        note.markdown = UserDefaults.standard.bool(forKey: "kMarkdownPreferencesKey")
```
jleandroperez: You can use the `SPMarkdownPreferencesKey` constant` here

jleandroperez (commented):
Same note as in iOS, we'd probably need to wire `attemptContentRecoveryIfNeeded` in `applicationWillBecomeActive`, otherwise payload won't be recovered until you effectively relaunch.

Thanks for building this Charlie!!

Simplenote/SimplenoteAppDelegate+Swift.swift
```diff
@@ -148,6 +148,11 @@ extension SimplenoteAppDelegate {
         // TODO: Temporary workaround. Let's get rid of this? please? 🔥🔥🔥
```
jleandroperez, line 148: 🤣  *temporary*

IntentsExtension/IntentHandlers/AppendNoteIntentHandler.swift
```diff
@@ -27,14 +27,24 @@ class AppendNoteIntentHandler: NSObject, AppendNoteIntentHandling {
...
-        note.content = existingContent + "\n\(content)"
         let uploader = Uploader(simperiumToken: token)
-        uploader.send(note)
+        do {
+            _ = try await uploader.send(note)
+            return AppendNoteIntentResponse(code: .success, userActivity: nil)
+        } catch {
+            return handleFailure(with: error, content: content)
+        }
+    }
```
jleandroperez, line 39: There's a bit of duplication in this API. You don't **really** need two Try / Catch (since if any of them both fails, you'll bail out.

Alternative A is something like this:

```
        // ...

        do {
            let downloader = Downloader(simperiumToken: token)
            let existingContent = try await downloader.getNoteContent(for: identifier)
            note.content = existingContent + "\n\(content)"

            let uploader = Uploader(simperiumToken: token)
            try await uploader.send(note)
        } catch {
            return handleFailure(with: error, content: content)
        }

        return AppendNoteIntentResponse(code: .success, userActivity: nil)
    }
```

Idea B is to just extract the `download` + `upload` API(s) into smaller methods, and it would look like...

```
        // ...

        do {
            let existingContent = try await downloadNoteContent(for: identifier, simperiumToken: token)
            note.content = existingContent + "\n\(content)"

            try await uploadNoteContent(note, simperiunToken: token)
        } catch {
            return handleFailure(with: error, content: content)
        }

        return AppendNoteIntentResponse(code: .success, userActivity: nil)
    }
```

IntentsExtension/Tools/Uploader.swift
```diff
@@ -28,11 +28,8 @@ class Uploader: NSObject {
         request.setValue(token, forHTTPHeaderField: Settings.authHeader)
 
         // Task!
-        let sc = URLSessionConfiguration.backgroundSessionConfigurationWithRandomizedIdentifier()
```
jleandroperez, line 31: I don't really remember why this had a randomized ID, to begin with?

Simplenote.xcodeproj/project.pbxproj
```diff
@@ -2834,7 +2842,7 @@
 					"@executable_path/../../../../Frameworks",
 				);
 				LOCALIZATION_PREFERS_STRING_CATALOGS = YES;
-				MACOSX_DEPLOYMENT_TARGET = 14.4;
+				MACOSX_DEPLOYMENT_TARGET = 12.0;
```
jleandroperez, line 2845: Glad you caught this one!!

Simplenote/RecoveryUnarchiver.swift
```diff
@@ -0,0 +1,61 @@
...
+
+public class RecoveryUnarchiver {
+    private let fileManager: FileManager
+
+    public init(fileManager: FileManager = .default) {
+        self.fileManager = fileManager
+    }
+
+    var simperium: Simperium {
+        SimplenoteAppDelegate.shared().simperium
```
jleandroperez: Since the AppDelegate will use this class, and it'll have access to the Simperium instance, you can get this via initializer

jleandroperez (approved):
Thank you for building this one Charlie!!

PR #1185, opened 2024-07-02
Add `Base.xcconfig` and push most code signing settings there

mokagio:
### Fix

This was prompted by some changes in the app identifiers (plural because we have one for release and one for debug) that required regenerating the provisioning profiles.

Internal ref pdnsEh-1Kz-p2

Adds a `Base.xcconfig` file and moves code signing settings there, so they are DRY and we don't need to deal with Xcode adding many lines to the project file every time we tweak them.

The only setting that stayed in the project itself (i.e. we'd need to use the Build Settings GUI to modify it) is `PROVISIONING_PROFILE_SPECIFIER` because this is a per-target setting and we don't have that level of granularity in the `xcconfig`s

### Test
You should be able to run a development build successfully after having downloaded a new development bundle from our Mission Control (Automatticians only).

I run a build locally in release mode to verify that mode builds, too.
image

### Review
Only one developer or infra engineer required to review these changes, but anyone can perform the review.

### Release

### DIFF
--- a/config/Simplenote.debug.xcconfig
+++ b/config/Simplenote.debug.xcconfig
@@ -1,1 +1,3 @@
-#include "Version.public.xcconfig"
+#include "Base.xcconfig"
+
+CODE_SIGN_IDENTITY = Apple Development
--- a/config/Simplenote.debug.xcconfig
+++ b/config/Simplenote.debug.xcconfig
@@ -1,3 +1,6 @@
 #include "Base.xcconfig"
 
 CODE_SIGN_IDENTITY = Apple Development
+
+// Notice we don't have a PROVISIONING_PROFILE_SPECIFIER defined here because this xcconfig is used at the project-level but that settings need to be target-level.
+// Once we'll have target-level xcconfigs, we shall add the setting there
--- a/config/Simplenote.release.xcconfig
+++ b/config/Simplenote.release.xcconfig
@@ -1,1 +1,3 @@
-#include "Version.public.xcconfig"
+#include "Base.xcconfig"
+
+CODE_SIGN_IDENTITY = Apple Distribution
--- a/config/Simplenote.release.xcconfig
+++ b/config/Simplenote.release.xcconfig
@@ -1,3 +1,6 @@
 #include "Base.xcconfig"
 
 CODE_SIGN_IDENTITY = Apple Distribution
+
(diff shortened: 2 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Message">
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">This PR is still a Draft: some checks will be skipped.</td>
    </tr>
  </tbody>
</table>

</p>

AliSoftware (approved):
Reviewed the code changes / diff and LGTM, thought disclaimer that I haven't tested it in Xcode; I'll let a SN dev check it works as expected with their setup (eg even if they don't have release certs or ASC account in their Xcode like we AppsInfra might have)

> The only setting that stayed in the project itself […] is `PROVISIONING_PROFILE_SPECIFIER` because this is a per-target setting and we don't have that level of granularity in the xcconfigs

Could be worth adding a `// comment` in the `.xcconfig` to mention that? (until we have time to create per-target xcconfig files in the future to solve that properly 😛)

config/Simplenote.debug.xcconfig
```diff
@@ -1 +1,3 @@
-#include "Version.public.xcconfig"
+#include "Base.xcconfig"
+
+CODE_SIGN_IDENTITY = Apple Development
```
mokagio, line 3:
```suggestion
CODE_SIGN_IDENTITY = Apple Development

// Notice we don't have a PROVISIONING_PROFILE_SPECIFIER defined here because this xcconfig is used at the project-level but that settings need to be target-level.
// Once we'll have target-level xcconfigs, we shall add the setting there
```

config/Simplenote.release.xcconfig
```diff
@@ -1 +1,3 @@
-#include "Version.public.xcconfig"
+#include "Base.xcconfig"
+
+CODE_SIGN_IDENTITY = Apple Distribution
```
mokagio, line 3:
```suggestion
CODE_SIGN_IDENTITY = Apple Distribution

// Notice we don't have a PROVISIONING_PROFILE_SPECIFIER defined here because this xcconfig is used at the project-level but that settings need to be target-level.
// Once we'll have target-level xcconfigs, we shall add the setting there
```

PR #1193, opened 2024-07-19
Magic Links: Add SPNavigationController

charliescheer:
### Fix
For the magic links setup to work we need to reshape how the auth view looks.  Right now it is a single screen to sign in or an email me a link screen which has a button on it for showing the password field.  This is fine for now, but doesn't really work with the new layout designs that we are putting into the other apps which has a series of screens depending on how the user wants to login and where they are in the process.

To be able to manage the transition from one screen to another we need some sort of view management system.  Natively MacOS doesn't support this, so in this PR I have created a new view controller that can act similar to an iOS navigation controller.  This will permit us to travel through a stack of views and return back to them if need be.

In this PR I have not made any changes to how the ui currently works except that we are pushing to the email view from sign up instead of cross fading.

Note that the sizes between the sign up and login views are a little off so the animation is a bit wonky, but that is a problem for a future PR this is just about getting the views transitioning.

Open to suggestions and feedback on how the animation works.  Currently for a push the new view slides in and the old view fades out.  For a pop the current view slides out and the next view fades in.

### Test
1. Launch SNMac, log out if you are logged in. 
2. From the sign up screen tap on the Login button.
- [ ] Confirm that the email auth view slides 
- [ ] Confirm that a back button appears
- [ ] confirm that the signup view fades away
- [ ] Tap the back button and confirm the animation happens in reverse and you are left with the sign up view

### Review

### Release

### DIFF
--- a/Simplenote/AuthViewController+Swift.swift
+++ b/Simplenote/AuthViewController+Swift.swift
@@ -164,6 +164,20 @@ extension AuthViewController {
         
         performSelector(onMainThread: secondaryActionSelector, with: nil, waitUntilDone: false)
     }
+
+    @IBAction
+    func switchAuthenticationMode(_ sender: Any) {
+        containingNavigationController?.push(nextViewController())
+    }
+
+    private func nextViewController() -> AuthViewController {
+        let nextMode = mode.nextMode()
+
+        let nextVC = AuthViewController()
+        nextVC.mode = nextMode
+
+        return nextVC
+    }
 }
 
 
--- a/Simplenote/AuthViewController+Swift.swift
+++ b/Simplenote/AuthViewController+Swift.swift
@@ -176,6 +176,7 @@ extension AuthViewController {
         let nextMode = mode.nextMode()
 
         let nextVC = AuthViewController()
+        nextVC.authenticator = authenticator
         nextVC.mode = nextMode
 
         return nextVC
--- a/Simplenote/AuthViewController+Swift.swift
+++ b/Simplenote/AuthViewController+Swift.swift
@@ -101,6 +101,8 @@ extension AuthViewController {
         passwordField.alphaValue                = mode.passwordFieldAlpha
         secondaryActionButton.alphaValue        = mode.secondaryActionFieldAlpha
         wordPressSSOButton.alphaValue           = mode.wordPressSSOFieldAlpha
+
+        switchAuthenticationView.isHidden                 = !mode.isSwitchVisible
     }
 
     /// Animates Visible / Invisible components, based on the specified state
--- a/Simplenote/SPNavigationController.swift
+++ b/Simplenote/SPNavigationController.swift
@@ -3,7 +3,7 @@ import Foundation
 class SPNavigationController: NSViewController {
     private var viewStack: [NSViewController] = []
 
-    private var backButton: NSButton = NSButton(title: String(), image: NSImage(named: NSImage.goBackTemplateName)!, target: nil, action: #selector(backWasPressed))
+    private var backButton: NSButton!
 
     var initialViewController: NSViewController {
         guard let first = viewStack.first else {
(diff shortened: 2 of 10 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Simplenote/AuthViewController+Swift.swift
```diff
@@ -164,6 +166,21 @@ extension AuthViewController {
...
         performSelector(onMainThread: secondaryActionSelector, with: nil, waitUntilDone: false)
     }
+
+    @IBAction
+    func switchAuthenticationMode(_ sender: Any) {
+        containingNavigationController?.push(nextViewController())
+    }
+
+    private func nextViewController() -> AuthViewController {
+        let nextMode = mode.nextMode()
```
jleandroperez, line 176: I know part of this is legacy, but it may not fly far. 

AFAIK you should be able to jump into more than one mode (with the new design), but we may revisit elsewhere!

Simplenote/SPNavigationController.swift
```diff
@@ -0,0 +1,184 @@
...
+            spacerView.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            spacerView.heightAnchor.constraint(equalToConstant: 30),
+            spacerView.topAnchor.constraint(equalTo: view.topAnchor)
+        ])
+
+        view.addSubview(backButton)
+        NSLayoutConstraint.activate([
+            backButton.leadingAnchor.constraint(equalTo: view.leadingAnchor, constant: 10),
+            backButton.topAnchor.constraint(equalTo: spacerView.bottomAnchor),
+            backButton.widthAnchor.constraint(equalToConstant: 50)
```
jleandroperez: Magic numbers in L51, L57 and here

Simplenote/SPNavigationController.swift
```diff
@@ -0,0 +1,184 @@
...
+
+        view.addSubview(initialView)
+
+        NSLayoutConstraint.activate([
+            initialView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            initialView.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            initialView.topAnchor.constraint(equalTo: backButton.bottomAnchor),
+            initialView.bottomAnchor.constraint(equalTo: view.bottomAnchor)
+        ])
+    }
```
jleandroperez, line 71: For readability we could prob. split this one into a couple smaller methods perhaps?

Simplenote/SPNavigationController.swift
```diff
@@ -0,0 +1,184 @@
...
+
+    // MARK: - Add a View to the stack
+    //
+    func push(_ viewController: NSViewController, animated: Bool = true) {
+        guard let currentView = topViewController?.view else {
+            return
+        }
+        attach(child: viewController)
+        currentView.removeConstraints(currentView.constraints)
```
jleandroperez: (It **should not happen**) but if the `topVC` view was nil, you'd essentially be unable to push a new VC. 
How about just dealing with `currentView` being an optional in L84 and L97?

Simplenote/SPNavigationController.swift
```diff
@@ -0,0 +1,184 @@
...
+        }
+    }
+
+    private func attach(child: NSViewController) {
+        addChild(child)
+        viewStack.append(child)
+    }
+
+    @discardableResult
+    private func attachView(subview: NSView, behindCurrent: Bool = false) -> (leading: NSLayoutConstraint, trailing: NSLayoutConstraint)? {
```
jleandroperez: You can get the currentView in L83, and it'd simplify things a tiny bit perhaps?

Simplenote/SPNavigationController.swift
```diff
@@ -0,0 +1,184 @@
...
+    // MARK: - Animation
+    //
+    private enum AnimationDirection {
+        case leadingToTrailing
+        case trailingToLeading
+    }
+
+    private func animateTransition(slidingView: NSView, fadingView: NSView, direction: AnimationDirection, onCompletion: @escaping () -> Void) {
+        guard let leadingConstraint = view.constraints.first(where: { $0.firstItem as? NSView == slidingView && $0.firstAttribute == .leading }),
+              let trailingConstraint = view.constraints.first(where: { $0.firstItem as? NSView == slidingView && $0.firstAttribute == .trailing }) else {
```
jleandroperez: This is a great candidate to extract:

```
extension NSView {

   func findConstraint(firstView: NSView, firstAttribute: ...) -> NSView?
}
```

jleandroperez:
✅ Code looks great
✅ Works great too!!!

One tiny detail. Notice what happens if you click on `Log In`:

Notice that the Window Height grows as you drill down, but doesn't shrink when you go back

https://github.com/user-attachments/assets/0ef70308-9193-40dd-baba-f85a3b548a94

jleandroperez (approved):
:shipit:

Simplenote/AuthViewController+Swift.swift
charliescheer: Saving that for the next PR when we start remaking the views

charliescheer:
> Notice that the Window Height grows as you drill down, but doesn't shrink when you go back

This is happening because the window has a set height when it is created and is not constrained to the views, so when it shrinks the view doesn't get smaller.  We are discussing changes to the views them selves which will fix this issue for the login.  This _shouldn't_ be a cause of the navigation controller being added here.  Should be able to fix this in the next PR for the auth view

PR #1195, opened 2024-07-24
Magic Links: Refactor Authentication Mode to handle auth view definitions

charliescheer:
### Fix
Primarily this PR doesn't do much that is visible, but it is laying the ground work to continuing updating the auth views to the new designs and layouts, so we can add magic links.  

So in this PR I have:
- Added defining the buttons on different views using an action descriptor struct
- Updated the existing views to use the new action descriptor structs
- dropped existing primary/secondary/wordpress properties and definitions from Auth mode

This shouldn't look too much different from the current state of the auth views, but because so many things have moved around here we should confirm that it all still works.

### Test
1. Launch Simplenote Mac
2. Confirm that the launch screen now has no fields on it and looks more like the iOS launch screen
3. Tap on sign up and confirm that the screen that appears is for signup.  Test that you can create an account
4. Tap on login and confirm you are given the option to login with email.  Test that login still works
5. Confirm on the login screen that you can tap on the secondary button to reveal the password field.  Confirm you can login with a password
6. Confirm on the login screen that the Wordpress SSO button is there and that if you tap on it the Wordpress sign on appears in your browser

### Review

### DIFF
--- a/Simplenote/AuthViewController+Swift.swift
+++ b/Simplenote/AuthViewController+Swift.swift
@@ -60,6 +60,12 @@ extension AuthViewController {
     var authWindowController: AuthWindowController? {
         view.window?.windowController as? AuthWindowController
     }
+
+    /// # All of the Action Views
+    ///
+    private var allActionViews: [NSButton] {
+        [actionButton, secondaryActionButton]
+    }
 }
 
 // MARK: - Refreshing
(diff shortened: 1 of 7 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

jleandroperez (approved):
Looking good!! For the record, few details in here, which have been already fixed in the follow up PR, so let's move this one along!!

:shipit:

Issue #1187, opened 2024-07-11
Magic Links: Switch to Code Authentication


jleandroperez:
### Details:
- [x] Implement macOS Navigation Controller #1193
- [x] Polish NagitationController Animations #1196
- [x] Refactor AuthUI to handle State and Mode Definitions #1195
- [x] Implement a new Code and Password Interfaces #1197
- [x] Ensure if you get rate limited, during login, you may proceed to the Enter Password UI
- [x] Verify the Magic Link Flow is still supported
- [x] Verify the Invalid Token UI is still triggered https://github.com/Automattic/simplenote-macos/pull/1198
- [x] Adjust UI Details https://github.com/Automattic/simplenote-macos/pull/1198
   - [x] Enter Code TextField should probably be readable (since it's easy to miss, and it expires quickly, anyways)
   - [x] Going Back from the Enter Code UI results in the "Enter Email" UI, with your email highlighted
   - [x] Enter password button's textColor doesn't match the border
   - [x] Enter Password UI needs a header, too, with the account's email (recently added in iOS).
- [x] Update Error Handling: Invalid and Expired Code Errors
- [x] Fix weird back button animation
- [x] Animate Window Resize

dangermattic:
Thanks for reporting! 👍

PR #1196, opened 2024-07-24
Navigation Bar: Polishing Animation

jleandroperez:
### Fix
In this PR we're polishing the Navigation Bar UX:

- `Back` will no longer shrink its height. This was resulting in a different top padding in the initial UI
- Polished the resize animation, when either increasing or reducing the height
- Simplified a tiny bit a couple API(s)

Ref. #1187 

### Test
1. Fresh install Simplenote macOS

- [x] Verify that if you click on `Log In`, the animation looks great
- [x] Verify that if you click on the `Back` button, there are no jumps, and the animation feels on spot
- [x] Verify that if you go back to `Log In` and click on `Continue with password`, everything looks good
- [x] Verify that if you click on `Back`, you end up in the initial UI

### DIFF
--- a/Simplenote/SPNavigationController.swift
+++ b/Simplenote/SPNavigationController.swift
@@ -85,8 +85,12 @@ class SPNavigationController: NSViewController {
         let currentView = topViewController?.view
 
         attach(child: viewController)
-        if let currentView {
-            currentView.removeConstraints(currentView.constraints)
+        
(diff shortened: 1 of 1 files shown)


charliescheer (approved):
👨‍🍳 💋

jleandroperez:
Thank you Charlie!!

PR #1202, opened 2024-07-26
Authentication: Addressing Debt

jleandroperez:
### Fix
In this PR we're addressing pending tech debt, in the AuthenticationViewController class.

### Test
Please smoke test Auth with Password + Code, and verify everything looks great!!

PR #1204, opened 2024-07-26
Magic Links: Fix window resizing animation + some clean up

charliescheer:
### Fix
We had noticed some wonky animations when pushing and popping from one view to another during the auth login/signup flow.  This was causing the size of the window to jump to meet the new view size.  Didn't look great.  In this PR we have updated the animations so that it will animate smoothly to the new size of the view loaded into the navigation controller.  Huge T/Y to @jleandroperez for the assist on this one.

Also while I was in there I fixed an issue where if you don't animate the navigation controller push/pop the views didn't size and the back button didn't hide correctly (we don't really need this, but if we are gonna build the component might as well make it work correctly)

 and I did a bit of cleanup to make the methods simpler.

relies on #1202 

### Test
1. Launch Simplenote Mac to the onboarding/auth screen.  
2. Tap on the sign up button, then the back button, then the login button, then the back button again.

- [ ] Confirm that the views animate smoothly
- [ ] Confirm that the back button appears when you go off the initial view and disappears when you return to it.
- [ ] Confirm the size of the window changes to match the new view that appears.

### Review

### Release

### DIFF
--- a/Simplenote/SPNavigationController.swift
+++ b/Simplenote/SPNavigationController.swift
@@ -17,7 +17,7 @@ class SPNavigationController: NSViewController {
     
     private var heightConstraint: NSLayoutConstraint!
     private var totalTopPadding: CGFloat {
-        Constants.buttonViewLeadingPadding + Constants.buttonViewHeight
+        Constants.buttonViewTopPadding + Constants.buttonViewHeight
     }
 
     init(initialViewController: NSViewController) {
--- a/Simplenote/SPNavigationController.swift
+++ b/Simplenote/SPNavigationController.swift
@@ -95,13 +95,6 @@ class SPNavigationController: NSViewController {
         let currentView = topViewController?.view
 
         attach(child: viewController)
-        
-        /// Disable Bottom Constraint
-        /// This allows for the enclosing NSWindow to resize, just enough to fit the `nextViewController.view`
-        ///
-        if let currentView, let bottomConstraint = view.firstContraint(firstView: currentView, firstAttribute: .bottom) {
-            bottomConstraint.isActive = false
-        }
 
         guard let (leadingAnchor, trailingAnchor) = attachView(subview: viewController.view, below: currentView) else {
             return
@@ -162,8 +155,7 @@ class SPNavigationController: NSViewController {
         NSLayoutConstraint.activate([
             leadingAnchor,
             trailingAnchor,
-            subview.topAnchor.constraint(equalTo: backButton.bottomAnchor) //,
-//            subview.bottomAnchor.constraint(equalTo: view.bottomAnchor)
+            subview.topAnchor.constraint(equalTo: backButton.bottomAnchor)
         ])
 
         NSLog("# After Window Height \(view.window?.frame.height.description) - size \(subview.intrinsicContentSize) - \(subview.fittingSize)")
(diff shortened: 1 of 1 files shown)


Simplenote/SPNavigationController.swift
```diff
@@ -91,25 +95,14 @@ class SPNavigationController: NSViewController {
...
-        }
-
-        guard let (leadingAnchor, trailingAnchor) = attachView(subview: viewController.view, below: currentView) else {
-            return
-        }
+        attachView(subview: viewController.view, below: currentView, animated: animated)
 
         guard animated else {
+            currentView?.removeFromSuperview()
+            backButton.animator().isHidden = hideBackButton
```
jleandroperez:
```suggestion
            backButton.isHidden = hideBackButton
```

Simplenote/SPNavigationController.swift
```diff
@@ -120,45 +113,56 @@ class SPNavigationController: NSViewController {
...
+
+        NSAnimationContext.runAnimationGroup { context in
+            context.duration = 0.4
+            context.timingFunction = .init(name: .easeInEaseOut)
+
+            heightConstraint?.animator().constant = finalHeight
+        }
 
-        return (leading: leadingAnchor, trailing: trailingAnchor)
+        return
```
jleandroperez: No return needed here

Simplenote/SPNavigationController.swift
```diff
@@ -120,45 +113,56 @@ class SPNavigationController: NSViewController {
...
-        if let currentBottomConstraint = view.firstContraint(firstView: currentViewController.view, firstAttribute: .bottom) {
-            currentBottomConstraint.isActive = false
-        }
   
-        attachView(subview: nextViewController.view, below: currentViewController.view)
-        
+        attachView(subview: nextViewController.view, below: currentViewController.view, animated: animated)
+
+        guard animated else {
+            self.dettach(child: currentViewController)
```
jleandroperez:
```suggestion
            dettach(child: currentViewController)
```

Simplenote/SPNavigationController.swift
```diff
@@ -24,21 +26,23 @@ class SPNavigationController: NSViewController {
...
-
+    
     override func loadView() {
         guard let initialViewController = topViewController else {
             fatalError()
         }
 
         view = NSView()
+        heightConstraint = view.heightAnchor.constraint(equalToConstant: .zero)
+        heightConstraint?.isActive = true
```
jleandroperez: Since we're now doing this, the `minimumHeightConstraint` workaround isn't required anymore. We can:

- Maybe convert this constraint into an IUO? or not?
- Move the `.activate` to L57 instead (where `miminumHeightConstraint` lives now)

**Also** I'm ... slightly worried about setting it to `.zero` by default. It is possible, at some point in time, the Window will actually be `.zero` height.

Simplenote/SPNavigationController.swift
```diff
@@ -120,45 +113,56 @@ class SPNavigationController: NSViewController {
...
+            heightConstraint?.constant = finalHeight
+            return
+        }
+
+        NSAnimationContext.runAnimationGroup { context in
+            context.duration = 0.4
+            context.timingFunction = .init(name: .easeInEaseOut)
+
+            heightConstraint?.animator().constant = finalHeight
+        }
```
jleandroperez, line 143: Do you think we could decouple `attach` + `resizeWindow` ?

And then, if we needed, have a third API that calls them both?

```
func attachView(subview: NSView, below siblingView: NSView?, animated: Bool) 
func resizeWindowHeight(for subview: NSView, animated: Bool
```

And then...

```
func attachViewAndResizeWindow(subview: NSView, below siblingView: NSView?, animated: Bool) 
```

jleandroperez: **Reason** why this caught my eye is: the `animation` is only applicable to the `Window Resize`, not to the View we're pushing, right?

**ALSO** there's the `animateTransition` API, so... you can `attach(..., animated: true)`, but it won't animate the transition, just the Window Resize

jleandroperez (commented):
Works great!! sending you a few notes, TY!!

jleandroperez (approved):
Looks amazing, thank you Charlie!!

:shipit:

PR #1216, opened 2024-10-19
Delete note if the content is completely removed

### LINKED ISSUE #1064: Automatically delete a note when it's content is empty

charliescheer:
### Fix
In issue https://github.com/Automattic/simplenote-macos/issues/1064 it was brought up that if a note is empty we leave it in the note list and it would be nice if they deleted themselves. I thought about this for a bit and decided that I agreed, but I wanted to make sure that we only moved the notes to the trash, that way the history for the note still exists and a user could reclaim the lost content if they wanted. So in this PR I have done that.

iOS already deleted new notes that were empty, so it made sense to me that we should have the same behavior in SNMac, and that we should update it to also delete notes that have become empty, not just new notes.

fixes #1064 

#### Confirm new notes with content are not deleted ####
1. Create a new note and add some content to it
2. Close the editor
- [ ] Confirm your new note continues to exist
- [ ] Relaunch the app and confirm your new note was saved to the DB

#### Confirm new notes without content are deleted ####
1. Create a new note, do not add content to it
2. Close the editor
- [ ] Confirm that your new note was deleted

#### Confirm an existing note that has all it's content removed, will be deleted ####
1. Go to a note you don't mind losing.
2. Delete all of the text content in the note
3. Exit the editor.
- [ ] Confirm that your note is deleted from the notes list

4. Go to the note trash, find the top note labeled "New Note..." restore it
5. Go to the regular notes list, open the note you restored, go to the notes history
- [ ] Confirm you can restore that notes history

### Review

### Release

### DIFF
--- a/Simplenote/NSTableView+Simplenote.swift
+++ b/Simplenote/NSTableView+Simplenote.swift
@@ -76,3 +76,13 @@ extension NSTableView {
         }
     }
 }
+
+extension NSTableView {
+    static var currentRowSelectionUserInfoKey: String {
+        "NSTableViewCurrentRowSelectionUserInfoKey"
+    }
+
+    static var previousRowSelectionUserInfoKey: String {
+        "NSTableViewPreviousRowSelectionUserInfoKey"
+    }
+}
--- a/Simplenote/NoteEditorViewController+Swift.swift
+++ b/Simplenote/NoteEditorViewController+Swift.swift
@@ -1,4 +1,5 @@
 import Foundation
+import CoreSpotlight
 import SimplenoteFoundation
 import SimplenoteInterlinks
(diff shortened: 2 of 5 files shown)


charliescheer:
After discussing this we have decided not to make these changes

PR #1233, opened 2025-03-01
Fixed issue where printing from light mode only displayed the first page

### LINKED ISSUE #1232: Print: only first page prints properly (subsequent pages blank)
I had a print issue today using Simplenote app version 2.21 on macOS (Sequoia 15.3.1, iMac Intel).

### Expected

To print all pages in a note when I use "Note > Print" menu item or `Cmd-P` keyboard shortcut.

### Observed

Only the first page contains text.

The print preview, PDF output, and paper printed all have blank pages from page 2+ and onward.

### Reproduced

1. Open Simplenote.app
2. Highlight a note
3. Print using keyboard: Cmd-P or using a mouse: Note > Print menu
4. Only the first page prints—rest is blank!

### Notes

Printing from web app https://app.simplenote.com/ works fine—this is for the Mac app only.

### Screenshots

Image

Image

charliescheer:
### Fix
Fixes #1232 

Interesting issue.  So on SNMac if you use light mode (if people still do such things) and you try to print out a multiple page note you will receive the first page and only the first page, but the document will be long enough to house the entire document...  Just after the first page there is no content.  More weirdly this does not happen if your computer is in dark mode.  Very odd.

Turns out the way that we were calculating the size of the print view was based on the print page size and that cut off the size of the view.  It seems that when in dark mode MacOS is doing some helpful translations to the print info which fix the size of the document and so you get all the details.  weird.  

In this PR I have changed the size of the document so that it will work with the size of the content if it is bigger than one page.

### Test
1. Open simplenote Mac, put your computer display into light mode
2. Open a note that would print out to longer than one page.  Attempt to print
3. Confirm in the preview that you see all pages of the text content.  
4. Print the doc to pdf and confirm you see all the pages of the text content
5. Try steps 2-4 again on a short note and confirm the note is still printing the correct size

### Review

### Release
> `RELEASE-NOTES.txt` was updated in e8c9bd with:
> 
> > - Fixed a bug when printing multiple pages when in light mode

### DIFF
--- a/RELEASE-NOTES.txt
+++ b/RELEASE-NOTES.txt
@@ -1,6 +1,7 @@
 2.23
 -----
 - Stop optional note processing for very long notes
+- Fixed a bug when printing multiple pages when in light mode
 
 2.22
 -----
--- a/Simplenote/NoteEditorViewController.m
+++ b/Simplenote/NoteEditorViewController.m
@@ -428,7 +428,9 @@ static NSString * const SPTextViewPreferencesKey        = @"kTextViewPreferences
 - (IBAction)printAction:(id)sender
 {
     // Create a copy of the editor view to be used as the print source
-    NSTextView *printView = [[NSTextView alloc] init];
+    // NOTE: forcing the print view into Text Kit 1 because TK2 was not printing multiple pages correctly
+    // ref: https://github.com/Automattic/simplenote-macos/issues/1232
+    NSTextView *printView = [[NSTextView alloc] initUsingTextLayoutManager:NO];
     [printView.textStorage appendAttributedString:self.noteEditor.attributedString];
     [printView setTextColor:[NSColor blackColor]];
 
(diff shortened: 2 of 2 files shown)


jleandroperez (commented):
@charliescheer I'm unable to trigger the original issue in `trunk` + `light mode`.



1. [NSView(s) have a maximum Size](https://developer.apple.com/documentation/appkit/nswindow/maxsize), which hint that perhaps this workaround may break in some edge scenarios.

2. The TextView itself doesn't have bounds defined, so **it's likely** that [usedRectForTextContainer](https://developer.apple.com/documentation/appkit/nslayoutmanager/usedrect(for:)) might not be accurate, when there's word wrapping involved

3. I'm also noticing this:



Running on Sequoia 15.3.1 / Xcode 16.2!

charliescheer:
Updating.  Did some more digging here and the problem seems to be a Text kit 2 issue.  By forcing the NSTextView back to Text kit 1 the issue is fixed

jleandroperez (approved):
✅ Verified there are no regressions (on my setup!!)

Nice catch!!!

:shipit:

roundhill (approved):
I can't see!

PR #1242, opened 2025-06-12
Provision SwiftLint using SwiftPM instead of CocoaPods

mokagio:
See https://linear.app/a8c/issue/AINFRA-451. Similar to https://github.com/Automattic/simplenote-ios/pull/1712.

### DIFF
--- a/Simplenote.xcodeproj/project.pbxproj
+++ b/Simplenote.xcodeproj/project.pbxproj
@@ -1932,9 +1932,13 @@
 			inputFileListPaths = (
 				"${PODS_ROOT}/Target Support Files/Pods-Automattic-SimplenoteTests/Pods-Automattic-SimplenoteTests-frameworks-${CONFIGURATION}-input-files.xcfilelist",
 			);
+			inputPaths = (
+			);
 			name = "[CP] Embed Pods Frameworks";
 			outputFileListPaths = (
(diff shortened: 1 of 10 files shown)


Simplenote.xcodeproj/project.pbxproj
```diff
@@ -2039,7 +2047,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "[ $CI ] && exit 0\n./Pods/SwiftLint/swiftlint\n\n";
+			shellScript = "[ $CI ] && exit 0\nrake lint\n\n";
```
AliSoftware, line 2050: Same concern as with companion iOS PR here.

AliSoftware (approved):
Approving to unblock, but with the same note about running rake from Xcode's Script Build Phase.

Could be ok for now to unblock progress then be addressed in later PR though?

Simplenote.xcodeproj/project.pbxproj
mokagio: For future reference: https://github.com/Automattic/simplenote-ios/pull/1712#discussion_r2141740829

> Moved this discussion internally to paaHJt-8yZ-p2

PR #1243, opened 2025-06-13
Fetch Simperium via SwiftPM and remove CocoaPods

mokagio:
Same as https://github.com/Automattic/simplenote-ios/pull/1716 but for this app. See that PR for links and testing steps.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

readme.md
```diff
@@ -16,11 +16,6 @@ We use a few tools to help with development. To install or update the required d
 
 `rake dependencies`
 
-#### CocoaPods
-
-Simplenote for iOS uses [CocoaPods](http://cocoapods.org/) to manage third party libraries.
-Third party libraries and resources managed by CocoaPods will be installed by the `rake dependencies` command above.
-
```
AliSoftware, line 23: Might be worth replacing this section with `#### SwiftPM` to indicate this project uses SwiftPM nowadays?

readme.md
```diff
@@ -16,11 +16,6 @@ We use a few tools to help with development. To install or update the required d
...
 #### Why isn't Sparkle fetched with CocoaPods
 
 At the time of writing, the version of Sparkle distributed via CocoaPods doesn't support Sandboxing, a feature that we value in Simplenote macOS.
```
AliSoftware, line 21: This section of the README might need a quick update.
 - Now that CocoaPods is no longer used in the project to begin with, it feels odd to still have a mention of it in this section about Sparkle out of nowhere.
 - This also raises the question if nowadays we couldn't fetch Sparkle via SwiftPM and an xcframework? 🤔

AliSoftware (approved):
Code LGTM, and ran `pod deintegrate` to confirm that it wouldn't produce any `diff` and that everything about CocoaPods was removed.

I still left a couple of notes about documentation updates to remove leftover CocoaPods mention and update mention to SwiftPM which I think are worth addressing before merge.

readme.md
mokagio: For other reader's reference, in Simplenote iOS we agree to keep this off and to not mention SwiftPM. More in https://github.com/Automattic/simplenote-ios/pull/1716#discussion_r2149447959

readme.md
mokagio: I updated the docs here https://github.com/Automattic/simplenote-macos/pull/1244

> This also raises the question if nowadays we couldn't fetch Sparkle via SwiftPM and an xcframework? 🤔

As part of that PR, I checked the [Sparkle docs](https://sparkle-project.org/documentation/) and learned they support SwiftPM, so we could use that.

But the question in my mind is actually whether we should remove Sparkle in the first place.

We can use TestFlight to ship betas, and prefer that to a private distribution channel. I love Sparkle and how it lets you do cool things like opting-in to beta and other distribution channels but it might be unnecessary for this App Store app.

PR #1252, opened 2026-02-12
Add CLAUDE.md with agent instructions

mokagio:
## Summary

Adds `CLAUDE.md` with instructions for Claude Code agents working in this repo:

- Slow command guidance (set higher `max_turns` for `swift`/`xcodebuild`/`fastlane`)
- PR checklist (read `Dangerfile` before opening PRs)

Extracted from #1250, decoupled from the `.claude/settings.json` permission changes.

Context: https://linear.app/a8c/issue/AINFRA-1965

## Test Plan

~~- [ ] CI passes~~

N.A. 

---

@Automattic/apps-infrastructure given there's no one left to officially maintain Simplenote, I guess we can own merging this


AGENTS.md
```diff
@@ -0,0 +1,10 @@
+# CLAUDE.md
+
+## Slow commands
+
+`swift`, `xcodebuild`, and `fastlane` commands can take several minutes.
+Account for this when setting `max_turns` on parallel subagents — 40 or higher is a safe default.
+
+## Opening Pull Requests
+
+Before opening a PR, read the `Dangerfile` and proactively satisfy its checks.
```
mokagio, line 10: Mmm... I noticed that none of the PRs in this chain added the milestone. It's possible that's due to the syntax in the `Dangerfile` being unclear about the milestone requirement vs the milestone due date. Compare it with the label checks, for example:

https://github.com/Automattic/simplenote-macos/blob/cd1ee8ad0956f971165f34902c76f2fa8772ca00/Dangerfile#L33-L40

mokagio: I'm going to close this group of PRs and take a smaller step still: only track the slow commands part.

mokagio:
Closing this in favor of taking a smaller step focused only on the build-test speed cycle.

AGENTS.md
mokagio: Follow up: one of my agents picked up this PR as the base for the `AGENTS.md` adoption work I requested.

At this point, I think this is acceptable. If a better pattern for Danger / PR guidelines will emerge, we'll get use something like https://github.com/Automattic/apps-infra-claude-plugins/tree/trunk/plugins/multi-repo-orchestrator to apply here, too.

mokagio:
Given the state of this project, and the load of this kind of PRs that I opened, I'm going to admin-merge this PR.

PR #1263, opened 2026-07-10
Probe: Ruby 3.4.9 native gem compilation across CI queues

mokagio:
**Do not merge.** Throwaway branch to answer one question before the Ruby 3.4.9 + Xcode 26.6 rollout lands on this repo for real.

## Rationale

The nokogiri failures we hit on newer Xcode images were caused by the pinned Ruby, not the toolchain — bumping to 3.4.9 let every gem compile from source. This branch checks that the same holds on all three queues the Apple repos touch (`mac`, `android`, `linter`), rather than assuming it from the doapple result.

## Intentional tradeoffs

The app build/test and SwiftLint steps are deleted so the probe is cheap. `.xcode-version` jumps 15.4 → 26.6, which the app build would not survive; landing that for real is a separate, much larger change than the Ruby bump.

Round 2 of each queue `depends_on` round 1 on purpose: `install_gems` keys its cache on `pipeline-platform-arch-ruby-Gemfile.lock`, so parallel rounds would both miss and prove nothing about the restore path.

## Gotchas

The `linter` queue rejects any command outside its handler list (`swiftlint`, `rubocop`, `danger`, gradle wrapper validation), so `bundle exec fastlane` cannot run there. `rubocop` is the only way to exercise gem installation on that agent — its handler reads `.ruby-version` and bundles `rubocop` out of our lock.

The S3 gem cache persists across builds. Round 1 is only genuinely cold on the *first* build for a given `(queue, ruby, Gemfile.lock)` triple; on re-runs both rounds restore. Check the `Restoring cache entry …` line before drawing conclusions.

## How to test

Watch the Buildkite build for this PR. Each probe step prints `Nokogiri::VERSION_INFO` (which reveals whether libxml2 was compiled or vendored), boots fastlane, and asserts `Gemfile.lock` came out of `bundle install` unchanged.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Message">
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
      <td data-sticky="false">This PR is still a Draft: some checks will be skipped.</td>
    </tr>
  </tbody>
</table>

</p>

PR #1264, opened 2026-07-22
Use latest Ruby and Bundler, 3.4.9 and 4.0.16

mokagio:
See https://linear.app/a8c/issue/AINFRA-2718. This is a routine update, done to pave the way for the upcoming `a8c-secrets` adoption work.


.buildkite/shared-pipeline-vars
```diff
@@ -5,7 +5,7 @@
 
 # The ~> modifier is not currently used, but we check for it just in case
 XCODE_VERSION=$(sed -E -n 's/^(~> )?(.*)/xcode-\2/p' .xcode-version)
-CI_TOOLKIT_PLUGIN_VERSION="3.4.2"
+CI_TOOLKIT_PLUGIN_VERSION="6.1.1"
```
AliSoftware, line 8: @copilot check in the Automattic/a8c-ci-toolkit-buildkite-plugin repo what breaking changes happened between 3.4.2 and 6.1.1 then which of those breaking changes might impact this project in particular (given how the helpers from that plugin are used in this repo), to ensure this major version bump here won't accidentally break things.

AliSoftware (approved):
:shipit:

PR #1269, opened 2026-07-22
Decrypt secrets outside of the repo

mokagio:
Prep step to then adopt [`a8c-secrets`](https://github.com/Automattic/a8c-secrets), which requires secrets outside of the repo. See https://linear.app/a8c/issue/AINFRA-2731. 

I modelled the script changes on the recent Gravatar SDK implementation, which benefitted from the back and forth with @AliSoftware.

**Same** 

- Precedence: internal store first, gitignored external file second
- Hard failure with a two-audience hint
- `Secrets.template.swift` → `*.external-contributors.swift` pair, writing to `SCRIPT_OUTPUT_FILE_0` under `DERIVED_FILE_DIR`
- `README` explaining sources generation and why they are red

**Different**, aside for the differences due to the repo not using `a8c-secrets` yet

- Unlike Gravatar, the script in Simplenote has two target consumers:`Simplenote` and `IntentsExtension`. The `SCRIPT_OUTPUT_FILE_0` indirection is "load-bearing" here, because so one script and one `.xcfilelist` back two identical phases, each writing into _its own target's derived dir_
- **Input-list guard.** I kept the pre-existing `ensure_is_in_input_files_list` check (errors if a source isn't declared in the `.xcfilelist`). Gravatar hasn't one, but I think it's useful and we should eventually adopt it there, too (especially if the endgame is for the script to be generic enough that it can run from `a8c-secrets`, i.e. without the source being in the repo and therefore the only knobs we can turn being input and ouput paths/lists)

<details>
<summary>AI-generated Details</summary>

### Fix

The decrypted `SPCredentials.swift` no longer lands in the checkout.
Each target that compiles it — `Simplenote` and `IntentsExtension` — carries a `Copy Secret` build phase writing into that target's own `$(DERIVED_FILE_DIR)`, replacing the `SimplenoteSecrets` aggregate target that wrote into `Simplenote/Credentials/`.
Internal contributors no longer copy anything into place by hand.

The shared demo credentials are gone with it.
`SPCredentials-demo.swift` becomes `SPCredentials.template.swift`, which external contributors copy to the gitignored `SPCredentials.external-contributors.swift` and fill in with their own Simperium credentials.
With neither that file nor the decrypted secrets present, the build fails in every configuration rather than silently compiling against demo values.

Part of [AINFRA-2731](https://linear.app/a8c/issue/AINFRA-2731), applying the pattern [Gravatar-SDK-iOS settled on](https://github.com/Automattic/Gravatar-SDK-iOS/blob/trunk/Scripts/build-phases/generate-secrets.sh).
The commit messages carry the per-step rationale.

### Test

```
bundle exec fastlane test
```

Both consumer targets should compile, each phase writing to its own `…/<Target>.build/DerivedSources/SPCredentials.swift`.
CI exercises both sources: `build-and-test` copies the template, so it runs the flow the readme gives external contributors, while `verify-app-store-target-builds` decrypts and builds against the real secrets.

</details>

### Review

One developer is enough.

The two `SPCredentials.swift` entries show in **red** in Xcode with no resolvable path.
That is expected — `DERIVED_FILE_DIR` only resolves while a specific target is building — and `Simplenote/DerivedSources/README.md` exists so the next reader does not "fix" it.

### DIFF
--- a/.buildkite/commands/build-and-test.sh
+++ b/.buildkite/commands/build-and-test.sh
@@ -3,5 +3,12 @@
 echo "--- :rubygems: Setting up Gems"
 install_gems
 
+echo "--- :closed_lock_with_key: Decrypting Secrets"
+# TODO: Drop once a8c-secrets is adopted — its CLI is light enough for the
+# `Copy Secret` build phase to decrypt on its own, the way Gravatar-SDK-iOS
+# does. `bundle` is not readily available to a build phase, so `configure_apply`
+# cannot move there.
+bundle exec fastlane run configure_apply
+
 echo "--- :hammer_and_wrench: Build and Test"
 bundle exec fastlane test
--- a/.buildkite/commands/build-and-test.sh
+++ b/.buildkite/commands/build-and-test.sh
@@ -3,12 +3,11 @@
 echo "--- :rubygems: Setting up Gems"
 install_gems
 
-echo "--- :closed_lock_with_key: Decrypting Secrets"
-# TODO: Drop once a8c-secrets is adopted — its CLI is light enough for the
-# `Copy Secret` build phase to decrypt on its own, the way Gravatar-SDK-iOS
-# does. `bundle` is not readily available to a build phase, so `configure_apply`
-# cannot move there.
-bundle exec fastlane run configure_apply
+echo "--- :key: Providing Credentials"
+# The unit tests do not need real credentials, so take the same path the readme
+# gives external contributors. That keeps this job out of the secret store and
+# makes it a regression test for the contributor flow.
+cp Simplenote/SPCredentials.template.swift Simplenote/SPCredentials.external-contributors.swift
 
 echo "--- :hammer_and_wrench: Build and Test"
 bundle exec fastlane test
--- a/.buildkite/commands/build-and-test.sh
+++ b/.buildkite/commands/build-and-test.sh
@@ -3,10 +3,8 @@
 echo "--- :rubygems: Setting up Gems"
 install_gems
 
+# Unit tests don't need real credentials
 echo "--- :key: Providing Credentials"
-# The unit tests do not need real credentials, so take the same path the readme
-# gives external contributors. That keeps this job out of the secret store and
-# makes it a regression test for the contributor flow.
 cp Simplenote/SPCredentials.template.swift Simplenote/SPCredentials.external-contributors.swift
 
 echo "--- :hammer_and_wrench: Build and Test"
--- a/.gitignore
+++ b/.gitignore
@@ -31,9 +31,6 @@ Simplenote/DerivedSources/*
 # Settings
 Simplenote/config.plist
 
-# Stale copies of the decrypted credentials, from before they moved to DerivedSources
-Simplenote/Credentials/
-
 # Bundler
 /vendor/
 /vendor/bundle/
--- a/.gitignore
+++ b/.gitignore
@@ -52,6 +52,8 @@ build/
 .configure-files/*
 !.configure-files/*.enc
 
+Simplenote/SPCredentials.external-contributors.swift
+
 # Swift Package Manager
 .build
 
--- a/.gitignore
+++ b/.gitignore
@@ -25,10 +25,13 @@ DerivedData
 .idea/
 *.hmap
 *.xcscmblueprint
-Simplenote/DerivedSources/
+Simplenote/DerivedSources/*
+!Simplenote/DerivedSources/README.md
 
 # Settings
 Simplenote/config.plist
+
+# Stale copies of the decrypted credentials, from before they moved to DerivedSources
 Simplenote/Credentials/
 
 # Bundler
--- a/Scripts/Build-Phases/copy-secret.sh
+++ b/Scripts/Build-Phases/copy-secret.sh
@@ -43,7 +43,7 @@ function ensure_is_in_input_files_list() {
     exit 1
   fi
 
-  if [ "$SCRIPT_INPUT_FILE_LIST_COUNT" -eq 0 ]; then
+  if [ "${SCRIPT_INPUT_FILE_LIST_COUNT:-0}" -eq 0 ]; then
     echo "error: No input file list given (.xcfilelist). Cannot continue."
     exit 1
   fi
--- a/Scripts/Build-Phases/copy-secret.sh
+++ b/Scripts/Build-Phases/copy-secret.sh
@@ -24,7 +24,7 @@ function ensure_is_in_input_files_list() {
   fi
   file_to_find=$1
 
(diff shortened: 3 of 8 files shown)


Scripts/Build-Phases/copy-secret.sh
```diff
@@ -53,37 +55,49 @@ SECRETS_ROOT="${HOME}/.configure/simplenote-macos/secrets"
...
-SECRETS_DESTINATION_FILE="${SRCROOT}/Simplenote/Credentials/SPCredentials.swift"
-mkdir -p $(dirname "$SECRETS_DESTINATION_FILE")
+SECRETS_DESTINATION_FILE="${SCRIPT_OUTPUT_FILE_0}"
+mkdir -p "$(dirname "$SECRETS_DESTINATION_FILE")"
 
-if cmp --silent -- ${SECRETS_FILE} ${SECRETS_DESTINATION_FILE}; then
+# `cp -v` names the destination, which differs per consumer target.
+apply() {
+    echo "Applying secrets from ${1}"
+    cp -v "$1" "$SECRETS_DESTINATION_FILE"
```
mokagio:
```suggestion
apply() {
    echo "Applying secrets from ${1}"
    # `cp -v` names the destination, which differs per consumer target.
    cp -v "$1" "$SECRETS_DESTINATION_FILE"
```

.gitignore
```diff
@@ -25,11 +25,11 @@ DerivedData
 .idea/
 *.hmap
 *.xcscmblueprint
-Simplenote/DerivedSources/
+Simplenote/DerivedSources/*
+!Simplenote/DerivedSources/README.md
 
 # Settings
 Simplenote/config.plist
-Simplenote/Credentials/
```
mokagio, line 32: I considered keeping this just to avoid leaking. But given the repo is in maintenance, meaning it doesn't see a lot of active development, the chance of a leak are low and I prioritized keeping things tidy.

Additionally, worth remembering, that even if something from `SPCredentials.swift` leaked, its content are already de facto public because they are part of the shipped binary.

Scripts/Build-Phases/copy-secret.sh
```diff
@@ -1,4 +1,26 @@
...
+#      reads it.
+#   2. Simplenote/SPCredentials.external-contributors.swift — gitignored, so
+#      external contributors can keep their own Simperium credentials with
+#      little-to-no risk of committing them, starting from a copy of the
+#      committed template.
+#
+# If neither is present, the build will fail.
+
+SECRETS_ROOT="${HOME}/.configure/simplenote-macos/secrets"
+SECRETS_FILE="${SECRETS_ROOT}/SPCredentials.swift"
```
mokagio, line 21: Of course, the paths here will soon change as we adopt `a8c-secrects`.

.buildkite/commands/build-and-test.sh
```diff
@@ -3,5 +3,11 @@
 echo "--- :rubygems: Setting up Gems"
 install_gems
 
+echo "--- :key: Providing Credentials"
+# The unit tests do not need real credentials, so take the same path the readme
+# gives external contributors. That keeps this job out of the secret store and
+# makes it a regression test for the contributor flow.
```
mokagio:
```suggestion
# Unit tests don't need real credentials
echo "--- :key: Providing Credentials"
```

readme.md
```diff
@@ -52,14 +52,8 @@ Simplenote is powered by the [Simperium Sync'ing protocol](https://www.simperium
...
-
-Please copy the **testing Simperium credentials** as follows:
-
-```
-mkdir -p Simplenote/Credentials && cp Simplenote/SPCredentials-demo.swift Simplenote/Credentials/SPCredentials.swift
-```
-
-This will allow you to compile and run the app on a device or a simulator.
+Credentials live in the `simperium*` properties of the `SPCredentials` type, which is generated at build time from production credentials kept outside the checkout, falling back to user-specified ones.
+When it finds neither, the build fails with instructions on how to provide them.
```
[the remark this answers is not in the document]
mokagio: Good catch — that sentence went stale when the demo credentials came out. Reworded in https://github.com/Automattic/simplenote-macos/commit/53218329.

*Posted by Claude (Opus 5) on behalf of @mokagio with approval.*

Scripts/Build-Phases/copy-secret.sh
```diff
@@ -20,13 +42,14 @@ function ensure_is_in_input_files_list() {
     echo "error: Input file list verification needs a path to verify!"
     exit 1
   fi
-  file_to_find=$1
 
-  if [ $SCRIPT_INPUT_FILE_LIST_COUNT -eq 0 ]; then
+  if [ "$SCRIPT_INPUT_FILE_LIST_COUNT" -eq 0 ]; then
```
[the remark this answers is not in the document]
mokagio: Agreed, and it makes the two guards consistent. Fixed in https://github.com/Automattic/simplenote-macos/commit/609051c9 — with the variable unset the script now prints `error: No input file list given (.xcfilelist). Cannot continue.` instead of aborting on the unbound variable.

*Posted by Claude (Opus 5) on behalf of @mokagio with approval.*

.gitignore
iangmaia: I get it, but at the same time, retaining one line (perhaps adding a comment) is cheap 🙂  anyway not a big deal

readme.md
```diff
@@ -48,18 +48,12 @@ You can also open the project by double clicking on `Simplenote.xcworkspace` fil
...
 
-
-Please copy the **testing Simperium credentials** as follows:
-
-```
-mkdir -p Simplenote/Credentials && cp Simplenote/SPCredentials-demo.swift Simplenote/Credentials/SPCredentials.swift
-```
-
-This will allow you to compile and run the app on a device or a simulator.
+Credentials live in the `simperium*` properties of the `SPCredentials` type, which is generated at build time from production credentials kept outside the checkout, falling back to user-specified ones.
```
iangmaia, line 55: Maybe good to include the template copy command here and explain what works without supplying your own credentials? 🤔

iangmaia:
> One developer is enough.

What? 😄

iangmaia (approved):
Looks good, but I'll leave the final merge to you (I left a couple of nits).

mokagio:
@iangmaia thanks for the review and sorry for the late followup.

I addressed your suggestions.

> > One developer is enough.
>
> What? 😄

That's a leftover of how the original team had the PR template configured:

https://github.com/Automattic/simplenote-macos/blob/03b8a157613f02b575f0c18bed5c618072c65357/.github/PULL_REQUEST_TEMPLATE.md#L10-L12

I think we could update the template to suite the current state of the project, cc @Automattic/greenhouse

PR #1274, opened 2026-07-23
Upgrade to the Sentry 2.x fastlane plugin

mokagio:
Even though the app is on dev hiatus, the upgrade will prevent further non-actionable PRs from Dependabot, such as #1273.

See [AINFRA-2742](https://linear.app/a8c/issue/AINFRA-2742).

Verified by running builds with Sentry upload in CI via https://github.com/Automattic/simplenote-macos/pull/1275

image
image

<details>
<summary>AI-generated details</summary>

### Fix

Upgrades `fastlane-plugin-sentry` from 1.x to 2.x, per [AINFRA-2742](https://linear.app/a8c/issue/AINFRA-2742), so Dependabot stops opening bumps we can't merge (e.g. #1273).

The plugin's 2.0.0 release removed the `sentry_upload_dsym` action outright, so the Gemfile bump and the call-site migration to `sentry_debug_files_upload` have to land in the same change. The replacement takes `path:` as an array instead of a single `dsym_path:` string; auth keys are unchanged. Same migration already shipped in [woocommerce-ios#17506](https://github.com/woocommerce/woocommerce-ios/pull/17506).

Two things ride along, each in its own commit:

- **A pre-existing RuboCop offense.** The `danger-dangermattic` 1.2.2 → 1.4.0 bump in 6097ca82 pulled in a RuboCop with the `Style/CollectionQuerying` cop, and the `Dangerfile` lints every file rather than just the diff. That leaves trunk red on every open PR — #1271 is still blocked on it.
- **How the upload finds the dSYMs.** `build_and_upload_app_store` read `ENV['DSYM_OUTPUT_PATH']`, which was always nil: `build_app` only exports it when the export produced a `.ipa`/`.pkg` *and* the derived path exists, and on macOS it derives `<name>.dSYM.zip` while gym writes `<name>.app.dSYM.zip`. Both lanes now read the dSYMs out of the build archive instead. Unlike `DSYM_OUTPUT_PATH`, `XCODEBUILD_ARCHIVE` is exported unconditionally — and `build_and_upload_app_store` already computes `archive_path` itself, so nothing has to predict a filename.

### Test

The upload only runs on a real signed release build, so CI can't exercise it. Verified locally:

1. A probe build confirmed `DSYM_OUTPUT_PATH` is nil (both env and lane context) while `Simplenote.app.dSYM.zip` sits in the build folder — the bug above, observed rather than inferred.
2. `sentry-cli debug-files find` against a real archive: `dSYMs/` resolves to the DWARF inside the dSYM bundle; the archive *root* resolves to the stripped binary in `Products/`. Hence the helper targets `dSYMs/`.
3. `bundle exec fastlane lanes` and `bundle exec rubocop` — clean.

`verify-app-store-target-builds` runs `build_simplenote`, so CI does cover the `output_name` addition through a real gym invocation.

### Review

Worth knowing: `sentry_debug_files_upload` searches a path recursively, which is what lets us hand it a directory and stop guessing filenames. The one remaining assumption is Xcode's `<archive>/dSYMs/` layout — public and stable, unlike gym's naming, which lives in a private method.

Not verified end-to-end: no CI job produces a signed release build, so the first real confirmation is the next release.

### Release

</details>

### DIFF
--- a/Scripts/update-translations.rb
+++ b/Scripts/update-translations.rb
@@ -77,7 +77,7 @@ def copy_comment(file, trans_strings, value)
 end
 
 langs = {}
-if ARGV.count.positive?
+if ARGV.any?
   ARGV.each do |key|
     unless (locale = ALL_LANGS[key])
       puts "Unknown language #{key}"
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -15,7 +15,6 @@ INTERNAL_SCHEME = 'Simplenote'
 APP_STORE_SCHEME = 'Simplenote'
 BUILD_FOLDER = File.join(PROJECT_FOLDER, 'build')
 APP_NAME = APP_STORE_SCHEME
-DSYM_ZIP_PATH = File.join(BUILD_FOLDER, "#{APP_NAME}.app.dSYM.zip").freeze
 APP_STORE_BUNDLE_IDENTIFIER = 'com.automattic.SimplenoteMac'
 APP_STORE_BUNDLE_IDENTIFIER_INTENTS = "#{APP_STORE_BUNDLE_IDENTIFIER}.IntentsExtension".freeze
 VERSION_FILE_PATH = File.join(PROJECT_FOLDER, 'config', 'Version.Public.xcconfig')
@@ -343,7 +342,7 @@ lane :build_internal_beta do |skip_dsym_upload: true|
 
   next if skip_dsym_upload || !UI.confirm('Do you want to upload the dSYM to Sentry?')
 
-  upload_dsym_to_sentry
+  upload_dsym_to_sentry(archive_path: lane_context[SharedValues::XCODEBUILD_ARCHIVE])
 end
 
 # Upload the localized metadata (from `fastlane/metadata/`) to App Store Connect
@@ -399,7 +398,7 @@ lane :build_and_upload_app_store do |options|
     archive_path: archive_path
   )
 
-  upload_dsym_to_sentry
+  upload_dsym_to_sentry(archive_path: archive_path)
 
   # Do not create the GitHub release unless explicitly requested
   if options[:create_github_release]
@@ -588,13 +587,24 @@ def generate_gitkeep_for_empty_locale_folders
   gitkeeps
 end
 
-def upload_dsym_to_sentry
-  UI.user_error!("Could not find dSYM at expected location '#{DSYM_ZIP_PATH}'.") unless File.exist?(DSYM_ZIP_PATH)
+# Upload the dSYMs from a build archive to Sentry.
+#
+# Reads them out of the archive rather than the zip `gym` writes alongside it:
+# `sentry-cli` searches a path recursively, so pointing it at the archive's
+# `dSYMs/` avoids having to predict that zip's filename. Aim at `dSYMs/` and not
+# the archive root, or the search settles on the stripped binary in `Products/`.
+#
+# @param archive_path [String] Path to the `.xcarchive` the build produced.
+#
+def upload_dsym_to_sentry(archive_path:)
+  dsyms_path = File.join(archive_path, 'dSYMs')
+  UI.user_error!("Could not find dSYMs at '#{dsyms_path}'.") unless File.directory?(dsyms_path)
 
   sentry_debug_files_upload(
     auth_token: get_required_env('SENTRY_AUTH_TOKEN'),
     org_slug: 'a8c',
     project_slug: 'simplenote-macos',
-    path: [DSYM_ZIP_PATH]
+    path: [dsyms_path],
+    type: 'dsym'
   )
 end
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -14,6 +14,10 @@ PROJECT = 'Simplenote.xcodeproj'
 INTERNAL_SCHEME = 'Simplenote'
 APP_STORE_SCHEME = 'Simplenote'
 BUILD_FOLDER = File.join(PROJECT_FOLDER, 'build')
+# Passed to the build actions as `output_name` so the artifact names are ours to
+# state rather than gym's to derive from the app name.
+APP_NAME = 'Simplenote'
+DSYM_ZIP_PATH = File.join(BUILD_FOLDER, "#{APP_NAME}.app.dSYM.zip").freeze
 APP_STORE_BUNDLE_IDENTIFIER = 'com.automattic.SimplenoteMac'
 APP_STORE_BUNDLE_IDENTIFIER_INTENTS = "#{APP_STORE_BUNDLE_IDENTIFIER}.IntentsExtension".freeze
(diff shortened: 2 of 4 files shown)


Scripts/update-translations.rb
```diff
@@ -77,7 +77,7 @@ def copy_comment(file, trans_strings, value)
 end
 
 langs = {}
-if ARGV.count.positive?
+if ARGV.any?
```
mokagio, line 80: Addresses an unrelated RuboCop violation surfaced by tooling upgrade.

fastlane/Fastfile
```diff
@@ -14,6 +14,10 @@ PROJECT = 'Simplenote.xcodeproj'
 INTERNAL_SCHEME = 'Simplenote'
 APP_STORE_SCHEME = 'Simplenote'
 BUILD_FOLDER = File.join(PROJECT_FOLDER, 'build')
+# Passed to the build actions as `output_name` so the artifact names are ours to
+# state rather than gym's to derive from the app name.
+APP_NAME = 'Simplenote'
```
mokagio:
```suggestion
APP_NAME = APP_STORE_SCHEME
```

fastlane/Fastfile
```diff
@@ -598,3 +586,25 @@ def generate_gitkeep_for_empty_locale_folders
...
+# Reads them out of the archive rather than the zip `gym` writes alongside it:
+# `sentry-cli` searches a path recursively, so pointing it at the archive's
+# `dSYMs/` avoids having to predict that zip's filename. Aim at `dSYMs/` and not
+# the archive root, or the search settles on the stripped binary in `Products/`.
+#
+# @param archive_path [String] Path to the `.xcarchive` the build produced.
+#
+def upload_dsym_to_sentry(archive_path:)
+  dsyms_path = File.join(archive_path, 'dSYMs')
+  UI.user_error!("Could not find dSYMs at '#{dsyms_path}'.") unless File.directory?(dsyms_path)
```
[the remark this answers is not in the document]
mokagio:
```suggestion
def upload_dsym_to_sentry(archive_path:)
  UI.user_error!('The given archive path value '#{archive_path} is missing. Did the build produced an .xcarchive?') if archive_path.to_s.empty?
  
  dsyms_path = File.join(archive_path, 'dSYMs')
  UI.user_error!("Could not find dSYMs at '#{dsyms_path}'.") unless File.directory?(dsyms_path)
```

fastlane/Fastfile
```diff
@@ -598,3 +586,27 @@ def generate_gitkeep_for_empty_locale_folders
...
+#
+# Reads them out of the archive rather than the zip `gym` writes alongside it:
+# `sentry-cli` searches a path recursively, so pointing it at the archive's
+# `dSYMs/` avoids having to predict that zip's filename. Aim at `dSYMs/` and not
+# the archive root, or the search settles on the stripped binary in `Products/`.
+#
+# @param archive_path [String] Path to the `.xcarchive` the build produced.
+#
+def upload_dsym_to_sentry(archive_path:)
+  UI.user_error!('The given archive path value '#{archive_path} is missing. Did the build produced an .xcarchive?') if archive_path.to_s.empty?
```
mokagio:
```suggestion
  UI.user_error!("The given archive path value '#{archive_path}' is missing. Did the build produced an .xcarchive?") if archive_path.to_s.empty?
```

fastlane/Fastfile
```diff
@@ -598,3 +586,27 @@ def generate_gitkeep_for_empty_locale_folders
...
+# Reads them out of the archive rather than the zip `gym` writes alongside it:
+# `sentry-cli` searches a path recursively, so pointing it at the archive's
+# `dSYMs/` avoids having to predict that zip's filename. Aim at `dSYMs/` and not
+# the archive root, or the search settles on the stripped binary in `Products/`.
+#
+# @param archive_path [String] Path to the `.xcarchive` the build produced.
+#
+def upload_dsym_to_sentry(archive_path:)
+  UI.user_error!("The given archive path value '#{archive_path}' is missing. Did the build produced an .xcarchive?") if archive_path.to_s.empty?
+
```
dangermattic: <table data-meta="generated_by_pr-check">
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td width="100%" data-sticky="false"><span data-href="https://github.com/Automattic/simplenote-macos/blob/5e1c9bd7af7dc0e8862ea773fd59e18855c9ead9/fastlane/Fastfile#L601"></span>Layout/TrailingWhitespace: Trailing whitespace detected.</td>
    </tr>
  </tbody>
</table>

mokagio:
```suggestion

```

fastlane/Fastfile
```diff
@@ -402,15 +398,7 @@ lane :build_and_upload_app_store do |options|
     archive_path: archive_path
   )
 
-  sentry_upload_dsym(
-    auth_token: get_required_env('SENTRY_AUTH_TOKEN'),
-    org_slug: 'a8c',
-    project_slug: 'simplenotemacos',
```
AliSoftware, line 408: I noticed that before your change there was a mix of two different slugs used in the Fastlane code, `simplenotemacos' here vs `simplenote-macos` on line L347, the latter being the one you ended up using in your new helper on line R607.

Just wanted to make sure you noticed and double checked you picked the right one

AliSoftware (commented):
> Verified by running builds with Sentry upload

In not 100% convinced this is enough to be confident that the upload worked as expected, because I remember having encountered the case multiple times in the past that this sentry fastlane action does not report a failure upstream if it failed (and why we missed the fact in the past in some other repos/apps that the dSYMs stopped being uploaded as expected)

So I'd suggest to double check in Sentry itself not just rely on those CI logs

fastlane/Fastfile
AliSoftware: Confirmed that the correct Sentry project name is indeed the one with a dash
image
image

AliSoftware:
> > Verified by running builds with Sentry upload
> 
> In not 100% convinced this is enough to be confident that the upload worked as expected, because I remember having encountered the case multiple times in the past that this sentry fastlane action does not report a failure upstream if it failed (and why we missed the fact in the past in some other repos/apps that the dSYMs stopped being uploaded as expected)
> 
> So I'd suggest to double check in Sentry itself not just rely on those CI logs

Checked in Sentry (took me quite some time to find the right page, it's quite burried in their UI!)

image

And confirmed they matched with the UUIDs listed in the logs from your screenshots in the PR description (from [this job's log lines](https://buildkite.com/automattic/simplenote-macos/builds/517/canvas?jid=019f8dad-3be8-4789-9e08-ba856e3b4b67&tab=output#019f8dad-3be8-4789-9e08-ba856e3b4b67/L1309-L1311) and [that one](https://buildkite.com/automattic/simplenote-macos/builds/517/canvas?sid=019f8dad-3bab-4246-b6e4-d1acda6a830b&tab=output#019f8dad-3be7-4a61-b33a-009437914353/L1376))

AliSoftware (approved):

PR #1285, opened 2026-08-26
Set up `a8c-secrets` identities

mokagio:
Just getting ahead on the adoption work while I wait for some builds to finish.

Closes https://linear.app/a8c/issue/AINFRA-2976


AliSoftware (approved):
Verified using `age-keygen -y` that the public keys in `keys.pub` match the private keys we have in Secret Store, and the CI private key in mobile-secrets.

